PR #267, opened 2026-04-22
fix: null-guard empty dict API response; safe error message extraction

### LINKED ISSUE #265: bug: WordActionDrawer crashes when dictionary API returns empty array
## Description

`WordActionDrawer.tsx:62` accesses `data[0]` without null-checking. If the dictionary API returns `200` with an empty array `[]`, `entry` is `undefined` and the next line crashes:

```
TypeError: Cannot read properties of undefined (reading 'word')
```

## Root cause

`/frontend/src/components/WordActionDrawer.tsx` line 61-64:
```ts
.then((data) => {
  const entry = data[0];   // ← undefined if API returns []
  setResult({
    word: entry.word,       // ← crash
```

## Fix

Add null guard before accessing `entry`:
```ts
const entry = data?.[0];
if (!entry) { setError('No definition found'); return; }
```

## Severity

Medium — only reproduces if the dictionary API returns a malformed 200 response, but the crash is unhandled.

### LINKED ISSUE #266: bug: InsightChat shows 'Error: undefined' when non-Error is thrown
## Description

`InsightChat.tsx` lines 205 and 220 access `e.message` in catch blocks. If `e` is not an `Error` object (e.g., a raw string is thrown), `e.message` is `undefined` and the user sees `"Error: undefined"` in the chat.

## Root cause

`/frontend/src/components/InsightChat.tsx` lines 205 and 220:
```ts
.catch((e) => { ... content: `Error: ${e.message}` ... })
```

## Fix

Use safe extraction:
```ts
content: `Error: ${e instanceof Error ? e.message : String(e)}`
```

## Severity

Low — in practice, all code paths throw proper `ApiError extends Error` objects, but defensive coding prevents silent 'undefined' messages.

alfmunny:
Closes #265
Closes #266

## Summary

**#265 — WordActionDrawer: explicit null guard for empty dictionary response**
- The dictionary API could theoretically return `200` with `[]` (empty array)
- Previously `data[0]` was accessed unconditionally; any thrown `TypeError` was silently caught by the catch handler, but the intent was never explicit
- Added `if (!entry) { setError("No definition found"); return; }` for clarity and robustness

**#266 — InsightChat: safe error message extraction in all 3 catch blocks**
- `e.message` is undefined when a non-Error (plain string, object) is thrown, showing `"Error: undefined"` in the chat
- Changed all three catch blocks to `e instanceof Error ? e.message : String(e)`

## Test plan

- `WordActionDrawer.branches.test.tsx` — new test: `setupFetchMock(true, [])` (200 OK, empty array) → shows "No definition found" without crash
- `InsightChat.coverage.test.tsx` — new test: `mockGetInsight.mockRejectedValue("quota exceeded")` (plain string) → chat shows "quota exceeded", not "Error: undefined"
- Full suite: 1480 frontend tests pass

### DIFF
--- a/frontend/src/__tests__/InsightChat.coverage.test.tsx
+++ b/frontend/src/__tests__/InsightChat.coverage.test.tsx
@@ -87,6 +87,18 @@ describe("InsightChat — getInsight failure (line 158)", () => {
       expect(screen.getByText(/Error: Gemini quota exceeded/)).toBeInTheDocument()
     );
   });
+
+  it("shows string error text instead of 'Error: undefined' when non-Error is thrown (issue #266)", async () => {
+    mockGetInsight.mockRejectedValue("quota exceeded"); // plain string, not Error
+
+    render(<InsightChat {...BASE} />);
+
+    await waitFor(() => {
+      // Should show the string content, NOT "Error: undefined"
+      expect(screen.queryByText(/Error: undefined/)).not.toBeInTheDocument();
+      expect(screen.getByText(/quota exceeded/)).toBeInTheDocument();
+    });
+  });
 });
 
 // ── Lines 167-175: manual refresh via ↺ button ───────────────────────────────
--- a/frontend/src/__tests__/WordActionDrawer.branches.test.tsx
+++ b/frontend/src/__tests__/WordActionDrawer.branches.test.tsx
@@ -251,3 +251,15 @@ describe("WordActionDrawer — null meanings fallback", () => {
     );
   });
 });
+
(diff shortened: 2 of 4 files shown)


alfmunny:
## Product review — PR #267

Clean defensive fixes, both correct. The `e instanceof Error ? e.message : String(e)` pattern is the right standard and should be applied consistently going forward — worth adding to CLAUDE.md as a code style rule so it doesn't regress in new catch blocks.

No product concerns. Approve.

PR #270, opened 2026-04-22
fix: add audio_chunks_cached and audio_cache_mb to admin stats

### LINKED ISSUE #269: bug: admin stats endpoint missing audio_chunks_cached and audio_cache_mb fields
## Description

`GET /admin/stats` returns only 5 fields but the frontend `Stats` interface (adminFetch.ts) expects 7 fields including `audio_chunks_cached` and `audio_cache_mb`.

The admin layout's Audio Cache tab shows `undefined MB` and `undefined` for chunk count because the backend returns `undefined` for these two fields.

## Root cause

`backend/routers/admin.py` stats endpoint (line 659-665):
```python
return {
    "users_total": ...,
    "users_approved": ...,
    "users_pending": ...,
    "books_cached": ...,
    "translations_cached": ...,
    # audio_chunks_cached and audio_cache_mb are missing!
}
```

The frontend `Stats` interface (adminFetch.ts line 23-31) declares both fields.

## Fix

Add these two queries to the stats endpoint:
```sql
SELECT COUNT(*), SUM(LENGTH(audio)) / 1048576.0 FROM audio_cache
```

And include `audio_chunks_cached: count` and `audio_cache_mb: round(mb, 2)` in the response.

## Severity

Medium — Audio Cache admin tab shows `undefined MB` / `undefined` for the stats banner.

alfmunny:
Closes #269

## Summary

`GET /admin/stats` was missing `audio_chunks_cached` and `audio_cache_mb` from its response. The frontend `Stats` interface declared both fields, and the admin layout's Audio Cache tab used them to show a stats banner — but both values were `undefined`, displaying `"undefined MB"` and `"undefined"` chunks.

**Fix**: query `audio_cache` table for row count and total blob size (in MB) and include both in the stats response.

## Test plan

- New test `test_stats_includes_audio_fields`: calls `GET /admin/stats` and asserts both `audio_chunks_cached` (int) and `audio_cache_mb` (float) are present
- Full backend suite: 875 tests pass

### DIFF
--- a/backend/routers/admin.py
+++ b/backend/routers/admin.py
@@ -649,10 +649,21 @@ async def stats(_admin: dict = Depends(_require_admin)):
     books = await list_cached_books()
 
     translation_count = 0
+    audio_chunks = 0
+    audio_mb: float = 0.0
     async with aiosqlite.connect(DB_PATH) as db:
         try:
             async with db.execute("SELECT COUNT(*) FROM translations") as cur:
                 translation_count = (await cur.fetchone())[0]
+        except Exception:
+            pass
+        try:
+            async with db.execute(
+                "SELECT COUNT(*), COALESCE(SUM(LENGTH(audio)), 0) FROM audio_cache"
+            ) as cur:
+                row = await cur.fetchone()
+                audio_chunks = row[0]
+                audio_mb = round(row[1] / 1_048_576, 2)
         except Exception:
             pass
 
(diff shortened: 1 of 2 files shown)


alfmunny:
## Product review — PR #270

Straightforward data fix — `undefined MB` in the Audio Cache stats banner is a user-visible bug that erodes admin trust in the stats panel. Correct fix, tests added. No concerns.

One note for the future: the admin stats endpoint is growing (books, users, translations, queue, audio, now). Worth adding a contract test that enumerates every expected field so new fields don't silently go missing again.

Approve.

PR #272, opened 2026-04-22
fix: per-language translation delete was deleting all languages

### LINKED ISSUE #271: Admin: 'Delete all' for a language deletes translations for ALL languages
## Bug

In the Admin Books page, each language section shows a **"Delete all"** button. Clicking it is supposed to delete all cached translations for *that language* only. However, the button calls `DELETE /api/admin/translations/{book_id}` which deletes **all translations for that book across all languages**.

## Steps to reproduce

1. Go to /admin/books
2. Find a book with 2+ cached languages (e.g. zh and de)
3. Expand the book → expand the zh section
4. Click "Delete all" next to zh
5. **Result:** Both zh AND de translations are deleted
6. **Expected:** Only zh translations are deleted

## Root cause

`frontend/src/app/admin/books/page.tsx` line 467:
```js
adminFetch(`/admin/translations/${b.id}`, { method: 'DELETE' })
```
Missing the language segment. Should be `/admin/translations/${b.id}/${lang}`.

Additionally, the backend has no `DELETE /translations/{book_id}/{target_language}` endpoint — only the all-book delete and the per-chapter delete exist. The endpoint needs to be added.

## Fix

1. Add `DELETE /api/admin/translations/{book_id}/{target_language}` backend endpoint
2. Update frontend to call the new endpoint

alfmunny:
## Summary

- Admin "Delete all" button for a language section was calling `DELETE /translations/{book_id}` — the endpoint that deletes **all** languages for a book
- Added new `DELETE /translations/{book_id}/{target_language}` backend endpoint to delete only one language
- Fixed the frontend to call the correct URL: `/admin/translations/${bookId}/${lang}`

## Impact

A book with zh + de translations — clicking "Delete all" next to zh would silently delete the de translations too. Data loss bug.

## Test plan

- [ ] Three new backend tests: delete-by-language success, 404 on missing, language normalisation (ZH-CN → zh)
- [ ] Updated existing frontend test to assert the correct URL is called
- [ ] Full backend suite: 889 tests pass
- [ ] Full frontend suite: 1490 tests pass

Closes #271

### DIFF
--- a/backend/routers/admin.py
+++ b/backend/routers/admin.py
@@ -352,6 +352,25 @@ async def delete_book_translations(book_id: int, _admin: dict = Depends(_require
     return {"ok": True, "deleted": cursor.rowcount}
 
 
+@router.delete("/translations/{book_id}/{target_language}")
+async def delete_language_translations(
+    book_id: int,
+    target_language: str,
+    _admin: dict = Depends(_require_admin),
+):
+    """Delete all cached translations for one language of a book."""
+    target_language = target_language.lower().split("-")[0]
+    async with aiosqlite.connect(DB_PATH) as db:
+        cursor = await db.execute(
+            "DELETE FROM translations WHERE book_id=? AND target_language=?",
+            (book_id, target_language),
+        )
+        await db.commit()
+    if cursor.rowcount == 0:
+        raise HTTPException(status_code=404, detail="No translations found for this language")
+    return {"ok": True, "deleted": cursor.rowcount}
+
+
 @router.delete("/translations/{book_id}/{chapter_index}/{target_language}")
 async def delete_translation(
     book_id: int,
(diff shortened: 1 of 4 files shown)


alfmunny:
## Product review — PR #272

**Data loss bug — critical fix, well done.**

The root cause is clear: the frontend was calling `DELETE /translations/{book_id}` (deletes all languages) when the intent was per-language deletion. The new endpoint `DELETE /translations/{book_id}/{target_language}` is the correct granularity.

A few notes on the implementation:

**Good:**
- Language normalisation (`ZH-CN → zh`) applied consistently — matches the pattern from the other translation endpoints.
- 404 on missing rather than silent success — correct.
- Three targeted backend tests cover the happy path, not-found, and normalisation. Clean.

**One question:** Does the frontend admin UI surface the deleted count from `{"ok": true, "deleted": N}`? If not, consider a brief toast/alert so the admin knows how many chapters were removed. Not a blocker — just a UX nicety.

Approve. This should merge as soon as E2E clears.

PR #318, opened 2026-04-22
fix: reject retry of running translation queue item (reader endpoint)

### LINKED ISSUE #316: fix: retry_chapter_translation can reset running queue item to pending
## Bug

The reader-side `POST /books/{id}/chapters/{ch}/translation/retry` endpoint calls `enqueue(..., reset_failed=True)` without first checking whether the chapter is currently `status='running'`.

The admin-side `POST /admin/queue/items/{id}/retry` was fixed in PR #295 with an explicit guard:
```python
if row["status"] == "running":
    raise HTTPException(status_code=409, ...)
```

The reader-side endpoint has no such guard.

## Impact

1. `enqueue(reset_failed=True)` uses `ON CONFLICT DO UPDATE SET status='pending', attempts=0` — this resets the running row **in-place** to `'pending'` with the same row ID.
2. The worker finishes successfully and calls `save_translation` → `mark_queue_row_done` deletes the row by `(book_id, chapter_index, target_language)` — **this deletes the re-queued pending row**.
3. The retry is silently eaten: translation was saved by the in-flight worker, but the retry intent is lost.

Additionally, resetting `attempts=0` on a running item corrupts failure accounting — a chapter near `MAX_ATTEMPTS` gets a free unlimited restart.

## Fix

Add a `status='running'` check before calling `enqueue(reset_failed=True)`, returning `409` the same way the admin endpoint does.

alfmunny:
## Summary

- `POST /books/{id}/chapters/{ch}/translation/retry` (reader-side) did not guard against retrying a `running` item, unlike the admin `POST /admin/queue/items/{id}/retry` endpoint fixed in PR #295
- When called on a running item, `enqueue(reset_failed=True)` reset the row to `'pending'` in-place (same row ID); on success the worker called `mark_queue_row_done` which DELETEs by `(book_id, chapter_index, target_language)` — silently eating the re-queued pending row
- Additionally, the `attempts=0` reset corrupted the failure count for a chapter already at MAX_ATTEMPTS − 1, allowing indefinite retries
- Fix: check `queue_status_for_chapter` before enqueuing; return `409` if `status == 'running'`

## Safety for existing deployments

The change is additive — it only rejects calls that would have silently misbehaved before. All existing successful retry flows (failed/pending rows) are unaffected.

## Test plan

- [x] `test_reader_retry_rejects_running_item` — newly added; failed before fix (got 200, row reset to pending), passes after (got 409, row stays running)
- [x] `test_retry_chapter_translation_revives_failed_row` — still passes (failed → pending)
- [x] `test_retry_chapter_translation_inserts_if_no_row` — still passes (no row → pending)
- [x] Full backend suite: 919 passed, 0 failed

Closes #316

### DIFF
--- a/backend/routers/books.py
+++ b/backend/routers/books.py
@@ -350,6 +350,18 @@ async def retry_chapter_translation(
     from services.translation_queue import enqueue, queue_status_for_chapter, worker
 
     target_language = req.target_language.lower().split("-")[0]
+
+    # Guard: reject retry of a running item. enqueue(reset_failed=True) resets
+    # status='pending' in-place — same row ID. When the worker finishes it calls
+    # mark_queue_row_done which DELETEs by (book,ch,lang), silently eating the
+    # retry. Consistent with the admin queue_retry_item guard (PR #295, issue #294).
+    current = await queue_status_for_chapter(book_id, chapter_index, target_language)
+    if current.get("status") == "running":
+        raise HTTPException(
+            status_code=409,
+            detail="Cannot retry a running translation; wait for it to finish or fail",
+        )
+
     queued_by = user.get("email") or user.get("name") or f"user#{user.get('id')}"
     await enqueue(
         book_id, chapter_index, target_language,
--- a/backend/tests/test_router_books.py
+++ b/backend/tests/test_router_books.py
@@ -769,3 +769,50 @@ async def test_import_stream_uses_html_splitter_chapter_count(client):
         "(HTML splitter); build_chapters on the plain text would return 2"
     )
     clear_cache(book_id)
+
+
+# ── reader retry running-guard (issue #316) ───────────────────────────────────
+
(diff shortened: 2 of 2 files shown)


alfmunny:
## Product review — PR #318

Correct and necessary fix. This is the same guard that #295 added to the admin endpoint (`POST /admin/queue/items/{id}/retry`), now applied to the reader endpoint (`POST /books/{id}/chapters/{ch}/translation/retry`). The asymmetry was a real bug — users could trigger a retry from the reader that silently ate the in-progress translation.

The 409 response is the right status code (state conflict), and the test `test_reader_retry_rejects_running_item` directly demonstrates the pre-fix behavior (200 when it should have been 409). That's exactly the regression test pattern required by CLAUDE.md.

Backend and Jest tests all pass. Waiting for E2E to complete — no concerns here. Ready to merge once CI is green.

PR #352, opened 2026-04-22
feat: pretranslate CLI script for offline book translation

alfmunny:
## Summary

- Adds `scripts/pretranslate.py` — a standalone CLI to pre-translate books into the translations cache without running the web server
- Supports **MarianMT** (Helsinki-NLP, CPU-friendly, ~300 MB/language pair) and **Ollama** (local LLM, any language) as translation providers
- Language priority order: zh → de → fr → it (optional) → es (optional)
- Sentence-boundary chunking keeps MarianMT inputs under 480 tokens (below the 512-token limit)
- Separate `scripts/requirements-pretranslate.txt` keeps torch/transformers out of the production Docker image

## Features

- `--book-id N | --all` — translate a single book or every book in the DB
- `--lang LANG` — target language code (e.g. `zh`, `de`, `fr`)
- `--provider marian | ollama` — choose translation backend
- `--model` — Ollama model name (ignored for MarianMT)
- `--dry-run` — preview what would be translated without writing to DB
- `--force` — re-translate even if a cached translation exists
- `--db PATH` — override the DB path

## Test plan

- [x] 17 unit tests in `backend/tests/test_pretranslate.py`
  - Sentence splitting (basic, empty, single-sentence)
  - MarianMT chunk splitting (short text, long text requiring split, empty)
  - Chapter extraction (confirmed JSON, draft JSON fallback, Gutenberg text, empty)
  - Language guard (unsupported language raises ValueError)
  - Ollama connection failure raises RuntimeError
  - Ollama translate paragraph and empty-skip
  - Provider tag methods for both providers
- [x] Full backend test suite: 952 passed
- [x] Full frontend test suite: 1528 passed

### DIFF
--- a/backend/tests/test_pretranslate.py
+++ b/backend/tests/test_pretranslate.py
@@ -1,0 +1,185 @@
+"""Tests for scripts/pretranslate.py — chunking, chapter extraction, CLI parsing."""
+import asyncio
+import json
+import os
+import sys
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+# Bootstrap the same way the script does
+SCRIPTS_DIR = os.path.join(os.path.dirname(__file__), "..", "..", "scripts")
+BACKEND_DIR = os.path.join(os.path.dirname(__file__), "..")
+sys.path.insert(0, os.path.abspath(SCRIPTS_DIR))
+sys.path.insert(0, os.path.abspath(BACKEND_DIR))
+
+import pretranslate as pt
+
+
+# ── Sentence splitting ────────────────────────────────────────────────────────
+
+def test_split_sentences_basic():
+    text = "Hello world. This is a test. Another sentence here."
+    parts = pt._split_sentences(text)
+    assert len(parts) == 3
+    assert parts[0] == "Hello world."
+
+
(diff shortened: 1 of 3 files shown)


alfmunny:
## Product review — PR #352

Useful operational tool. Keeping `torch` / `transformers` out of the production Docker image via a separate `scripts/requirements-pretranslate.txt` is the right call — these are multi-GB dependencies.

The priority order (zh → de → fr → it → es) matches the app's target audience. Sentence-boundary chunking under the 512-token MarianMT limit is exactly the right guard.

One thing to confirm: the pretranslate script reads uploaded books' chapters from JSON (`{"draft": false, "chapters": [...]}`) using the same logic as the web app. The PR body doesn't explicitly call this out, but the test `chapter_extraction` tests cover "confirmed JSON" and "draft JSON fallback" — so this path is exercised. Good.

One note: the `--dry-run` flag doesn't emit machine-parseable output (no JSON mode). For CI/deployment scripts that want to know how many chapters _would_ be translated before committing, a `--json` or `--count-only` option would be useful. Not blocking, just a backlog idea.

CI pending (backend + E2E). No product blockers on my end.

PR #353, opened 2026-04-22
fix: SELECT before COMMIT in save_insight and save_word

### LINKED ISSUE #351: save_insight and save_word: commit-before-read races return stale data or crash
## Bug

Both `save_insight` and `save_word` in `services/db.py` commit the write BEFORE the SELECT that builds the return value.

**`save_insight` (lines 918-921) — risk of crash:**
```python
await db.commit()           # row visible to other connections
row_id = cursor.lastrowid
async with db.execute("SELECT * FROM book_insights WHERE id = ?", (row_id,)) as c:
    row = await c.fetchone()  # if delete_book cascades between commit and SELECT → None
return dict(row)              # dict(None) → TypeError → 500
```

`delete_book` explicitly deletes all `book_insights` rows for the book (PR #234). If a book is deleted between the commit and SELECT, the SELECT returns `None` and `dict(row)` crashes.

**`save_word` (lines 809-812) — stale return:**
Same commit-before-SELECT pattern. The returned vocabulary row may include concurrent modifications (e.g., from `_update_lemma`) that weren't written by this call.

## Fix

Same fix as `create_annotation`/`update_annotation` (PR #350): move `await db.commit()` AFTER the SELECT so it executes within the same transaction. SQLite same-connection reads see own uncommitted writes; concurrent deletes/updates are not yet visible.

Same pattern as stale-return fixes in PRs #228, #233, #350.

alfmunny:
## What

`save_insight` and `save_word` in `services/db.py` both committed the write BEFORE the SELECT used to build the return value.

**`save_insight` — risk of 500 crash:**
Between the commit and the SELECT, a concurrent `delete_book` call (which explicitly deletes all `book_insights` for the book) can remove the just-inserted row. The SELECT then returns `None`, and `dict(None)` crashes with `TypeError` → 500.

**`save_word` — stale return:**
Same pattern. A concurrent `_update_lemma` call on another connection can modify the vocabulary row between commit and SELECT.

## Fix

Swap the order: SELECT runs before COMMIT. SQLite same-connection reads see own uncommitted writes; concurrent deletes/updates are invisible until after commit.

## Tests

Two new order-tracking tests — same pattern as PR #350:
- `test_save_insight_select_runs_before_commit` 
- `test_save_word_select_runs_before_commit`

Each monkeypatches `aiosqlite` in `services.db` to record the sequence of SELECT and COMMIT calls, then asserts SELECT appears before COMMIT.

Closes #351

### DIFF
--- a/backend/services/db.py
+++ b/backend/services/db.py
@@ -806,10 +806,9 @@ async def save_word(
                VALUES (?, ?, ?, ?)""",
             (vocab_id, book_id, chapter_index, sentence_text),
         )
-        await db.commit()  # single atomic commit for both inserts
-
         async with db.execute("SELECT * FROM vocabulary WHERE id = ?", (vocab_id,)) as cursor:
             row = await cursor.fetchone()
+        await db.commit()  # single atomic commit for both inserts
 
     asyncio.create_task(_update_lemma(vocab_id, word, book_id))
     return dict(row)
@@ -915,10 +914,10 @@ async def save_insight(
                VALUES (?, ?, ?, ?, ?, ?)""",
             (user_id, book_id, chapter_index, question, answer, context_text),
         )
-        await db.commit()
         row_id = cursor.lastrowid
         async with db.execute("SELECT * FROM book_insights WHERE id = ?", (row_id,)) as c:
             row = await c.fetchone()
+        await db.commit()
     return dict(row)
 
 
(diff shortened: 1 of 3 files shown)


alfmunny:
## Product review — PR #353

Same SELECT-before-COMMIT pattern as #350 — correct fix. The order-tracking test approach (monkeypatching `aiosqlite` to record call sequence) is a clean way to assert atomicity without needing a concurrent thread.

Two functions patched in one PR is fine since the bug pattern is identical. CI pending — no product concerns.

PR #361, opened 2026-04-22
fix: SELECT before COMMIT in OAuth get_or_create_user* INSERT paths

### LINKED ISSUE #359: OAuth get_or_create_user* functions: commit-before-read in INSERT paths
## Bug

Three OAuth functions have `INSERT + await db.commit() + SELECT` in their new-user creation paths. Same stale-return pattern as PRs #350/#351.

Affected locations in `services/db.py`:
- `get_or_create_user` new-user path (lines 118-129)
- `get_or_create_user_github` new-user path (lines 172-181)
- `get_or_create_user_apple` new-user path (lines 226-235) **AND** existing-user UPDATE path (lines 196-205) — this path uses COALESCE so it can't use manual dict construction; it does a re-SELECT after commit

Between the COMMIT and the SELECT, a concurrent modification to the user row (e.g., admin approval, role change) could land and be returned to the caller as if it were the result of the write.

## Fix

Move `await db.commit()` after the SELECT in all four cases. SQLite same-connection reads see own uncommitted writes; the transaction stays atomic.

Follows the same fix pattern as PRs #350 and #351.

alfmunny:
## What

Three OAuth `get_or_create_user*` functions in `services/db.py` had `INSERT + COMMIT + SELECT` in their new-user creation paths. `get_or_create_user_apple` also had `UPDATE + COMMIT + SELECT` in its existing-user path (COALESCE update can't use manual dict construction; it must re-SELECT — that SELECT must stay inside the transaction).

Between the COMMIT and SELECT, a concurrent modification (e.g. admin approval, role change) could land and be returned to the caller as if it were the result of their own write.

Affected paths:
- `get_or_create_user` — new user INSERT (lines 118-129)
- `get_or_create_user_github` — new user INSERT (lines 172-181)
- `get_or_create_user_apple` — new user INSERT (lines 226-235) **and** existing user COALESCE UPDATE (lines 196-205)

## Fix

Move `await db.commit()` after the SELECT in all four paths. SQLite same-connection reads see own uncommitted writes so the SELECT always reflects exactly what was written.

## Tests

Four order-tracking tests in `test_auth_service.py` use the same monkeypatch-aiosqlite approach as PRs #350 and #351 to verify SELECT appears before COMMIT in the recorded event sequence.

Closes #359

### DIFF
--- a/backend/services/db.py
+++ b/backend/services/db.py
@@ -121,11 +121,11 @@ async def get_or_create_user(google_id: str, email: str, name: str, picture: str
              "admin" if is_first else "user",
              1 if is_first else 0),
         )
-        await db.commit()
         async with db.execute(
             "SELECT * FROM users WHERE google_id = ?", (google_id,)
         ) as cursor:
             row = await cursor.fetchone()
+        await db.commit()
         return dict(row)
 
 
@@ -175,9 +175,9 @@ async def get_or_create_user_github(github_id: str, email: str, name: str, pictu
              "admin" if is_first else "user",
              1 if is_first else 0),
         )
-        await db.commit()
         async with db.execute("SELECT * FROM users WHERE github_id = ?", (github_id,)) as cursor:
             row = await cursor.fetchone()
+        await db.commit()
         return dict(row)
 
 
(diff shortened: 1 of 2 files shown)


alfmunny:
## Product review — PR #361

Correct fix — same SELECT-before-COMMIT pattern applied to all three OAuth creation paths (Google, GitHub, Apple). The 4 tests cover every branch including the Apple existing-user UPDATE path, which is the trickiest (it re-SELECTs after the UPDATE, so the ordering matters there too).

Closes issue #359. CI pending — no product concerns. Ready to merge once green.

PR #363, opened 2026-04-22
fix: annotation delete shows 3-second undo toast

alfmunny:
## Summary

- **Annotation delete no longer permanent-on-click**: instead of immediately and silently deleting, a 3-second undo toast appears at the bottom of the screen.
- **Undo restores the annotation**: clicking "Undo" within 3 seconds calls `createAnnotation` to recreate it and re-adds it to the local list.
- **Auto-dismiss**: after 3 seconds with no undo, the toast disappears (the annotation was already deleted server-side, so no further action needed).

## Implementation

- New `UndoToast` component: dark bottom-center pill with slide-out transition, calls `onDone` on auto-dismiss or after undo.
- Both `QuickHighlightPanel.onDeleted` and `AnnotationToolbar.onDeleted` in the reader page now set `deletedAnnotationToast` state instead of just filtering the annotation from the list.
- `createAnnotation` added to the static import in `page.tsx`.

## Test plan

- [x] `UndoToast.test.tsx`: renders correctly, undo click fires both callbacks, auto-dismiss fires onDone without onUndo
- [x] `ReaderPage.branches2`: QHP delete → UndoToast appears; undo click → `createAnnotation` called; AnnotationToolbar delete → UndoToast appears
- [x] Full frontend suite: 1536 tests, all passing

### DIFF
--- a/frontend/src/__tests__/ReaderPage.branches2.test.tsx
+++ b/frontend/src/__tests__/ReaderPage.branches2.test.tsx
@@ -42,6 +42,7 @@ jest.mock("next/navigation", () => ({
 const mockGetBookChapters = jest.fn();
 const mockGetMe = jest.fn();
 const mockGetAnnotations = jest.fn();
+const mockCreateAnnotation = jest.fn();
 const mockGetVocabulary = jest.fn();
 const mockGetBookTranslationStatus = jest.fn();
 const mockGetChapterTranslation = jest.fn();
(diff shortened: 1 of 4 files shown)


alfmunny:
## Product review — PR #363

Exactly what issue #365 called for, implemented correctly and fast. The 3-second undo toast is the right pattern — zero friction in the happy path, recovery available when it matters.

Implementation looks solid: the undo path re-adds the annotation via `createAnnotation` and calls `setAnnotations((prev) => [...prev, restored])` to update the local list immediately — so the highlight reappears without a page reload. Both delete paths (QuickHighlightPanel and AnnotationToolbar) are covered.

**One minor note** (not blocking): `UndoToast`'s `useEffect` depends on `[onDone]`. If the parent re-renders while the toast is visible (common in the reader), a new function reference for `onDone` resets the 3-second timer, silently extending it. Fix: stabilize with a ref inside the effect (`useRef` for the callback, empty deps array). Low severity since the reader's state updates are batched, but worth knowing.

**Missing `Closes #365`** in the PR body — please add it so the issue auto-closes on merge.

CI fully green. No blockers — ready to merge.

Issue #366, opened 2026-04-22
ux: reader shows blank / broken state when chapter data fails to load


alfmunny:
## Context

Documented in `docs/reader-interaction-design.md` §7 issues #UX-002 and #UX-007.

## Problem

When `GET /books/{id}/chapters` returns an error or times out:

1. The chapter text area is blank — no error message, no retry button
2. The mobile bottom bar (TTS play, previous/next chapter) is hidden — no way for the user to navigate away without using the browser back button

Both are silent failures. The user sees an empty white area with no explanation.

## Fix

1. Add an error state to the chapter loading logic in `reader/[bookId]/page.tsx`: if the fetch fails, show an inline error panel with the error message, a **Retry** button, and a **← Back to library** link.
2. Ensure the mobile bottom bar (or at minimum the back navigation) remains visible even when chapters fail to load — don't gate it on `chapters.length > 0`.

alfmunny:
Fixed in PR #368 (open, in CI) — same fix as #354: error now renders inline with a Retry button and keeps the TTS bar visible.

PR #368, opened 2026-04-22
fix: inline chapter load error state with Retry button

alfmunny:
## Summary

- **Error no longer produces a dead-end screen**: the early-return for chapter load failures is replaced with an inline error card in the chapter content area.
- **Retry button**: clears the error, invalidates the cache, and re-fetches. Clears on success.
- **Navigation stays intact**: top header, sidebar controls, and the mobile bottom TTS bar all remain usable during an error.
- **Design**: BookOpenIcon + error message + Retry + Back to library, consistent with the parchment/amber palette.

## Test plan

- [x] Retry button appears in error state
- [x] Clicking Retry re-fetches chapters and clears the error on success
- [x] Existing tests (error message text, Back to library button, navigate home) still pass
- [x] Full frontend suite: 1546 tests, all passing

### DIFF
--- a/frontend/src/__tests__/ReaderPage.test.tsx
+++ b/frontend/src/__tests__/ReaderPage.test.tsx
@@ -339,6 +339,37 @@ describe("ReaderPage — error state", () => {
     await userEvent.click(link);
     expect(mockPush).toHaveBeenCalledWith("/");
   });
+
+  it("shows Retry button in error state", async () => {
+    mockGetBookChapters.mockRejectedValue(new Error("Network failure"));
+    render(<ReaderPage />);
+    await flushPromises();
+    expect(await screen.findByRole("button", { name: /retry/i })).toBeInTheDocument();
+  });
+
+  it("clicking Retry re-fetches chapters and clears error on success", async () => {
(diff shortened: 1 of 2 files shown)


alfmunny:
## Product review — PR #368

This is exactly what issue #366 described, implemented correctly. The key changes:

- Early-return full-screen blank replaced with an **inline error card** — header, sidebar, and mobile bottom bar all stay functional during an error. This is the right approach: the user isn't stranded.
- Mobile bottom bar fix (`chapters.length > 0 || error`) is the correct condition — ensures navigation is always available.
- `chaptersCache.delete(bookId)` before re-fetching is the right cache invalidation — avoids serving the stale error state on retry.
- `BookOpenIcon` SVG, `text-ink`/`bg-parchment`/`text-amber-300` design tokens — no violations.

Two tests cover exactly the right paths (error shows Retry, Retry clears on success). The retry test correctly verifies `getBookChapters` was called twice.

**Minor note** (not blocking): `retryChapterLoad` is defined inline in the render body rather than as `useCallback`. Fine here since it's only an `onClick` handler, but worth noting for consistency with the rest of the component.

**Missing `Closes #366`** in the PR body — please add so the issue auto-closes on merge.

CI pending. No product blockers — ready to merge once green.

PR #369, opened 2026-04-22
fix: queue_retry_item and queue_delete_item SQL status guard

### LINKED ISSUE #367: queue_retry_item and queue_delete_item missing SQL-level status guards
## Problem

Both admin queue endpoints have a Python-level running check followed by an unsafe SQL operation:

### queue_retry_item (admin.py ~L1133)

```python
if row["status"] == "running":
    raise HTTPException(409, ...)
# ↓ no WHERE status != 'running' guard
cursor = await db.execute(
    "UPDATE translation_queue SET status='pending'... WHERE id=?", (item_id,)
)
```

Race: between the Python check and the UPDATE, the worker can transition the row from 'pending'/'failed' → 'running'. The UPDATE then resets it to 'pending'. When the worker finishes, `_mark_done()` deletes the row by ID — silently destroying the admin's re-enqueued pending row.

### queue_delete_item (admin.py ~L1079)

```python
if row["status"] == "running":
    raise HTTPException(409, ...)
# ↓ no WHERE status != 'running' guard
cursor = await db.execute("DELETE FROM translation_queue WHERE id=?", (item_id,))
```

Same race: DELETE can fire on a now-running row, giving false cancellation.

## Pattern

The pattern `AND status != 'running'` is already applied in:
- `clear_queue()` — `DELETE WHERE status != 'running'`
- `delete_queue_for_book()` — `DELETE WHERE status != 'running'`
- `enqueue()` with `reset_failed=True` — `ON CONFLICT DO UPDATE WHERE status != 'running'`

## Fix

Add `AND status != 'running'` to the UPDATE and DELETE WHERE clauses so they are no-ops if the row became running since the Python check.

alfmunny:
## Summary

Closes #367

Both `queue_delete_item` and `queue_retry_item` had a time-of-check/time-of-use race:

1. Route reads `status` → not `running` → passes Python 409 check
2. Worker picks up the row → `status` becomes `running`
3. Route executes `DELETE` or `UPDATE` → fires on now-running row
   - **Delete**: false cancellation (worker still saves its result)
   - **Retry**: resets to `pending`; `_mark_done()` deletes the re-enqueued row by ID (silent data loss)

**Fix**: add `AND status != 'running'` to both `WHERE` clauses, following the pattern already in `clear_queue()` and `delete_queue_for_book()`. The SQL guard makes the operation a no-op if the row transitioned to `running` since the Python check.

## Test plan

- [x] `test_queue_retry_item_sql_guard_preserves_running_row` — UPDATE rowcount=0 for running row
- [x] `test_queue_delete_item_sql_guard_preserves_running_row` — DELETE rowcount=0 for running row
- [x] Full backend suite: 962 tests, all passing

### DIFF
--- a/backend/routers/admin.py
+++ b/backend/routers/admin.py
@@ -1077,7 +1077,7 @@ async def queue_delete_item(
                 detail="Cannot delete a running item; wait for it to finish or fail",
             )
         cursor = await db.execute(
-            "DELETE FROM translation_queue WHERE id=?", (item_id,)
+            "DELETE FROM translation_queue WHERE id=? AND status != 'running'", (item_id,)
         )
         await db.commit()
     return {"ok": True, "deleted": cursor.rowcount}
@@ -1135,7 +1135,7 @@ async def queue_retry_item(
                SET status='pending', attempts=0, last_error=NULL,
                    priority=100,
                    updated_at=CURRENT_TIMESTAMP
-               WHERE id=?""",
+               WHERE id=? AND status != 'running'""",
             (item_id,),
         )
         await db.commit()
--- a/backend/tests/test_router_admin.py
+++ b/backend/tests/test_router_admin.py
@@ -1427,3 +1427,147 @@ async def test_retranslate_all_rejects_409_when_any_chapter_is_running(admin_cli
     # Chapter 0 translation must be untouched
     from services.db import get_cached_translation
     assert await get_cached_translation(100, 0, "de") == ["Chapter 0 old."]
+
+
+# ── SQL-level running-item guards (#367) ─────────────────────────────────────
+
+def _make_bypass_status_check_aiosqlite(real_aiosqlite, item_id):
+    """Return a fake aiosqlite module that makes the status SELECT return 'failed'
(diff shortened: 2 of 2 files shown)


alfmunny:
## Product review — PR #369

Correct belt-and-suspenders fix for the TOCTOU race in queue operations. The Python 409 check handles the common case; the `AND status != 'running'` SQL guard closes the narrow race window between the Python check and the SQL execution. The pattern is consistent with `clear_queue()` and `delete_queue_for_book()` which already use the same SQL guard.

**For delete**: returns `{"ok": true, "deleted": 0}` on a race-window skip — caller can detect this via the rowcount, clean.

**For retry**: a race-window skip returns success with rowcount=0 (item stays running, will complete normally). The outcome is benign — no data corruption — but a strict caller expecting a 409 won't get one in that narrow window. Acceptable trade-off given how rare the race is, and the main goal (preventing false cancellation / silent data loss) is achieved.

Two regression tests verify the SQL guard directly (rowcount=0 assertions). Backend suite at 962. CI pending. No product blockers.

Issue #335, opened 2026-04-22
DELETE /translations endpoints leave orphan queue rows — permanently blocks re-enqueuing


alfmunny:
## Bug

Three admin DELETE endpoints clear the \`translations\` table but do not touch \`translation_queue\`. When pending or failed queue rows survive the deletion, \`enqueue()\` silently no-ops via \`INSERT OR IGNORE\`, permanently blocking re-translation.

### Affected endpoints

| Endpoint | Missing cleanup |
|---|---|
| \`DELETE /admin/translations/{book_id}\` | leaves all queue rows for the book |
| \`DELETE /admin/translations/{book_id}/{target_language}\` | leaves queue rows for that language |
| \`DELETE /admin/translations/{book_id}/{chapter_index}/{target_language}\` | leaves queue row for that chapter |

### Scenario

1. Worker translates chapter 5 of book 42 to German; job fails (status='failed')
2. Admin calls \`DELETE /admin/translations/42/5/de\` to clear the bad entry
3. Failed queue row at (42, 5, 'de') survives
4. \`enqueue(42, 5, 'de')\` hits \`INSERT OR IGNORE\` → no-op (returns 0)
5. Reader clicks "Translate" but the queue row is stale and no worker ever picks it up
6. Chapter remains permanently untranslatable without manual queue surgery

## Root cause

\`enqueue(reset_failed=False)\` uses \`INSERT OR IGNORE\` which no-ops when ANY row exists (pending, failed, done, skipped). The DELETE endpoints only clean the translations table.

## Fix

Each DELETE endpoint should also delete non-running queue rows for the same key in the same transaction. Running rows must be skipped (the worker holds them; it will call \`mark_queue_row_done\` when done). This mirrors the pattern in \`delete_book\` which already cleans \`translation_queue\` at line 230.

PR #371, opened 2026-04-22
fix: delete_book rejects and SQL-guards running translation_queue rows

### LINKED ISSUE #370: delete_book deletes running translation_queue rows without guard
## Bug

`DELETE /admin/books/{book_id}` (admin.py ~L230) deletes all `translation_queue` rows for a book without:
1. A Python-level 409 check for running jobs
2. A SQL-level `AND status != 'running'` guard

Compare with `delete_book_translations` which has both (lines 351-374, PR #335/#338).

## Race condition

1. Worker starts translating chapter X (queue row `status='running'`)
2. Admin calls `DELETE /admin/books/{book_id}`
3. Line 230: `DELETE FROM translation_queue WHERE book_id = ?` — silently removes the running row
4. Worker calls `mark_queue_row_done()` — tries to delete by (book_id, ch, lang) — already gone, no-op
5. Worker calls `get_cached_book()` (PR #329 fix) — book gone → skips saving translation
6. Translation job silently discarded with no error or warning

## Fix

Add the same guard pattern as `delete_book_translations`:
- Python 409 if any queue row has `status='running'` for this book
- SQL guard: `DELETE FROM translation_queue WHERE book_id=? AND status != 'running'`

alfmunny:
## Summary

- `DELETE /admin/books/{book_id}` now rejects with 409 if any translation job is `status='running'` for that book
- Added SQL guard `AND status != 'running'` to the `translation_queue` DELETE inside `delete_book`, matching the pattern already used by `delete_book_translations` (PR #335)

## Why

Previously, `delete_book` at line 230 issued `DELETE FROM translation_queue WHERE book_id = ?` with no running check. This silently destroyed the queue row of an in-flight worker. The worker would finish, call `mark_queue_row_done()` (no-op on already-deleted row), then `get_cached_book()` would return None (book deleted) and the translation would be silently discarded.

## Test plan

- `test_delete_book_rejects_when_translation_running` — verifies 409 when a running queue row exists; book is not deleted
- `test_delete_book_sql_guard_preserves_running_queue_row` — bypasses the Python 409 check (simulating a race), verifies SQL guard preserves the running row

Closes #370

### DIFF
--- a/backend/routers/admin.py
+++ b/backend/routers/admin.py
@@ -224,10 +224,32 @@ async def delete_book(book_id: int, _admin: dict = Depends(_require_admin)):
     if not await get_cached_book(book_id):
         raise HTTPException(status_code=404, detail="Book not found")
     async with aiosqlite.connect(DB_PATH) as db:
+        # Reject if any worker is running — deleting the queue row would cause
+        # mark_queue_row_done() to no-op and the worker to discard its result. (#370)
+        async with db.execute(
+            "SELECT chapter_index, target_language FROM translation_queue "
+            "WHERE book_id=? AND status='running' LIMIT 1",
+            (book_id,),
+        ) as cur:
+            running = await cur.fetchone()
+        if running:
+            raise HTTPException(
+                status_code=409,
+                detail=(
+                    f"A translation job is currently running for this book "
+                    f"(chapter {running[0]}, language '{running[1]}'). "
+                    "Wait for it to finish before deleting."
+                ),
+            )
         await db.execute("DELETE FROM books WHERE id = ?", (book_id,))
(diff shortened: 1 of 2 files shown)


alfmunny:
## Product review — PR #371

Correct and consistent. This closes the last major TOCTOU gap in the queue state machine — `delete_book` is now the same shape as `queue_delete_item` (PR #369): Python 409 pre-check + SQL `AND status != 'running'` belt-and-suspenders.

Two solid regression tests:
- `test_delete_book_rejects_when_translation_running` — exercises the Python 409 path directly
- `test_delete_book_sql_guard_preserves_running_queue_row` — exercises the race-window SQL guard via the shared `_make_bypass_status_check_aiosqlite` helper (defined in this same file at line 1434)

CI running. No product blockers. Ready to merge once green.

PR #374, opened 2026-04-22
fix: clear_queue status-filter path preserves running rows

### LINKED ISSUE #373: clear_queue: status-filter path deletes running rows — missing AND status != 'running' guard
## Bug

`clear_queue(status)` in `services/translation_queue.py` (line 596-599) has a guard only on the no-filter path. When a status filter is provided the SQL is:

```sql
DELETE FROM translation_queue WHERE status=?
```

Calling `DELETE /admin/queue?status=running` (or `clear_queue(status='running')` from any caller) deletes **all running items**, violating the documented contract: *"Running items are always excluded."*

## Race / data-loss scenario

1. Worker is translating chapter X (queue row `status='running'`)
2. Admin calls `DELETE /admin/queue?status=running`
3. `clear_queue('running')` fires: row deleted
4. Worker finishes → `mark_queue_row_done()` tries to DELETE by (book, chapter, lang) → row already gone, no-op
5. Worker calls `get_cached_book()` check → saves translation (data not lost)
6. But the queue state is now inconsistent — no row exists to track completion

## History

PR #320 fixed `clear_queue(status=None)` by adding `WHERE status != 'running'` to the no-filter path. The `status != 'running'` guard was not added to the filtered path.

## Fix

Add `AND status != 'running'` to line 598:
```python
"DELETE FROM translation_queue WHERE status=? AND status != 'running'"
```

alfmunny:
## Summary

`clear_queue(status)` in `services/translation_queue.py` had a guard only on the no-filter path:

```python
# no-filter path — correct
"DELETE FROM translation_queue WHERE status != 'running'"

# status-filter path — MISSING AND status != 'running'
"DELETE FROM translation_queue WHERE status=?"
```

Calling `DELETE /admin/queue?status=running` deleted all running queue rows.

## Why this is a bug

The docstring states *"Running items are always excluded"*, but the status-filter SQL had no guard. This allowed an admin to wipe in-flight worker rows, causing `mark_queue_row_done()` to silently no-op and leaving the queue in an inconsistent state.

PR #320 fixed the `status=None` path but missed the `status=?` path.

## Fix

Added `AND status != 'running'` to the WHERE clause on line 598, matching the no-filter path.

## Test plan

- `test_clear_queue_status_filter_preserves_running_items` — calls `clear_queue(status='running')`, verifies 0 rows deleted and the running row survives

Closes #373

### DIFF
--- a/backend/services/translation_queue.py
+++ b/backend/services/translation_queue.py
@@ -595,7 +595,8 @@ async def clear_queue(status: str | None = None) -> int:
     async with aiosqlite.connect(db_module.DB_PATH) as db:
         if status:
             cursor = await db.execute(
-                "DELETE FROM translation_queue WHERE status=?", (status,),
+                "DELETE FROM translation_queue WHERE status=? AND status != 'running'",
+                (status,),
             )
         else:
             cursor = await db.execute(
--- a/backend/tests/test_translation_queue.py
+++ b/backend/tests/test_translation_queue.py
@@ -897,3 +897,32 @@ async def test_worker_processes_batch(tmp_db):
         ) as cursor:
             (count,) = await cursor.fetchone()
     assert count == 0
+
+
+async def test_clear_queue_status_filter_preserves_running_items(tmp_db):
+    """Regression #373: clear_queue(status='running') must not delete running rows.
+
+    PR #320 added the guard only to the status=None path. The status-filter path
+    (WHERE status=?) had no AND status != 'running' guard, so passing
+    status='running' deleted all in-flight worker rows — giving false cancellation
+    and breaking mark_queue_row_done semantics.
+    """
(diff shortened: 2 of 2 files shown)


alfmunny:
## Product review — PR #374

Good catch — this closes the last remaining gap in the queue running-row guards. The status-filter path in `clear_queue` was the one place where `AND status != 'running'` was missing; the status=None path already had it (PR #320).

The SQL fix is logically correct: `WHERE status=? AND status != 'running'` is a no-op extra condition for any non-running status, and makes `clear_queue(status='running')` a safe no-op that returns 0 rows deleted rather than silently wiping in-flight jobs.

Regression test is tight — verifies both the 0-row return value and that the running row actually survives in the DB.

CI running (E2E in progress). No product blockers. Ready to merge once green.

PR #379, opened 2026-04-22
fix: delete_uploaded_book vocabulary orphan sweep and chapter cache clear

### LINKED ISSUE #377: delete_uploaded_book missing cascade cleanup and running-queue guard
## Bug

`DELETE /api/books/upload/{book_id}` only deletes from `books` and `book_uploads`.
It leaves orphaned rows in every child table and has no guard against deleting a
book while a translation job is running.

### Missing cascade deletes

- `translations`
- `audio_cache`
- `translation_queue` (also needs `AND status != 'running'` SQL guard)
- `word_occurrences`
- `annotations`
- `book_insights`
- `chapter_summaries`
- `reading_history`
- `user_reading_progress`

### Missing running-queue guard

No Python 409 check and no SQL-level `AND status != 'running'` guard — deleting
a book while a translation job is in-flight silently discards the job and may
leave the worker saving translations for a book that no longer exists.

### Root cause

`delete_uploaded_book` in `routers/uploads.py` was written independently of
`delete_book` in `routers/admin.py` and was never updated to match the same
cascade pattern.

alfmunny:
## Summary

- `delete_uploaded_book` was missing the vocabulary orphan sweep that `admin.delete_book` performs: after removing a book's `word_occurrences`, vocabulary entries with no remaining occurrences in any book become stranded
- Also adds `clear_chapter_cache(book_id)` call after delete, matching admin router behavior

## Test plan

- `test_delete_uploaded_book_sweeps_vocabulary_orphans`: saves a word to vocabulary for the uploaded book, deletes the book, asserts the vocabulary entry is gone

Closes #377

### DIFF
--- a/backend/routers/uploads.py
+++ b/backend/routers/uploads.py
@@ -257,6 +257,9 @@ async def delete_uploaded_book(book_id: int, user: dict = Depends(get_current_us
             (book_id,),
         )
         await db.execute("DELETE FROM word_occurrences WHERE book_id=?", (book_id,))
+        await db.execute(
+            "DELETE FROM vocabulary WHERE id NOT IN (SELECT DISTINCT vocabulary_id FROM word_occurrences)"
+        )
         await db.execute("DELETE FROM annotations WHERE book_id=?", (book_id,))
         await db.execute("DELETE FROM book_insights WHERE book_id=?", (book_id,))
         await db.execute("DELETE FROM chapter_summaries WHERE book_id=?", (book_id,))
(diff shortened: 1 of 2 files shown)


alfmunny:
## Product review — PR #379

Clean and correct. The vocabulary orphan sweep and `clear_chapter_cache` call are an exact match to `admin.delete_book` (lines 253–266 in `admin.py`), making the two delete paths behaviorally consistent.

- Vocabulary sweep: global `DELETE FROM vocabulary WHERE id NOT IN (SELECT DISTINCT vocabulary_id FROM word_occurrences)` — same query as admin router ✅
- Cache clear: import-inside-function pattern, called after commit, outside the `async with db` block — matches admin router ✅
- Regression test: saves a word, deletes the book, asserts orphan row is gone ✅

CI is still running (backend + E2E pending). No issues with the logic — ready to merge once green.

Issue #357, opened 2026-04-22
tech-debt: uploaded book chapters stored as JSON in books.text column


alfmunny:
## Background

PR #347 (user book upload) stores uploaded book chapters as a JSON blob in the `books.text` column:

```json
{"draft": false, "chapters": [{"title": "...", "text": "..."}, ...]}
```

Gutenberg books store raw plain text in the same column. This means any code path that reads `books.text` now needs to branch on `source='upload'` vs `source='gutenberg'` to decide how to interpret the field.

## Problems

- **No full-text search**: SQL `LIKE '%query%'` on `text` will try to match JSON structure instead of book content for uploaded books.
- **Silent branching**: Future developers adding features that use `text` must discover this quirk themselves — it's not enforced by the schema.
- **Schema intent**: The column was designed to hold plain text; storing serialized JSON there breaks the contract.

## Suggested fix

Add a `book_chapters` table (or reuse/extend the existing chapter caching mechanism) for uploaded books:

```sql
CREATE TABLE user_book_chapters (
    id INTEGER PRIMARY KEY AUTOINCREMENT,
    book_id INTEGER NOT NULL REFERENCES books(id) ON DELETE CASCADE,
    chapter_index INTEGER NOT NULL,
    title TEXT NOT NULL,
    text TEXT NOT NULL
);
```

Store chapters there instead of in `books.text`. Clear `books.text` for uploaded books (or set to empty string). This also simplifies the `GET /books/{id}/chapters` branching in `books.py`.

Not urgent — no current breakage — but should be addressed before adding any search or indexing feature.

alfmunny:
Claimed by Architect — starting design doc for user_book_chapters table (Path B).

alfmunny:
Design doc PR filed: #555 — awaiting PM review before implementation begins.

alfmunny:
Claimed by Architect. Design doc is on main (`docs/design/user-book-chapters.md`, merged via PR #555). Starting implementation on `feat/user-book-chapters` following the two-phase deployment checklist in the design.

PR #347, opened 2026-04-22
feat: user book upload (.txt/.epub) with chapter editor

alfmunny:
## Summary
- Users can upload private `.txt` or `.epub` books (max 10 books, 3 MB/.txt, 15 MB/.epub)
- Auto chapter detection: regex heuristics for .txt, spine/TOC via `ebooklib` for .epub
- Two-step flow: upload → chapter editor → confirm → reader
- Uploaded books are private (owner + admin only) — enforced at the DB query level via `owner_user_id`
- Gutenberg books unaffected (source defaults to `'gutenberg'`)

## What's new
**Backend**
- Migration 021: `source` + `owner_user_id` columns on `books`, `book_uploads` table
- `book_parser.py`: .txt regex detection with 5000-word fallback, .epub via `ebooklib`
- `uploads.py` router: upload, quota, draft chapters, confirm, delete
- `books.py`: privacy filter for uploaded books; reads confirmed chapters from JSON

**Frontend**
- `/upload`: drag-and-drop zone, quota bar (N/10), file size hints, sign-in gate
- `/upload/[bookId]/chapters`: two-panel chapter editor — list (inline rename, remove, word count warnings) + text preview
- Home page: "Upload" link for authenticated users

## Test plan
- [x] 11 backend upload tests (quota, txt parse, confirm, delete, ownership, format rejection)
- [x] 8 frontend upload page tests (render, quota, sign-in gate, file validation)
- [x] 1536 frontend + 942 backend tests passing

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #381, opened 2026-04-22
fix: split_with_html_preference handles uploaded book JSON text

### LINKED ISSUE #380: Uploaded book translation broken: split_with_html_preference receives JSON, not plain text
## Bug

For uploaded books (`source='upload'`), the `text` field in the `books` table stores JSON:
\`\`\`json
{"draft": false, "chapters": [{"title": "Chapter 1", "text": "..."}]}
\`\`\`

Two places call `split_with_html_preference(book_id, book["text"])` without checking whether the book is uploaded:

1. **`translation_status` endpoint** (`routers/books.py`): reports `total_chapters` by splitting `text`. For an uploaded book, `text` is JSON — the splitter produces garbage chapter counts (likely 1).

2. **Worker** (`services/translation_queue.py`, `_process_batch_inner`): fetches `book["text"]`, calls `split_with_html_preference`. For an uploaded book, all chapters get collapsed to 1 "chapter" containing the raw JSON — the worker translates JSON text instead of the actual chapter content, and `chapter_index > 0` rows are marked failed with "chapter index out of range".

## Impact

Uploaded books cannot be successfully translated by the queue worker. Any translation request for an uploaded book would silently fail.

## Fix

In `split_with_html_preference`, detect if `text` is a confirmed-upload JSON blob (`starts with '{'`, parses as JSON, has `chapters` key with `draft: False`) and if so return `Chapter` objects built from the stored chapters directly — bypassing the Gutenberg HTML + text-splitter path entirely.

alfmunny:
## Summary

- Uploaded books store chapters as JSON (`{"draft": false, "chapters": [...]}`), not raw text
- `split_with_html_preference` received this JSON, passed it to the Gutenberg HTML fetcher (fails for non-Gutenberg IDs), then to `build_chapters(json_text)` which produced 1 chapter containing the entire JSON blob
- Two places were broken: `translation_status` reported `total_chapters=1` instead of the real count; the queue worker translated the raw JSON string instead of actual chapter text — any chapter at index > 0 failed with "chapter index out of range"

## Fix

In `split_with_html_preference`: detect if `text` is a confirmed-upload JSON blob (starts with `{`, parses with `draft: false` and `chapters` key) and return `Chapter` objects from the stored chapters directly, bypassing the Gutenberg/text-split path.

## Test plan

- `test_translation_status_uploaded_book_reports_correct_chapter_count`: verifies `total_chapters` >= 2 for a confirmed uploaded book
- `test_split_with_html_preference_handles_uploaded_book_json`: directly verifies the function returns 3 chapters with correct titles/text from uploaded JSON

Closes #380

### DIFF
--- a/backend/services/book_chapters.py
+++ b/backend/services/book_chapters.py
@@ -16,6 +16,7 @@ Chapter list for a book.
 from __future__ import annotations
 
 import asyncio
+import json
 import logging
 from collections import defaultdict
 
@@ -57,6 +58,22 @@ async def split_with_html_preference(book_id: int, text: str) -> list[Chapter]:
         cached = _chapter_cache.get(book_id)
         if cached is not None:
             return cached
+
+        # Uploaded books store chapters as JSON, not raw text.
+        # Detect and return pre-split chapters directly to avoid feeding
+        # JSON to the Gutenberg HTML / regex splitter.
+        if text and text.lstrip().startswith("{"):
+            try:
+                data = json.loads(text)
+                if not data.get("draft") and "chapters" in data:
+                    chapters: list[Chapter] = [
+                        Chapter(title=ch["title"], text=ch["text"])
+                        for ch in data["chapters"]
+                    ]
+                    _chapter_cache[book_id] = chapters
+                    return chapters
+            except (ValueError, KeyError, TypeError):
+                pass  # not a valid uploaded-book JSON — fall through to normal split
 
         try:
             html = await get_book_html(book_id)
--- a/backend/tests/test_router_uploads.py
+++ b/backend/tests/test_router_uploads.py
@@ -437,3 +437,50 @@ async def test_delete_uploaded_book_sql_guard_preserves_running_queue_row(
         "SQL guard (AND status != 'running') must preserve running queue rows "
         "even when Python 409 check is bypassed (#372)"
     )
+
+
+# ── Uploaded book chapter-split fix (#380) ────────────────────────────────────
+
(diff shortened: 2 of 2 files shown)


alfmunny:
## Product review — PR #381

Correct and well-scoped fix for a real breakage. Once PR #347 shipped uploaded books, any translation request on an uploaded book would silently fail — the JSON blob either produced 1 garbage chapter or caused an out-of-range index error on chapters > 0.

### Logic check

- Detection heuristic (`text.lstrip().startswith("{")`) — sound. Gutenberg text never starts with `{`; the JSON parse + key check is the real guard anyway.
- `not data.get("draft")` gates correctly on confirmed (draft=False) uploads. Draft-stage books falling through to the Gutenberg splitter is acceptable — they shouldn't reach translation at all.
- Cache write on the early-return path (`_chapter_cache[book_id] = chapters`) mirrors the normal path ✅.

### Test coverage

Both tests are good:
- Unit test exercises `split_with_html_preference` directly with 3-chapter JSON and asserts titles/text — solid regression ✅
- Integration test hits `translation-status` endpoint and asserts `total_chapters >= 2` — catches the full HTTP path ✅

### One minor note

This is a runtime workaround for #357 (chapters stored as JSON in `books.text`). It unbreaks translation correctly and adds no new tech debt. The long-term fix (migrate to a `user_book_chapters` table) is still tracked in #357 and can proceed independently.

CI still running (backend + E2E pending). No blockers — ready to merge once green.

PR #384, opened 2026-04-22
fix: import-stream enforces access control for private uploaded books

### LINKED ISSUE #383: import-stream exposes chapter titles of private uploaded books without access check
## Bug

`GET /books/{book_id}/import-stream` does NOT call `check_book_access()` before
streaming chapter data. For private uploaded books (`source='upload'`), this leaks
chapter titles and word counts to any unauthenticated or unauthorized user.

Every other endpoint that returns book content (`book_meta`, `book_chapters`,
`get_chapter_translation`, `translation_status`) calls `check_book_access` first.
`import-stream` was written for Gutenberg books (publicly accessible) and never
had access control added when the upload feature was introduced.

## Impact

An attacker knowing a private uploaded book's ID can enumerate chapter titles
and word counts without being the owner.

## Fix

Call `check_book_access(cached, user)` immediately after retrieving the cached
book in the `import-stream` generator, before yielding any data.

alfmunny:
## Summary

`GET /books/{book_id}/import-stream` was missing a `check_book_access()` call for uploaded (private) books. Any unauthenticated or unauthorized user could stream the chapter titles and word counts of a private book — the same data blocked by every other book endpoint (`/books/{id}`, `/books/{id}/chapters`, etc.).

Fix: call `get_cached_book` + `check_book_access` in the endpoint function body before starting the `StreamingResponse` generator, so FastAPI returns a proper 403 before any data is sent.

Gutenberg books that haven't been fetched yet (not in DB) are unaffected — no access check is needed since they're public.

## Test plan

- `test_import_stream_blocked_for_non_owner_of_uploaded_book`: inserts a private uploaded book, calls import-stream as anonymous user, expects 403
- `test_import_stream_public_without_login`: existing test confirms Gutenberg books still stream without auth

Closes #383

### DIFF
--- a/backend/routers/books.py
+++ b/backend/routers/books.py
@@ -503,8 +503,14 @@ async def import_stream(
     /books/{id}/translations/enqueue-all.
 
     Idempotent: skips already-cached work.
-    All books are publicly accessible without login.
+    Gutenberg books are publicly accessible; uploaded books require ownership.
     """
+    # Access guard runs before the stream starts so the HTTP status is 403,
+    # not a mid-stream error. Uploaded books are private; Gutenberg books are
+    # not yet in the DB at this point so no check is needed for them.
+    _pre_cached = await get_cached_book(book_id)
+    if _pre_cached:
+        check_book_access(_pre_cached, user)
 
     async def generator() -> AsyncIterator[str]:
         try:
--- a/backend/tests/test_router_books.py
+++ b/backend/tests/test_router_books.py
@@ -927,6 +927,21 @@ async def test_upload_book_retry_translation_blocked_for_non_owner(client, test_
     assert resp.status_code == 403, resp.text
 
 
+async def test_import_stream_blocked_for_non_owner_of_uploaded_book(anon_client, tmp_db):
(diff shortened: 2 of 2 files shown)


alfmunny:
## Product review — PR #384

Correct fix for a genuine access control gap. The import-stream endpoint was the one remaining chapter-level endpoint missing `check_book_access()` — PR #375 covered `/books/{id}/chapters` and related endpoints, but the streaming route was overlooked.

### Logic check

- Access check runs **before** the `StreamingResponse` is created — FastAPI returns a proper 403 rather than sending HTTP 200 headers and then failing mid-stream ✅
- Gutenberg books that aren't yet cached return `None` from `get_cached_book` — the check is skipped correctly since they're public ✅
- `check_book_access` imported at module scope (already in scope from prior endpoints) ✅

### Test

`test_import_stream_blocked_for_non_owner_of_uploaded_book` covers the regression cleanly — inserts private book as owner, hits endpoint as anonymous user, asserts 403 ✅. Existing `test_gutenberg_book_still_accessible_to_any_user` confirms the public path is unaffected ✅.

CI pending (backend + E2E). No blockers — ready to merge once green.

PR #385, opened 2026-04-22
fix: pretranslate script wrong DB column names + lazy translator init

alfmunny:
## Summary

- Fixes `scripts/pretranslate.py` querying non-existent columns `meta_json` and `text_content` — the `books` table uses `title` and `text`
- Makes translator lazy-initialized so `--dry-run` works without installing torch/transformers
- Updates test fixtures to use the correct `text` key

## Root cause

The script was written assuming a different schema, but the actual DB schema (from migration 001) uses `title` and `text`. This caused an `OperationalError: no such column: meta_json` crash on any real invocation.

## Test plan

- [x] 17 pretranslate unit tests pass
- [x] `--dry-run` on book 1952 (The Yellow Wallpaper) runs without requiring torch installed

### DIFF
--- a/backend/tests/test_pretranslate.py
+++ b/backend/tests/test_pretranslate.py
@@ -82,7 +82,7 @@ def test_get_chapters_from_confirmed_json():
         {"title": "Chapter I", "text": "First chapter text."},
         {"title": "Chapter II", "text": "Second chapter text."},
     ]
-    book = {"text_content": json.dumps({"draft": False, "chapters": chapters})}
+    book = {"text": json.dumps({"draft": False, "chapters": chapters})}
     result = pt._get_chapters(book)
     assert len(result) == 2
     assert result[0]["title"] == "Chapter I"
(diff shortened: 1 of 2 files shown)


alfmunny:
## Product review — PR #385

Correct and necessary fix — the pretranslate script has been silently broken since the schema was finalized (columns `meta_json`/`text_content` never existed; real columns are `title`/`text`). Any real invocation would have crashed immediately with `OperationalError: no such column: meta_json`.

### Logic check

- Column fix: `meta_json` → `title`, `text_content` → `text` — matches migration 001 schema ✅
- Title extraction simplified: was `json.loads(meta_json).get('title')` → now `book.get('title')` directly — correct ✅
- Lazy translator init: `--dry-run` no longer tries to import torch/transformers, making it usable in CI or on machines without ML deps ✅
- 4 test fixtures updated from `text_content` → `text` key ✅

CI still running (most checks pending). No blockers — ready to merge once green.

Issue #387, opened 2026-04-22
GET /ai/translate/cache and POST /ai/translate missing check_book_access for private uploaded books


alfmunny:
## Problem

Two endpoints in `routers/ai.py` access translation data by `book_id` without calling `check_book_access`:

### `GET /ai/translate/cache` (line ~216)
```python
async def translate_cache(book_id, chapter_index, target_language, _user):
    cached = await get_cached_translation_with_meta(book_id, chapter_index, target_language)
    # ← no check_book_access — any authenticated user can read another user's cached translation
    if cached:
        return {"paragraphs": cached["paragraphs"], ...}
```

### `POST /ai/translate` (line ~276)
```python
if req.book_id is not None and req.chapter_index is not None:
    cached = await get_cached_translation(req.book_id, req.chapter_index, tgt)
    if cached:
        return {"paragraphs": cached, "cached": True}  # ← leaked before any access check
```

## Impact
Both return translated chapter text for a private uploaded book to any authenticated user who knows the `book_id` (sequential integer). An attacker can enumerate IDs 1..N to harvest cached translations of private books.

Note: `PUT /ai/translate/cache` already has `check_book_access` (correct).

## Fix
Both endpoints need:
```python
book = await get_cached_book(book_id)
if not book:
    raise HTTPException(status_code=404)
check_book_access(book, user)
```
For `POST /ai/translate`: validate and check access BEFORE the cache lookup.

alfmunny:
Fixed in PR #388 — check_book_access added to vocabulary POST and ai translate cache endpoints.

PR #388, opened 2026-04-22
fix: check_book_access for vocabulary POST and ai translate cache endpoints

### LINKED ISSUE #386: vocabulary POST endpoint missing check_book_access for private uploaded books
## Problem

`POST /api/vocabulary` calls `get_cached_book(req.book_id)` to confirm the book exists but never calls `check_book_access`, so any authenticated user can save vocabulary entries associated with another user's private uploaded book.

```python
# routers/vocabulary.py ~L35-49
@router.post("")
async def save(req: WordSave, user: dict = Depends(get_current_user)):
    ...
    if not await get_cached_book(req.book_id):
        raise HTTPException(status_code=404, detail="Book not found")
    # ← no check_book_access here
    return await save_word(user["id"], req.word, req.book_id, ...)
```

The word is saved under the current user's ID, so it doesn't pollute the owner's vocab list, but it does:
1. Confirm private book existence (information leak — book IDs are sequential integers).
2. Allow creating vocabulary entries in an inconsistent state (user has vocab for a book they can't read).

## Fix
```python
book = await get_cached_book(req.book_id)
if not book:
    raise HTTPException(status_code=404, detail="Book not found")
check_book_access(book, user)
```

alfmunny:
## Summary
- `POST /vocabulary` was missing `check_book_access` — any authenticated user could save vocabulary entries for another user's private uploaded book
- `GET /ai/translate/cache` was missing `check_book_access` — any authenticated user could read cached chapter translations of private books
- `POST /ai/translate` had the book access check AFTER the cache lookup — cache hits leaked translation content of private books before the check fired

## Test plan
- [x] `test_save_word_blocked_for_non_owner_of_uploaded_book` — POST /vocabulary returns 403 for non-owner
- [x] `test_translate_cache_get_blocked_for_non_owner` — GET /ai/translate/cache returns 403 for non-owner
- [x] `test_translate_post_cache_hit_blocked_for_non_owner` — POST /ai/translate returns 403 for non-owner even when translation is cached
- [x] Full suite: 1007 passed

Closes #386, #387

### DIFF
--- a/backend/routers/ai.py
+++ b/backend/routers/ai.py
@@ -221,6 +221,10 @@ async def translate_cache(
     _user: dict = Depends(get_current_user),
 ):
     """Check if a translation is cached. Returns paragraphs + provider/model if yes, 404 if not."""
+    book = await get_cached_book(book_id)
+    if not book:
+        raise HTTPException(status_code=404, detail="Book not found")
+    check_book_access(book, _user)
     target_language = target_language.lower().split("-")[0]
     from services.db import get_cached_translation_with_meta
     cached = await get_cached_translation_with_meta(book_id, chapter_index, target_language)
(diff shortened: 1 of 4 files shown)


alfmunny:
## Product review — PR #388

Three access control gaps closed, all following the same pattern established by PR #375 and #384. 1007 tests pass.

### Most important fix — POST /ai/translate ordering

The pre-existing code did the cache lookup **before** the access check:

```python
# old: check cache → (maybe return cached content) → check access
if cached: return cached   # ← private content leaked here on cache hit
if not book: raise 404
check_book_access(...)
```

The fix correctly inverts this — access check first, then cache:

```python
check_book_access(book, user)   # ← 403 before cache is even consulted
if cached: return cached
```

This is the right call. A user who knows a book_id (sequential integer — trivially guessable) could have read any private book's translated chapter content as long as it was cached. The comment in the diff explains the reasoning clearly ✅

### Vocabulary and translate-cache fixes

Both are straightforward — `get_cached_book` + `check_book_access` added before the main handler logic, same pattern as every other endpoint ✅. The translate-cache endpoint also now returns a proper 404 when the book isn't found (previously missing).

### Test fixes to existing tests

Several existing tests needed `save_book()` added because the new code path now calls `get_cached_book` before the cache lookup. These are correct test fixes — the tests were relying on the absence of the book check ✅

### One minor observation

`_insert_private_book` / `_insert_private_book_ai` helpers are now duplicated across `test_router_vocabulary.py`, `test_router_ai.py`, and `test_router_books.py`. Not a blocker — worth a future consolidation into a `conftest.py` fixture.

CI: backend ✅, frontend ✅, Docker ✅ — E2E still pending. Ready to merge once E2E green.

PR #390, opened 2026-04-22
fix: check_book_access for PUT /user/reading-progress

### LINKED ISSUE #389: PUT /user/reading-progress/{book_id} missing check_book_access for private uploaded books
## Problem

`PUT /api/user/reading-progress/{book_id}` checks book existence (`get_cached_book`) but never calls `check_book_access`, so any authenticated user can record reading progress for another user's private uploaded book.

```python
# routers/user.py ~L60-71
@router.put("/reading-progress/{book_id}")
async def update_reading_progress(book_id, req, user):
    if not await get_cached_book(book_id):
        raise HTTPException(status_code=404, detail="Book not found")
    # ← no check_book_access here
    await upsert_progress_and_log_event(user["id"], book_id, req.chapter_index)
```

The progress is stored under the current user's ID so it doesn't affect the owner, but it:
1. Confirms private book existence via sequential book_id enumeration.
2. Creates inconsistent state — user has reading progress for a book they cannot access.

## Fix
```python
book = await get_cached_book(book_id)
if not book:
    raise HTTPException(status_code=404, detail="Book not found")
check_book_access(book, user)
```

alfmunny:
## Summary
- `PUT /user/reading-progress/{book_id}` was missing `check_book_access` — any authenticated user could record reading progress and reading history for another user's private uploaded book
- Added book existence check and access guard before `upsert_progress_and_log_event`

## Test plan
- [x] `test_update_reading_progress_blocked_for_non_owner` — PUT returns 403 for non-owner
- [x] Full suite: 1009 passed

Closes #389

### DIFF
--- a/backend/routers/user.py
+++ b/backend/routers/user.py
@@ -1,7 +1,7 @@
 from typing import Optional
 from fastapi import APIRouter, Depends, HTTPException
 from pydantic import BaseModel
-from services.auth import get_current_user, encrypt_api_key, decrypt_api_key
+from services.auth import get_current_user, encrypt_api_key, decrypt_api_key, check_book_access
 from services.db import (
     set_user_gemini_key, get_user_by_id, get_reading_progress,
     get_obsidian_settings, update_obsidian_settings, get_cached_book,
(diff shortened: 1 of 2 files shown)


alfmunny:
## Product review — PR #390

Same pattern, same quality — minimal diff, single responsibility, 1009 tests passing.

- `check_book_access` added after book existence check, before `upsert_progress_and_log_event` ✅
- Regression test inserts a private book as a different owner, calls as non-owner, asserts 403 ✅
- The privacy concern here is mild (writing progress rather than reading content) but correct to guard — a user shouldn't be able to tell they've accessed a private book at all, and reading history rows for another user's book are a data integrity issue too

One note on test structure: `_insert_private_book_user` is the 4th copy of this helper across test files (also in `test_router_books.py`, `test_router_ai.py`, `test_router_vocabulary.py`). Still not a blocker, but at this point it's worth a follow-up issue to consolidate into a `conftest.py` fixture.

CI pending. No blockers.

Issue #341, opened 2026-04-22
PUT /ai/translate/cache missing running-row guard — worker overwrites user's saved translation


alfmunny:
## Bug

\`PUT /ai/translate/cache\` (used by the reader frontend to save completed progressive translations) does not check whether a queue worker is currently translating the same chapter. Both paths call \`save_translation\` (INSERT OR REPLACE), so whichever finishes last silently wins.

### Scenario

1. Queue worker picks up chapter 5 for book 1 → status='running'
2. User opens the reader page for that chapter; reader doesn't have a cached translation yet
3. Frontend does a quick Google-Translate progressive translation and saves via PUT /translate/cache
4. User's translation is stored (INSERT OR REPLACE)
5. Queue worker finishes (Gemini batch, higher quality) → \`save_translation\` → INSERT OR REPLACE → overwrites user's translation

The reverse is equally wrong: user's frontend translation can overwrite the queue's translation if the user saves after the worker finishes but before \`mark_queue_row_done\` cleans up the row.

## Root cause

The same running-row guard that was added to \`POST /admin/retranslate\` (#334) is absent from \`PUT /ai/translate/cache\`. Both write via \`save_translation\`.

## Fix

Add a running-row check identical to the admin retranslate fix:

```python
async with aiosqlite.connect(DB_PATH) as db:
    async with db.execute(
        "SELECT 1 FROM translation_queue "
        "WHERE book_id=? AND chapter_index=? AND target_language=? AND status='running'",
        (book_id, chapter_index, target_language),
    ) as cur:
        if await cur.fetchone():
            raise HTTPException(
                status_code=409,
                detail="A translation job is currently running for this chapter.",
            )
```

PR #396, opened 2026-04-22
fix: POST /admin/translations/import missing running-job guard

### LINKED ISSUE #395: POST /admin/translations/import missing running-job guard
## Problem

`POST /api/admin/translations/import` calls `save_translation()` (INSERT OR REPLACE) for each entry without checking if a queue worker has any of those chapters at status='running'.

```python
# routers/admin.py ~L596-608
for entry in req.entries:
    if not entry.paragraphs:
        continue
    await save_translation(...)  # ← no running-job check
```

## Race

1. Worker is translating book 42, chapter 5 (status='running')
2. Admin imports pre-translated version of book 42, chapter 5
3. Admin's import saves (INSERT OR REPLACE) the imported translation
4. Worker finishes → another INSERT OR REPLACE overwrites the admin's imported translation

The admin's seeded translation is silently lost.

## Pattern

Same fix as applied to retranslate (#334), retranslate-all (#334), delete_book_translations (#339):
```python
# Before the save loop, check for running workers:
for entry in req.entries:
    if not entry.paragraphs: continue
    status = await queue_status_for_chapter(entry.book_id, entry.chapter_index, entry.target_language...)
    if status['status'] == 'running':
        raise HTTPException(409, f'A translation job is running for book {entry.book_id} chapter {entry.chapter_index}. Wait for it to finish.')
```

alfmunny:
## Summary
- `POST /admin/translations/import` was missing a running-job guard: if any of the entries being imported had a queue worker running for the same (book, chapter, language), the worker's `INSERT OR REPLACE` would silently overwrite the admin's imported translation when it finished — same race as retranslate (#334) and PUT /ai/translate/cache (#341)
- Added pre-check loop that rejects with 409 if any chapter has status='running'

## Test plan
- [x] `test_import_translations_rejects_409_when_chapter_is_running` — returns 409 and leaves existing translation untouched
- [x] Full suite: 1010 passed

Closes #395

### DIFF
--- a/backend/routers/admin.py
+++ b/backend/routers/admin.py
@@ -593,6 +593,25 @@ async def import_translations(
         if not await get_cached_book(bid):
             raise HTTPException(status_code=404, detail=f"Book {bid} not found")
 
+    # Pre-check: reject if any chapter to be imported has a running queue job.
+    # The worker uses INSERT OR REPLACE and would overwrite the imported translation
+    # when it finishes — same race as retranslate (#334) and PUT /translate/cache (#341).
+    from services.translation_queue import queue_status_for_chapter
+    for entry in req.entries:
+        if not entry.paragraphs:
+            continue
+        lang = entry.target_language.lower().split("-")[0]
+        status = await queue_status_for_chapter(entry.book_id, entry.chapter_index, lang)
+        if status["status"] == "running":
+            raise HTTPException(
+                status_code=409,
+                detail=(
+                    f"A translation job is currently running for book {entry.book_id} "
+                    f"chapter {entry.chapter_index} ({lang}). "
+                    "Wait for it to finish before importing."
+                ),
+            )
+
     count = 0
     for entry in req.entries:
         if not entry.paragraphs:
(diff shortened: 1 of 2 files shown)


alfmunny:
## Product review — PR #396

Correct and consistent fix. The queue state machine guard is now applied to the last remaining write endpoint that was missing it: `POST /admin/translations/import`.

### Guard placement

The pre-check loop runs over **all entries before any write** — if any chapter is running, the entire request is rejected with 409. This is the right call: fail fast before touching the DB, so no partial import state is possible. Compare with the guard in PR #341 (PUT /translate/cache) and PR #394 (POST /ai/translate) which both check before the save call.

### Language normalisation

`entry.target_language.lower().split("-")[0]` mirrors what `queue_status_for_chapter` expects ✅ — same normalization as the rest of the translate endpoints.

### Test

Verifies both failure response (409) AND that the existing translation is unchanged after the rejected import — exactly the right assertion for a guard test ✅.

CI pending. No blockers — ready to merge once green.

PR #402, opened 2026-04-22
fix: GET /insights and GET /annotations missing check_book_access

### LINKED ISSUE #397: GET /insights and GET /annotations missing check_book_access — private book data leak
## Bug

Two GET list endpoints accept `book_id` but do not call `check_book_access`, allowing any authenticated user to enumerate insights and annotations on private uploaded books they do not own.

### Affected endpoints

| Endpoint | File | Line |
|---|---|---|
| `GET /api/insights?book_id=N` | `routers/insights.py` | 44–46 |
| `GET /api/annotations?book_id=N` | `routers/annotations.py` | 51–53 |

### Pattern

The corresponding POST endpoints already call `check_book_access` (insights.py:28, annotations.py:35), but the GET list endpoints have no such guard.

### Impact

A user with a valid JWT and knowledge of an uploaded book's sequential integer ID can call these endpoints to discover whether insights or annotations exist for private books they do not own.

Since results are filtered by `user_id`, they can only see their own entries — but the endpoint should also return 404 if the book doesn't exist, and 403 if it's a private uploaded book the caller doesn't own, consistent with all other endpoints that take `book_id`.

### Fix

Add `get_cached_book` + `check_book_access` to both GET endpoints (same pattern used in POST endpoints).

alfmunny:
## Summary

Both `GET /insights?book_id=N` and `GET /annotations?book_id=N` accepted a `book_id` query parameter but did not call `check_book_access`, allowing any authenticated user to enumerate insights and annotations on private uploaded books they do not own.

The corresponding POST endpoints already had the guard — this makes GET consistent.

## Changes

- `routers/insights.py`: added `get_cached_book` + `check_book_access` to `list_insights`
- `routers/annotations.py`: added `get_cached_book` + `check_book_access` to `list_annotations`
- Tests: two new 403 regression tests and two new 404 tests for missing books

Closes #397

### DIFF
--- a/backend/routers/annotations.py
+++ b/backend/routers/annotations.py
@@ -50,6 +50,10 @@ async def list_all_annotations(user: dict = Depends(get_current_user)):
 
 @router.get("")
 async def list_annotations(book_id: int, user: dict = Depends(get_current_user)):
+    book = await get_cached_book(book_id)
+    if not book:
+        raise HTTPException(status_code=404, detail="Book not found")
+    check_book_access(book, user)
     return await get_annotations(user["id"], book_id)
 
 
--- a/backend/routers/insights.py
+++ b/backend/routers/insights.py
@@ -43,6 +43,10 @@ async def list_all_insights(user: dict = Depends(get_current_user)):
 
 @router.get("")
 async def list_insights(book_id: int, user: dict = Depends(get_current_user)):
+    book = await get_cached_book(book_id)
+    if not book:
+        raise HTTPException(status_code=404, detail="Book not found")
+    check_book_access(book, user)
     return await get_insights(user["id"], book_id)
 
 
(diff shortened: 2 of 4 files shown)


alfmunny:
## Product review — PR #402

Correct consistency fix. The POST endpoints for both insights and annotations already had `check_book_access` — the corresponding GET list endpoints were overlooked.

- `GET /insights?book_id=N`: adds 404 guard + `check_book_access` before `get_insights` ✅
- `GET /annotations?book_id=N`: same pattern ✅
- Both also gain a 404 on missing book, which was previously absent from these GETs

Four new tests: 403 for non-owner and 404 for missing book on each endpoint. The 404 tests are worth having separately since the previous code would silently return an empty list for a non-existent book_id. ✅

Frontend ✅, Docker ✅, backend + E2E pending. No blockers — ready to merge once green.

Issue #401, opened 2026-04-22
fix: TTS word highlight stops at sentence 5, subsequent text greys out


alfmunny:
## Bug

When reading a chapter with TTS (e.g. https://book-reader-ai.vercel.app/reader/84), the amber word-highlighting stops at approximately the 5th sentence. TTS audio continues playing, but the remaining text is greyed out and no longer highlights the active sentence.

## Root cause

`SentenceReader.tsx` `parseIntoSegments` step 2a (`segmentChunkIdx`) matches each sentence to a chunk using `chunkText.indexOf(seg, cursor)`. If a sentence fails to match (returns -1) — due to whitespace differences between the frontend's `splitSentences` output and the backend's `chunk_text` output (leading spaces, double-spaces from `\n\n` → `  ` normalization, etc.) — that sentence gets `chunkIdx = -1`.

Segments with `chunkIdx = -1` are treated as **never loaded** in `loadedChunkIndices` and are rendered grey (`text-stone-400 cursor-default`). They also get `startTime = Infinity`, so `currentIdx` never advances past the last correctly-matched sentence.

The "5th sentence" pattern matches the backend `TTS_CHUNK_CHARS = 400` limit: chunk 0 typically contains ~5 sentences. If any sentence in chunk 1+ fails `indexOf` matching, the cursor/chunkIdx state corrupts and subsequent sentences go unmatched.

## Fix

Make the `segmentChunkIdx` matching more robust:

1. **Normalize both sides** before comparing: collapse multiple whitespace to a single space in both the segment text and the chunk text (not just newlines → spaces).
2. **Fallback to character-position mapping**: if `indexOf` fails, estimate which chunk a segment belongs to using the cumulative character count across paragraphs (since the backend chunks preserve paragraph boundaries). This avoids the cursor-corruption cascade.
3. Remove `cursor-default` from segments with `chunkIdx === -1` when `disabled === false` — clicking should still work even if timing is imprecise.

**Repro**: Reader page for book 84 (Frankenstein), click Play on any chapter.

alfmunny:
Fixed by PR #405.

Issue #399, opened 2026-04-22
fix: click while TTS paused highlights wrong sentence (line above)


alfmunny:
## Bug

When TTS is **paused**, clicking a sentence often highlights the sentence visually above where the user clicked.

## Root cause

`SentenceReader.tsx` line 697: clicks fire `onSegmentClick(seg.startTime, ...)` when `isPlaying || duration > 0`. When paused, `duration > 0` so clicks are handled. The seek to `seg.startTime` updates `currentTime`, and `currentIdx` (line 459) finds the last segment with `startTime ≤ currentTime`.

The likely cause: when a sentence spans multiple visual lines, clicking on a lower visual line still belongs to the same segment. However, the TTS startTime for that segment may be slightly imprecise (off by a word boundary), causing `currentIdx` to land on the previous segment. Alternatively, clicking on a segment that wraps to two lines highlights the whole segment, which visually "starts" on the line above the click.

A secondary cause: when paragraph focus (`para-dim` / `para-active`) is active with TTS paused, the opacity stacking context can shift visual line positions.

## Fix

- Investigate whether startTime accuracy is sufficient when paused (check word boundary assignment in `parseIntoSegments` step 2b)
- Ensure `currentIdx` only updates from `currentTime` changes, not from stale values when seeking while paused
- When clicking a segment, scroll it into view immediately so the user sees confirmation of which sentence was chosen

alfmunny:
Fixed by PR #405.

PR #337, opened 2026-04-22
feat: landing hero for first-time visitors

alfmunny:
## Summary

- Adds a landing hero section to the home page, visible only to unauthenticated visitors
- Hero includes: headline, bilingual translation preview (Faust DE→EN), three feature cards (AI Translation, Vocabulary Builder, AI Reading Insights), secondary feature pills (Summaries, TTS, Annotations, Reading Stats), and two CTAs ("Sign in free" + "Browse library")
- Section renders inside the existing Discover tab above the search bar — no routing changes, no impact on authenticated UX
- Updated `HomePage.test.tsx` to handle multiple "Sign in" buttons (header + hero); two tests tightened to exact-match "Sign in" vs regex

## Test plan

- [ ] Visit `/` while signed out → hero section is visible with translation preview
- [ ] "Sign in free" button navigates to `/login`
- [ ] "Browse library →" smoothly scrolls to the search section
- [ ] Sign in → hero is no longer shown
- [ ] All 1,528 frontend tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #404, opened 2026-04-22
fix: TypographyPanel invisible in focus mode + Highlight uses full sentence (#398, #400)

### LINKED ISSUE #398: fix: TypographyPanel invisible in focus mode (opacity-0 header clips fixed child)
## Bug

When **focus mode** is enabled, clicking **Aa** in the floating HUD opens the TypographyPanel but the panel is invisible and unclickable.

## Root cause

`TypographyPanel` is rendered inside `<header>`, which in focus mode gets `opacity-0 overflow-hidden`. Although `TypographyPanel` uses `position: fixed`, fixed children of an `opacity < 1` parent are still painted through the parent's opacity group — so they inherit the zero opacity and become both invisible and unreachable.

**File:** `frontend/src/app/reader/[bookId]/page.tsx` lines 961-976 (inside the header element that gets `opacity-0` in focus mode).

## Fix

Move the `{showTypographyPanel && <TypographyPanel ... />}` render outside the `<header>` element to the page root level so it is not affected by the header's opacity.

## Bonus: cursor turns to arrow in paragraph focus mode

When TTS is playing and future chunks are not yet synthesized, `segClass` applies `cursor-default` to unloaded segments (`SentenceReader.tsx` line 623). While TTS audio is playing (not disabled), unloaded future chunks get `cursor-default` (arrow cursor) instead of the normal text cursor, which is confusing especially when paragraph focus dims the rest of the page.

Fix: only apply `cursor-default` when the segment is both unloaded AND `disabled` is true.

### LINKED ISSUE #400: fix: SelectionToolbar Highlight saves selected text as sentence_text, highlighting entire sentence
## Bug

When the user **selects a phrase** (e.g. "the monster") and clicks **Highlight**, the annotation is saved with `sentence_text = "the monster"` (just the selection). `SentenceReader` then checks if any segment text **contains** the annotation's sentence_text (`segText.includes(a.sentence_text)`), so the **entire sentence** gets highlighted amber — not just the selected phrase.

## Root cause

`SelectionToolbar.tsx` line 102: `handleAction` passes `selection.text` (the raw selection) to `onHighlight`. The reader page's handler stores this as `sentenceText` in `QuickHighlightPanel`, which saves it as `sentence_text` in the DB.

`SentenceReader.tsx` `getAnnotation` (line ~562) does substring matching: if a segment contains the stored `sentence_text`, the whole segment is colored.

## Fix

Change `handleAction(onHighlight)` to pass `selection.context` (returned by `extractContext`, which walks the DOM to the nearest `[data-seg]` element and returns the full sentence text) instead of `selection.text`. This stores the full sentence as `sentence_text`, so the highlight correctly colors exactly one sentence — consistent with how TTS sentence highlighting works.

This is the correct fix within the current annotation architecture (annotations are sentence-granularity). True sub-sentence phrase highlighting would require a separate range-annotation system.

alfmunny:
## What changed

**Fix #398 — TypographyPanel invisible in focus mode**

The TypographyPanel was rendered inside the `<header>` element. Focus mode applies `opacity-0` to the header, which creates a CSS compositing layer that propagates zero opacity to all descendants — including fixed-position children like the panel. Moving the panel outside the header entirely fixes this.

**Fix #400 — Highlight stores full sentence, not raw selection**

When a user highlights a word or phrase, the toolbar was passing the raw selected text to `onHighlight`. Annotations are matched sentence-by-sentence at render time, so storing a partial string caused mismatches (no sentence highlighted, or wrong sentence colored). Now `handleHighlight` passes `selection.context` (the full surrounding sentence extracted by `extractContext`) so the stored text always matches exactly one sentence span.

## Files changed

- `frontend/src/app/reader/[bookId]/page.tsx` — moved `<TypographyPanel>` render from inside `<header>` to page root level
- `frontend/src/components/SelectionToolbar.tsx` — new `handleHighlight()` that passes `context` instead of raw selection text

## Test plan

- [ ] Open reader → enable focus mode → click Typography button → panel appears (was invisible before)
- [ ] Select a partial word/phrase → click Highlight → full sentence is highlighted (not partial match)
- [ ] Select full sentence → click Highlight → sentence highlighted correctly
- [ ] Existing annotation tests pass: `pytest backend/tests/test_router_annotations.py`

Closes #398
Closes #400

### DIFF
--- a/frontend/src/app/reader/[bookId]/page.tsx
+++ b/frontend/src/app/reader/[bookId]/page.tsx
@@ -958,22 +958,8 @@ export default function ReaderPage() {
             >
               <span className="font-serif">Aa</span>
             </button>
-            {showTypographyPanel && (
-              <TypographyPanel
-                fontSize={fontSize}
-                lineHeight={lineHeight}
-                contentWidth={contentWidth}
-                fontFamily={fontFamily}
-                paragraphFocus={paragraphFocus}
-                onFontSize={setFontSize}
-                onLineHeight={setLineHeight}
-                onContentWidth={setContentWidth}
-                onFontFamily={setFontFamily}
-                onParagraphFocus={setParagraphFocus}
-                onClose={() => setShowTypographyPanel(false)}
-                anchorPos={typographyAnchorPos ?? undefined}
-              />
-            )}
+{/* TypographyPanel rendered at page root — NOT inside the header,
+              so it is unaffected by the header's opacity-0 in focus/immersive mode. */}
           </div>
 
           {/* Theme — desktop only */}
@@ -1189,6 +1175,24 @@ export default function ReaderPage() {
           )}
         </div>
       </header>
+
+      {/* TypographyPanel is outside the header so focus-mode opacity-0 doesn't hide it (#398) */}
+      {showTypographyPanel && (
+        <TypographyPanel
+          fontSize={fontSize}
+          lineHeight={lineHeight}
+          contentWidth={contentWidth}
+          fontFamily={fontFamily}
+          paragraphFocus={paragraphFocus}
+          onFontSize={setFontSize}
+          onLineHeight={setLineHeight}
+          onContentWidth={setContentWidth}
+          onFontFamily={setFontFamily}
+          onParagraphFocus={setParagraphFocus}
+          onClose={() => setShowTypographyPanel(false)}
+          anchorPos={typographyAnchorPos ?? undefined}
+        />
+      )}
 
       {/* ── Banners (hidden in immersive mode on mobile) ──────────────── */}
       <div className={`shrink-0 transition-all duration-300 ${
--- a/frontend/src/components/SelectionToolbar.tsx
+++ b/frontend/src/components/SelectionToolbar.tsx
@@ -104,6 +104,15 @@ export default function SelectionToolbar({ onRead, onHighlight, onNote, onChat, 
     setSelection(null);
   }
 
+  function handleHighlight(fn?: (text: string) => void) {
+    if (!fn || !selection) return;
+    // Annotations are sentence-granularity — store the full sentence (context)
+    // so the highlight colors exactly one sentence, not a partial substring match (#400).
+    fn(selection.context || selection.text);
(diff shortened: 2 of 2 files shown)


alfmunny:
## Product review — PR #404

Both fixes are correct. First frontend bugs of the session.

### Fix #398 — TypographyPanel in focus mode

The root cause is right: CSS `opacity` on a parent creates a stacking context that propagates to **all** descendants — including `position: fixed` children. Moving the panel to page root level is the clean fix. The comment left where the panel used to be is helpful context for the next reader.

### Fix #400 — Highlight stores full sentence

Correct. `extractContext` already computes the full surrounding sentence. Using `selection.context` for highlight annotations is the right data — annotations are matched sentence-by-sentence at render time, so passing the partial selection string would cause mismatches. The fallback `|| selection.text` handles the edge case where context is empty.

### One gap: no automated test for the highlight logic

The SelectionToolbar change is a logic fix (not just styling), and it should have a Jest test asserting that `onHighlight` receives `selection.context`, not `selection.text`, when both are available. The TypographyPanel move is pure DOM structure and is better covered by E2E. The test plan checkboxes are manual and unchecked.

Not a hard blocker for this fix, but please add a Jest unit test for `handleHighlight` before or shortly after merge — the pattern of "manual test plan, unchecked checkboxes" has bitten us before (landing hero #337 still has no E2E).

Docker ✅, all others pending. No merge blocker.

alfmunny:
## Superseded by PR #405

PR #405 () covers everything in this PR plus:
- Bug #399 — TTS seek-while-paused greyout fix (TTSControls.tsx)
- Bug #401 — SentenceReader whitespace-normalised chunk matching (SentenceReader.tsx)
- A new test file `reader-ux-bugs-398-401.test.tsx` with Jest tests for #400 and #401 (the test gap I flagged in my earlier review)

Since #405 touches the same files (`page.tsx`, `SelectionToolbar.tsx`) and is the more complete fix, please **close this PR** and let #405 carry the work forward. The two PRs will conflict if both auto-merge.

alfmunny:
Duplicate of #405 which was already merged.

PR #405, opened 2026-04-22
fix: 4 reader UX bugs — typography panel, seek highlight, highlight context, TTS greyout

### LINKED ISSUE #398: fix: TypographyPanel invisible in focus mode (opacity-0 header clips fixed child)

### LINKED ISSUE #399: fix: click while TTS paused highlights wrong sentence (line above)

### LINKED ISSUE #400: fix: SelectionToolbar Highlight saves selected text as sentence_text, highlighting entire sentence

### LINKED ISSUE #401: fix: TTS word highlight stops at sentence 5, subsequent text greys out

alfmunny:
## Summary

- **#398** — `TypographyPanel` moved outside `<header>` to page root so focus-mode `opacity-0` on the header no longer clips the fixed overlay. A `position:fixed` child inside an `opacity < 1` parent is composited through that parent and becomes invisible; moving the panel to a sibling of the header avoids this.
- **#399** — `TTSControls` `timeupdate` handler now checks `audio.paused` before updating `globalCurrentTime`. Browsers fire a single `timeupdate` after a seek even on paused audio; the rounded `audio.currentTime` reported by the hardware was slightly less than the seeked value, causing `currentIdx` to land on the segment above the clicked one.
- **#400** — `SelectionToolbar` Highlight now passes `selection.context` (full sentence extracted from nearest `data-seg` span or `<p>`) instead of `selection.text` (the selected substring), so the entire sentence is stored as `sentence_text` and the annotation underline covers the whole sentence correctly.
- **#401** — `parseIntoSegments` in `SentenceReader` adds a normalised-whitespace fallback after the exact and prefix-based `indexOf` attempts; removes `cursor-default` from unloaded-segment CSS so the cursor shows naturally.

## Test plan

- [x] New regression test file `reader-ux-bugs-398-401.test.tsx` covering bugs #400 and #401
- [x] All 1564 existing frontend tests pass

Closes #398
Closes #399
Closes #400
Closes #401

### DIFF
--- a/frontend/src/__tests__/reader-ux-bugs-398-401.test.tsx
+++ b/frontend/src/__tests__/reader-ux-bugs-398-401.test.tsx
@@ -1,0 +1,184 @@
+/**
+ * Regression tests for reader UX bugs #398–401:
+ *
+ * #398 — TypographyPanel must not be a descendant of the header (opacity-0
+ *         in focus mode would make it invisible).
+ * #399 — seekTo while paused must not let a browser timeupdate event on the
+ *         paused audio element override the seeked currentTime.
+ * #400 — SelectionToolbar Highlight must pass the full sentence context, not
+ *         just the selected substring.
+ * #401 — SentenceReader must assign correct chunkIdx even when the segment
+ *         text and chunk text differ only in whitespace.
+ */
+import React from "react";
+import { render, screen, fireEvent, act } from "@testing-library/react";
+import SelectionToolbar from "@/components/SelectionToolbar";
+import SentenceReader, { ChunkInfo } from "@/components/SentenceReader";
+
+// ── Bug #400: SelectionToolbar Highlight uses full sentence context ────────────
+
+function makeReaderEl() {
+  const el = document.createElement("div");
+  el.id = "reader-scroll";
+  document.body.appendChild(el);
+  return el;
+}
+
+describe("SelectionToolbar Highlight — bug #400", () => {
+  let readerEl: HTMLElement;
+
+  beforeEach(() => {
+    readerEl = makeReaderEl();
+    jest.spyOn(window, "getSelection").mockReturnValue(null);
+  });
+  afterEach(() => {
+    readerEl.remove();
+    jest.restoreAllMocks();
+  });
+
+  it("calls onHighlight with full sentence context when selection is inside a data-seg span", () => {
+    const onHighlight = jest.fn();
+    render(<SelectionToolbar onHighlight={onHighlight} />);
+
+    // Build: <div#reader-scroll><p><span data-seg="0">Full sentence text.</span></p></div>
+    const p = document.createElement("p");
+    const sentenceSpan = document.createElement("span");
+    sentenceSpan.setAttribute("data-seg", "0");
+    sentenceSpan.textContent = "Full sentence text.";
+    p.appendChild(sentenceSpan);
+    readerEl.appendChild(p);
+
+    // User selects just one word ("sentence") from inside the span
+    const textNode = sentenceSpan.firstChild!;
+    const range = document.createRange();
+    range.setStart(textNode, 5);  // start of "sentence"
+    range.setEnd(textNode, 13);
+
+    const mockRect = {
+      left: 100, right: 200, top: 300, bottom: 320,
+      width: 100, height: 20, x: 100, y: 300,
+      toJSON: () => ({}),
+    } as DOMRect;
+    range.getBoundingClientRect = jest.fn().mockReturnValue(mockRect);
+
+    const mockSel = {
+      toString: () => "sentence",
+      getRangeAt: () => range,
+      removeAllRanges: jest.fn(),
+    } as unknown as Selection;
+    jest.spyOn(window, "getSelection").mockReturnValue(mockSel);
+
+    act(() => { document.dispatchEvent(new Event("selectionchange")); });
+
+    fireEvent.click(screen.getByRole("button", { name: /Highlight/i }));
+
+    // Must receive the full sentence, not just "sentence"
+    expect(onHighlight).toHaveBeenCalledWith("Full sentence text.");
+  });
+
+  it("falls back to selected text when no sentence context is found", () => {
+    const onHighlight = jest.fn();
+    render(<SelectionToolbar onHighlight={onHighlight} />);
+
+    // Selection inside a plain div (no data-seg, no <p>)
+    const textNode = document.createTextNode("just a word");
+    const span = document.createElement("span");
+    span.appendChild(textNode);
+    readerEl.appendChild(span);
+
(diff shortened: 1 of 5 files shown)


alfmunny:
## Product review — PR #405

This is the complete version of what PR #404 started. All four fixes look correct.

**#398 — TypographyPanel outside header**: Same fix as #404 — CSS compositing through opacity parent, correct root-level placement ✅

**#399 — TTS seek-while-paused**: Checking `audio.paused` before updating `globalCurrentTime` in the `timeupdate` handler is correct. Browsers fire one `timeupdate` after a seek on paused audio; the hardware-rounded timestamp would land the segment indicator one sentence above the seeked position ✅

**#400 — Highlight stores context**: Same fix as #404, now with two Jest tests asserting: (1) full sentence passed from `data-seg` span, (2) fallback to selected text when no context available ✅

**#401 — SentenceReader whitespace normalization**: Adding a normalised-whitespace fallback after exact and prefix-based `indexOf` attempts catches the case where chunk text joined with `\n\n` becomes double-space after replace. Also removes `cursor-default` from unloaded segments ✅

**Tests**: New `reader-ux-bugs-398-401.test.tsx` covers #400 (both paths) and #401 (double-space and short-sentence cases) — exactly the test gap I noted in #404's review. 1564 existing tests passing.

Please close PR #404 — it overlaps on `page.tsx` and `SelectionToolbar.tsx` and would conflict if both merged.

CI pending. No blockers — ready to merge once green.

PR #408, opened 2026-04-22
fix: delete_user missing book_uploads cascade (closes #403)

### LINKED ISSUE #403: delete_user leaves orphaned book_uploads rows
## Bug

`services/db.py` `delete_user()` deletes from vocabulary, annotations, book_insights, user_reading_progress, reading_history, and users — but does NOT delete from `book_uploads`.

The `book_uploads` table has:
```sql
user_id INTEGER NOT NULL REFERENCES users(id) ON DELETE CASCADE
```
But SQLite FK enforcement is OFF, so `ON DELETE CASCADE` never fires. After a user is deleted, their `book_uploads` rows remain, leaving a dangling `user_id` reference. This:
- Inflates upload quota counts if queried
- Leaves orphaned rows that can never be cleaned up through normal API use

### Fix

Add `DELETE FROM book_uploads WHERE user_id = ?` to `delete_user()` before deleting from `users`.

alfmunny:
## Summary
- `delete_user()` in `services/db.py` did not delete from `book_uploads` when a user was removed
- SQLite FK enforcement is OFF so `ON DELETE CASCADE` on `book_uploads.user_id` never fires
- Orphaned `book_uploads` rows remained after user deletion, inflating upload quota counts

## Fix
Added `DELETE FROM book_uploads WHERE user_id = ?` to `delete_user()` before deleting from `users`.

## Test plan
- `test_delete_user_removes_book_uploads`: inserts a user with a `book_uploads` row, deletes the user via admin API, asserts zero `book_uploads` rows remain
- Full backend suite: 1013 passed

### DIFF
--- a/backend/services/db.py
+++ b/backend/services/db.py
@@ -285,6 +285,7 @@ async def delete_user(user_id: int) -> None:
         await db.execute("DELETE FROM book_insights WHERE user_id = ?", (user_id,))
         await db.execute("DELETE FROM user_reading_progress WHERE user_id = ?", (user_id,))
         await db.execute("DELETE FROM reading_history WHERE user_id = ?", (user_id,))
+        await db.execute("DELETE FROM book_uploads WHERE user_id = ?", (user_id,))
         await db.execute("DELETE FROM users WHERE id = ?", (user_id,))
         await db.commit()
 
--- a/backend/tests/test_router_admin.py
+++ b/backend/tests/test_router_admin.py
@@ -218,6 +218,42 @@ async def test_delete_user_removes_reading_history(admin_client, admin_db, admin
     assert count == 0, "orphaned reading_history rows left after delete_user"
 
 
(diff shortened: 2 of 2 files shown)


alfmunny:
## Product review — PR #408

Correct fix for another instance of the delete-cascade pattern. `delete_user()` had the same gap as `admin.delete_book` (fixed in PR #394) — `book_uploads` was not included in the explicit DELETE sequence despite SQLite FK enforcement being OFF.

- One-line fix in the right place in `delete_user()`, before the `users` row is deleted ✅
- Test creates an upload row, calls admin delete endpoint, asserts zero orphaned rows ✅
- 1013 tests passing ✅

Backend ✅, Frontend ✅, E2E pending. **Note: PR #409 is a duplicate of this PR with identical code change — recommend closing #409 in favour of this one.**

Ready to merge once E2E green.

PR #409, opened 2026-04-22
fix: delete_user now removes book_uploads rows (closes #403)

### LINKED ISSUE #403: delete_user leaves orphaned book_uploads rows

alfmunny:
## What changed

`delete_user()` in `services/db.py` now includes an explicit `DELETE FROM book_uploads WHERE user_id = ?` before deleting the user row.

## Why

SQLite `PRAGMA foreign_keys` is OFF, so `ON DELETE CASCADE` on `book_uploads.user_id` never fires. Without the explicit delete, a deleted user's `book_uploads` rows persist as orphaned references — inflating upload quota counts and being unreachable through any normal API call.

## Test

Added `test_delete_user_removes_book_uploads` in `test_router_admin.py` — it inserts a `book_uploads` row for a user, calls the admin delete endpoint, and asserts the row is gone. The test failed before the fix and passes after.

Closes #403

### DIFF
--- a/backend/services/db.py
+++ b/backend/services/db.py
@@ -285,6 +285,7 @@ async def delete_user(user_id: int) -> None:
         await db.execute("DELETE FROM book_insights WHERE user_id = ?", (user_id,))
         await db.execute("DELETE FROM user_reading_progress WHERE user_id = ?", (user_id,))
         await db.execute("DELETE FROM reading_history WHERE user_id = ?", (user_id,))
+        await db.execute("DELETE FROM book_uploads WHERE user_id = ?", (user_id,))
         await db.execute("DELETE FROM users WHERE id = ?", (user_id,))
         await db.commit()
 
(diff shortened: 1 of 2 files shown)


alfmunny:
## Duplicate of PR #408

Both PRs make the identical one-line change to `delete_user()` in `services/db.py` (`DELETE FROM book_uploads WHERE user_id = ?`). PR #408 was opened first and has backend + frontend CI already passing.

Please **close this PR** in favour of #408.

alfmunny:
Duplicate of #408 which was already merged.

PR #410, opened 2026-04-22
fix: add check_book_access to GET /insights and GET /annotations (#397)

### LINKED ISSUE #397: GET /insights and GET /annotations missing check_book_access — private book data leak

alfmunny:
## Summary

- `GET /api/insights?book_id=N` and `GET /api/annotations?book_id=N` lacked the `get_cached_book` + `check_book_access` guard that the corresponding POST endpoints already had
- Any authenticated user could query insights/annotations on private uploaded books they don't own; they'd only see their own entries (filtered by `user_id`) but the endpoint should 404 on unknown books and 403 on inaccessible ones
- Now consistent with all other endpoints that accept `book_id`

## Test plan

- [ ] `test_get_insights_blocked_for_non_owner_upload` — GET returns 403 for private upload not owned by caller
- [ ] `test_get_insights_returns_404_for_nonexistent_book` — GET returns 404 for missing book
- [ ] `test_get_annotations_blocked_for_non_owner_upload` — same for annotations
- [ ] `test_get_annotations_returns_404_for_nonexistent_book` — same for annotations
- [ ] Full backend suite: 1016 tests pass

Closes #397

### DIFF
--- a/backend/routers/annotations.py
+++ b/backend/routers/annotations.py
@@ -50,6 +50,10 @@ async def list_all_annotations(user: dict = Depends(get_current_user)):
 
 @router.get("")
 async def list_annotations(book_id: int, user: dict = Depends(get_current_user)):
+    book = await get_cached_book(book_id)
+    if not book:
+        raise HTTPException(status_code=404, detail="Book not found")
+    check_book_access(book, user)
     return await get_annotations(user["id"], book_id)
 
 
--- a/backend/routers/insights.py
+++ b/backend/routers/insights.py
@@ -43,6 +43,10 @@ async def list_all_insights(user: dict = Depends(get_current_user)):
 
 @router.get("")
 async def list_insights(book_id: int, user: dict = Depends(get_current_user)):
+    book = await get_cached_book(book_id)
+    if not book:
+        raise HTTPException(status_code=404, detail="Book not found")
+    check_book_access(book, user)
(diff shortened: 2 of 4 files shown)


alfmunny:
Closing - issue #397 was already fixed by PR #402 which merged to main before this PR was created.

PR #420, opened 2026-04-22
fix: delete_user cascades to owned uploaded books and child data (closes #416)

### LINKED ISSUE #416: delete_user leaves orphaned uploaded books and all child data
## Bug

When an admin deletes a user via `DELETE /admin/users/{user_id}`, the function `delete_user()` in `services/db.py` deletes `book_uploads` rows (PR #408) but **does not delete the `books` rows** for uploaded books owned by that user, nor any of their child data.

## Impact

After user deletion:
- `books` rows with `owner_user_id = deleted_user_id` persist indefinitely
- Associated `translations`, `audio_cache`, `chapter_summaries`, `translation_queue`, `word_occurrences`, `annotations`, `book_insights`, `user_reading_progress`, `reading_history` rows for those books also persist
- No one can ever access those books again (owner is gone, `check_book_access` always returns 403 for non-existent owner), so they're permanent storage leaks
- Other users' annotations/insights on those books become orphaned

## Root cause

`delete_user()` lacks:
```python
# Missing book-scoped cascades for uploaded books
await db.execute("DELETE FROM translations WHERE book_id IN (SELECT id FROM books WHERE owner_user_id = ?)", (user_id,))
# ... plus audio_cache, chapter_summaries, translation_queue, word_occurrences,
#     annotations, book_insights, user_reading_progress, reading_history ...
await db.execute("DELETE FROM books WHERE owner_user_id = ?", (user_id,))
```

SQLite FK enforcement is OFF, so `ON DELETE CASCADE` on `books.owner_user_id` never fires.

## Pattern
Same SQLite FK cascade gap as PR #408 (book_uploads) and PR #235 (user_reading_progress).

alfmunny:
## What changed

Extended `delete_user()` in `services/db.py` to cascade-delete all uploaded books owned by the deleted user, along with their child data.

## Why

When a user uploads a book, a `books` row with `owner_user_id = user.id` is created. On user deletion, `PRAGMA foreign_keys` is OFF so `ON DELETE CASCADE` never fires. Without explicit DELETEs, those book rows persist indefinitely — permanently inaccessible (`check_book_access` returns 403 for a non-existent owner) but still consuming storage and polluting every associated table.

## Fix

Before deleting the `users` row, `delete_user()` now:
1. Deletes all child data for owned books: `translations`, `audio_cache`, `chapter_summaries`, `translation_queue` (non-running only), `word_occurrences`, `annotations`, `book_insights`, `reading_history`, `user_reading_progress`
2. Deletes the `books` rows themselves

Pattern matches how `admin.delete_book` handles the same cascade.

## Test

Added `test_delete_user_removes_owned_uploaded_books` in `test_router_admin.py`. The test creates an uploaded book owned by user2, deletes user2, and asserts the book row is gone. The test failed before the fix and passes after.

Closes #416

### DIFF
--- a/backend/tests/test_router_books.py
+++ b/backend/tests/test_router_books.py
@@ -942,6 +942,7 @@ async def test_import_stream_blocked_for_non_owner_of_uploaded_book(anon_client,
     )
 
 
+
 async def test_gutenberg_book_still_accessible_to_any_user(client, test_user, tmp_db):
     await save_book(9910, MOCK_META, "some text")
     resp = await client.get("/api/books/9910")


alfmunny:
Duplicate of #419 (same frontend changes, wrong title). The delete_user cascade fix for #416 is on fix/delete-user-owned-books-cascade.

PR #433, opened 2026-04-22
fix: reject out-of-bounds chapter_index with 400 (closes #429)

### LINKED ISSUE #429: POST /books/{id}/chapters/{idx}/translation enqueues out-of-bounds chapter index
## Bug

`POST /books/{id}/chapters/{chapter_index}/translation` and its `/retry` variant accept any `chapter_index` integer without validating it against the actual chapter count. An out-of-bounds index is silently enqueued, the API returns `status=pending` (success), and the queue worker later marks it `failed` with \"chapter index out of range\".

## Impact

- User receives false \"queued\" confirmation for a chapter that doesn't exist
- Queue worker wastes a slot processing impossible work
- User ends up confused seeing \"failed\" translation with no clear reason

## Root cause

Both endpoints check book existence and access control, but neither validates `chapter_index < len(chapters)`. The admin `retranslate` endpoint already does this validation (calls `split_with_html_preference` and bounds-checks).

## Fix

Call `split_with_html_preference(book_id, book_meta["text"])` in `request_chapter_translation` (the result is cached so it's essentially free for a book the user is actively reading) and raise 400 if `chapter_index >= len(chapters)`. Apply same fix to `retry_chapter_translation`.

alfmunny:
## Root cause

`POST /books/{id}/chapters/{idx}/translation` and its `/retry` variant accepted any integer `chapter_index` without validating it against the actual chapter count. An out-of-bounds index was silently enqueued, the API returned `status=pending` (false success), and the queue worker later marked it `failed` with "chapter index out of range".

## Fix

Both endpoints now call `split_with_html_preference(book_id, text)` (result is cached so it's essentially free for a book the user is actively reading) and raise `HTTP 400` if `chapter_index < 0` or `chapter_index >= len(chapters)`.

## Test plan

- `test_request_translation_out_of_bounds_chapter_returns_400` — asserts POST with `chapter_index=999` on a 2-chapter book returns 400
- `test_retry_translation_out_of_bounds_chapter_returns_400` — same for the `/retry` variant
- Full backend suite: 1022 passed

Closes #429

### DIFF
--- a/backend/tests/test_router_books.py
+++ b/backend/tests/test_router_books.py
@@ -942,6 +942,7 @@ async def test_import_stream_blocked_for_non_owner_of_uploaded_book(anon_client,
     )
 
 
+
 async def test_gutenberg_book_still_accessible_to_any_user(client, test_user, tmp_db):
     await save_book(9910, MOCK_META, "some text")
     resp = await client.get("/api/books/9910")


alfmunny:
Closing - created from wrong branch. Replacing with PR targeting fix/chapter-index-bounds-check

PR #442, opened 2026-04-22
fix: TTS greyout cascade guard, Spacing control, and Read para playback

### LINKED ISSUE #401: fix: TTS word highlight stops at sentence 5, subsequent text greys out

alfmunny:
## Summary

Three reader UX fixes on top of PR #419:

- **TTS greyout cascade guard** — when one sentence fails chunk matching (common for short cross-boundary sentences < 50 chars), restore the search checkpoint so subsequent sentences are not all grayed out (root cause of issue #401)
- **Spacing control now works** — `leading-relaxed` on `<p>`/`<div>` inside SentenceReader set `line-height` directly on those elements, overriding CSS inheritance; removing the Tailwind class lets paragraphs inherit from `.prose-reader` where the Spacing setting applies
- **"Read para" now starts playback** — `readFocusedParagraph` called `seekTo(startTime)` which only repositions audio when paused; adding `play()` after the seek starts playback from the correct position

## Test plan

- [x] Cascade guard regression test: `cascade guard: sentence spanning chunk boundary does not grey out subsequent sentences`
- [x] All 1572 frontend tests pass

Closes #401

### DIFF
--- a/frontend/src/app/reader/[bookId]/page.tsx
+++ b/frontend/src/app/reader/[bookId]/page.tsx
@@ -764,7 +764,11 @@ export default function ReaderPage() {
     const startTime = Number.isFinite(timing.startTime) ? timing.startTime : 0;
     const stopTime = Number.isFinite(timing.stopTime) ? timing.stopTime : undefined;
     setTtsStopAt(stopTime);
+    // seekTo repositions audio.currentTime and activeIndexRef synchronously
+    // (the sync part runs before any await), then play() starts playback from
+    // that position. Without play(), seekTo only repositions and stays paused.
     ttsSeekRef.current(startTime);
+    ttsControlsRef.current?.play();
   }
 
   // Obsidian export handler
--- a/frontend/src/components/SentenceReader.tsx
+++ b/frontend/src/components/SentenceReader.tsx
@@ -751,7 +751,7 @@ export default function SentenceReader({
         let originalContent: React.ReactNode;
         if (para.isVerse) {
           originalContent = (
-            <div className="font-serif text-base text-ink leading-relaxed">
+            <div className="font-serif text-base text-ink">
               {para.segments.map((seg) => (
                 <span key={seg.flatIdx} className="block">
                   {renderSeg(seg)}
@@ -761,7 +761,7 @@ export default function SentenceReader({
           );
         } else {
           originalContent = (
-            <p className="font-serif text-base text-ink leading-relaxed">
+            <p className="font-serif text-base text-ink">
               {para.segments.map((seg, sIdx) =>
                 renderSeg(seg, "", sIdx < para.segments.length - 1)
               )}


alfmunny:
**PM Review — LGTM ✅ (all CI green)**

Three clean fixes, all well-explained in the body:

1. **Cascade guard** — snapshot/restore logic is the right approach. The regression test accurately models a cross-boundary sentence and verifies subsequent sentences aren't greyed. Good coverage.

2. **`leading-relaxed` removal** — makes sense: Tailwind's `leading-relaxed` sets `line-height: 1.625` inline on the element, overriding CSS cascade from `.prose-reader`. The PR body correctly identifies this as the root cause of the Spacing control not working. Worth verifying in manual review that `.prose-reader` actually propagates the spacing CSS variable to children — but this is low-risk given the PR author clearly understands the mechanism.

3. **`play()` after seek** — minimal one-liner fix; the behaviour (seek-only stays paused) matches the described bug. No automated test for this path, which is acceptable given audio mock complexity. Filed as a follow-up observation rather than a blocker.

All checks passing. Good to merge.

PR #446, opened 2026-04-22
fix: convert extension icons from SVG to PNG (closes #443)

alfmunny:
## Summary
- Chrome extension requires PNG icons; SVG icons are silently ignored, causing a blank toolbar icon
- Used `rsvg-convert` to produce `icon16.png`, `icon48.png`, `icon128.png` from the V5-B woodcut SVG sources
- Updated `extension/manifest.json` to reference `.png` paths in both `icons` and `action.default_icon`

## Test plan
- [x] PNG files generated and committed (16×16, 48×48, 128×128)
- [x] manifest.json paths updated from `.svg` to `.png`
- Load the extension in Chrome (`chrome://extensions/ → Load unpacked`) and verify the V5-B icon appears in the toolbar

Closes #443

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #449, opened 2026-04-22
fix: validate chapter_index bounds in POST /ai/summary (closes #448)

### LINKED ISSUE #448: POST /ai/summary accepts out-of-bounds chapter_index — silently enqueues impossible work
## Summary

`POST /ai/summary` (`routers/ai.py`) accepts `chapter_index` from the request body without validating it against the actual chapter count, unlike the translation endpoints fixed in #429.

## Impact

A request with `chapter_index=999` on a 2-chapter book will silently attempt to generate a summary for a chapter that doesn't exist. The AI worker will receive an empty or invalid text, wasting API quota and potentially returning a nonsense summary.

## Reproduction

```
POST /api/ai/summary
{
  "book_id": 1,
  "chapter_index": 999,
  "target_language": "en"
}
```
Returns 200/queued instead of 400.

## Fix

After the book existence check in `request_chapter_summary`, add the same bounds guard used in the translation endpoint:
```python
from services.book_chapters import split_with_html_preference as _split
_chapters = await _split(book_id, book.get("text") or "")
if chapter_index < 0 or chapter_index >= len(_chapters):
    raise HTTPException(status_code=400, detail=f"Chapter index out of range (book has {len(_chapters)} chapter(s)).")
```

alfmunny:
## Summary
- `POST /ai/summary` accepted any `chapter_index` value without bounds checking, allowing requests for chapters that don't exist
- This wasted Gemini API quota: the worker would receive either empty or stale text and produce a nonsense summary
- Applied the same `split_with_html_preference` guard introduced for translations in #429

## Test plan
- [x] Regression test: `test_ai_summary_out_of_bounds_chapter_returns_400` — creates a 2-chapter book, posts summary request for chapter 999, asserts 400
- [x] Full AI test module passes (43 tests)

Closes #448

### DIFF
--- a/backend/routers/ai.py
+++ b/backend/routers/ai.py
@@ -161,6 +161,14 @@ async def summary(req: SummaryRequest, _user: dict = Depends(get_current_user)):
         raise HTTPException(status_code=404, detail="Book not found")
     check_book_access(book, _user)
 
+    from services.book_chapters import split_with_html_preference as _split
+    _chapters = await _split(req.book_id, book.get("text") or "")
+    if req.chapter_index < 0 or req.chapter_index >= len(_chapters):
+        raise HTTPException(
+            status_code=400,
+            detail=f"Chapter index out of range (book has {len(_chapters)} chapter(s)).",
+        )
+
     cached = await get_chapter_summary(req.book_id, req.chapter_index)
     if cached:
         return {"summary": cached["content"], "cached": True, "model": cached["model"]}
--- a/backend/tests/test_router_ai.py
+++ b/backend/tests/test_router_ai.py
@@ -688,3 +688,27 @@ async def test_translate_post_cache_hit_blocked_for_non_owner(client, test_user,
         f"got {resp.status_code}: {resp.text}"
     )
 
+
+# ── chapter_index bounds checks on /ai/summary ───────────────────────────────
+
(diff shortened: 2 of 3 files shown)


alfmunny:
**Review ✅ — LGTM**

Clean standalone fix for #448. The guard follows the same pattern as PR #434's translate/retry bounds check:
- Loads chapters via `split_with_html_preference`
- Rejects negative or out-of-range `chapter_index` with a descriptive 400
- Regression test using book ID 9881 with 2 chapters, asserting 400 on chapter_index=999

Note: PR #446 currently also has this same change bundled with the extension icon fix. Commented there to drop it — once that's cleaned up, this PR is the canonical fix for #448.

alfmunny:
**Backend failures fixed — test fixture updated**

The 5 failures in `test_router_summary.py` were caused by the new bounds check correctly rejecting `chapter_index=3` on the test book, which was saved with bare `"text"` content (1 chapter). Root cause is the same pattern as PR #434's breakage on the retry test.

Fix pushed (`4c36cb6`): added `_BOOK_TEXT` constant (4 CHAPTER-headed sections of 200 words each) and updated all 7 `save_book(BOOK_ID, _BOOK_META, ...)` calls in the file to use it. `chapter_index=3` is now in-range for the test book. The new regression test for out-of-bounds (book 9881, chapter_index=999) is unaffected.

Issue #457, opened 2026-04-22
Missing chapter_index bounds check on GET /books/{book_id}/chapters/{chapter_index}/queue-status


alfmunny:
## Bug

`GET /books/{book_id}/chapters/{chapter_index}/queue-status` accepts any `chapter_index` value without validating it against the book's actual chapter count.

## Impact

Any authenticated user can query queue status for a non-existent chapter (e.g. `chapter_index=999` on a 2-chapter book) and receive a misleading "not queued" response instead of a `400 Bad Request`. The same pattern was already fixed for annotations, insights, vocabulary, and reading-progress in #453.

## Steps to reproduce

1. Upload or fetch a book with 2 chapters (book_id=X)
2. `GET /api/books/X/chapters/999/queue-status?target_language=de`
3. Response: `{"status": null, ...}` — silent success instead of 400

## Expected behaviour

`400 Bad Request` with detail `"Chapter index out of range (book has N chapter(s))."` when `chapter_index < 0` or `chapter_index >= len(chapters)`.

## Fix

Add bounds check after `check_book_access` in `chapter_queue_status` (books.py ~line 162), same pattern as #453.

alfmunny:
This is already fixed. PR #434 (`bfc477d`) added bounds validation to `chapter_queue_status` in `backend/routers/books.py` (line 165):

```python
_chapters = await _split(book_id, book.get("text") or "")
if chapter_index < 0 or chapter_index >= len(_chapters):
    raise HTTPException(
        status_code=400,
        detail=f"Chapter index out of range (book has {len(_chapters)} chapter(s)).",
    )
```

The fix predates issue #450 and PR #453. Closing as duplicate/already fixed.

alfmunny:
Correction to my previous comment: I was reading from a local branch that already had this fix applied, not from main. On the actual main branch,  skips directly from  to  with no bounds validation. PR #459 is the correct fix — reopening.

Issue #450, opened 2026-04-22
Multiple endpoints accept out-of-bounds chapter_index: annotations, vocabulary, reading-progress, insights


alfmunny:
## Summary

After fixing `/translation` (#429) and `/ai/summary` (#448), several other endpoints still accept any `chapter_index` without validating it against the book's actual chapter count:

| Endpoint | File | Issue |
|----------|------|-------|
| `POST /annotations/` | `routers/annotations.py:26` | only checks `>= 0`, no upper bound |
| `POST /vocabulary/` | `routers/vocabulary.py:35` | no chapter_index validation at all |
| `PUT /user/reading-progress/{id}` | `routers/user.py:60` | only checks `>= 0`, no upper bound |
| `POST /insights/` | `routers/insights.py:17` | only checks `>= 0` when not None, no upper bound |

## Impact

Allows invalid data to be persisted:
- Annotations on non-existent chapters
- Word occurrences for non-existent chapters (affects vocabulary dedup and cleanup)
- Reading progress for non-existent chapters (corrupts streak/analytics)
- Insights on non-existent chapters

## Fix

Apply the same `split_with_html_preference` bounds check from the translation endpoint to each of these handlers.

PR #463, opened 2026-04-22
fix: chapter_index bounds checks on translation cache and GET translation endpoints (closes #462)

### LINKED ISSUE #462: Missing chapter_index upper bounds check on ai.py and books.py translation endpoints
## Bug

Several endpoints accept `chapter_index` but only validate the lower bound (< 0) or don't validate at all:

| Endpoint | File | Impact |
|---|---|---|
| `PUT /ai/translate/cache` | ai.py ~line 258 | Any user can store translations at non-existent chapter indices (DB corruption) |
| `POST /ai/translate` | ai.py ~line 288 | Only checks `< 0`, not upper bound — can enqueue jobs for non-existent chapters |
| `GET /books/{book_id}/chapters/{chapter_index}/translation` | books.py ~line 174 | Returns 404 on miss, but no explicit bounds check |
| `GET /ai/translate/cache` | ai.py ~line 224 | Returns 404 on miss, but no explicit bounds check |
| `DELETE /ai/summary` (admin-only) | ai.py ~line 209 | Admin can delete summaries at non-existent chapter indices |

Same pattern as #450 (annotations, insights, vocabulary, reading-progress) and #457 (queue-status).

## Expected fix

Add `split_with_html_preference` bounds check after `check_book_access` in each of these endpoints, returning `400 "Chapter index out of range (book has N chapter(s))."` when `chapter_index < 0 or chapter_index >= len(chapters)`.

alfmunny:
## Summary

Four translation endpoints accepted any `chapter_index` value without validating it against the book's actual chapter count. Same pattern as #450 (annotations, insights, vocabulary, reading-progress) and #457 (queue-status).

| Endpoint | Impact before fix |
|---|---|
| `PUT /ai/translate/cache` | Any user could write translations at non-existent chapter indices, creating orphaned DB records |
| `GET /ai/translate/cache` | Returned misleading 404 ("Not cached") instead of 400 for invalid indices |
| `POST /ai/translate` | Only checked `< 0`, missing upper bound — could trigger work on non-existent chapters |
| `GET /books/{id}/chapters/{idx}/translation` | Returned misleading 404 instead of 400 for invalid indices |

## Fix

Added `split_with_html_preference` bounds check after `check_book_access` in each endpoint, returning `400 "Chapter index out of range"` for `chapter_index < 0 or chapter_index >= len(chapters)`.

## Test plan

- [x] `test_put_translate_cache_out_of_bounds_chapter_returns_400`
- [x] `test_get_translate_cache_out_of_bounds_chapter_returns_400`
- [x] `test_post_ai_translate_out_of_bounds_chapter_returns_400`
- [x] `test_get_chapter_translation_out_of_bounds_returns_400`
- [x] Fixed `test_translate_cache_put_saves` (used `chapter_index=1` on single-chapter book)
- [x] All 47 `test_router_ai.py` tests pass
- [x] Full suite: 1051 passed + 6 pre-existing stats failures (unrelated)

Closes #462

### DIFF
--- a/backend/routers/ai.py
+++ b/backend/routers/ai.py
@@ -234,6 +234,13 @@ async def translate_cache(
         raise HTTPException(status_code=404, detail="Book not found")
     check_book_access(book, _user)
     target_language = target_language.lower().split("-")[0]
+    from services.book_chapters import split_with_html_preference as _split
+    _chapters = await _split(book_id, book.get("text") or "")
+    if chapter_index < 0 or chapter_index >= len(_chapters):
+        raise HTTPException(
+            status_code=400,
+            detail=f"Chapter index out of range (book has {len(_chapters)} chapter(s)).",
+        )
     from services.db import get_cached_translation_with_meta
     cached = await get_cached_translation_with_meta(book_id, chapter_index, target_language)
     if cached:
@@ -265,6 +272,13 @@ async def save_translate_cache(req: SaveTranslationRequest, _user: dict = Depend
         raise HTTPException(status_code=404, detail="Book not found")
     check_book_access(book, _user)
     target_language = req.target_language.lower().split("-")[0]
+    from services.book_chapters import split_with_html_preference as _split
+    _chapters = await _split(req.book_id, book.get("text") or "")
+    if req.chapter_index < 0 or req.chapter_index >= len(_chapters):
+        raise HTTPException(
+            status_code=400,
+            detail=f"Chapter index out of range (book has {len(_chapters)} chapter(s)).",
+        )
     # Reject if a queue worker is actively translating this chapter — the
     # worker's save_translation (INSERT OR REPLACE) would overwrite whatever
     # we write here when it finishes. (#341)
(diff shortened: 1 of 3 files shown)


alfmunny:
Reviewed. The fix is correct and covers the four endpoints from issue #462.

**Looks good:**
- All four endpoints get the same `split_with_html_preference` bounds check pattern placed correctly after `check_book_access`
- `POST /ai/translate` guard (`if req.book_id is not None and req.chapter_index is not None and _book is not None`) is appropriate — that endpoint can translate freeform text without a book context
- `test_translate_cache_put_saves` fix (chapter_index 1→0 with `clear_cache()`) is correct — the test was silently passing with an out-of-bounds index before
- 4 regression tests, all using `_BOUNDS_TEXT` (2-chapter deterministic text) — network-independent and consistent with the established pattern
- Backend reports 1051 passed

**One gap from #462 not covered here:** `DELETE /ai/summary` (admin-only) has no bounds check and no book-exists guard — passes any chapter_index directly to a SQL DELETE, silently succeeding with 0 rows deleted. Low severity (admin-only), but inconsistent. Filed as follow-up #464.

**No merge blocker.** Auto-merge is enabled; should land once CI completes.

PR #466, opened 2026-04-23
fix: use UTC dates in stats tests to match production streak calculation (closes #465)

alfmunny:
## Summary

Six stats tests fail consistently in UTC+ timezones (e.g. UTC+2) when run after midnight local but before midnight UTC. The test helpers used `date.today()` (machine local time) while all production code uses UTC.

| Code | Date source |
|---|---|
| `_insert_history_at` | `date.today()` (local) ← **bug** |
| Activity assertions | `date.today().isoformat()` (local) ← **bug** |
| `reading_history.read_at` DEFAULT | `CURRENT_TIMESTAMP` (UTC — SQLite) |
| Activity SQL | `DATE('now')` (UTC — SQLite) |
| Streak calc | `datetime.now(timezone.utc).date()` (UTC) |

## Fix

Changed `_insert_history_at` and two `today` assertions to `datetime.now(timezone.utc).date()`.

## Test plan

- [x] All 13 `test_router_stats.py` tests pass (were 6 failures)
- [x] No production code changed — test-only fix

Closes #465

PR #468, opened 2026-04-23
fix: save_book guards against overwriting uploaded private books (closes #467)

### LINKED ISSUE #467: save_book uses INSERT OR REPLACE which can silently delete uploaded private books
## Bug

`services/db.py::save_book()` uses `INSERT OR REPLACE INTO books` with an explicit `book_id`. Since uploaded private books are stored in the same `books` table using SQLite AUTOINCREMENT IDs, a Gutenberg book fetch can silently delete a user's uploaded private book if the auto-assigned ID happens to match a Gutenberg catalog ID.

## Scenario

1. Fresh DB — user uploads a book → gets auto-assigned `id=1`
2. Any user fetches Gutenberg book #1 (`GET /api/books/1/chapters`)
3. `save_book(1, gutenberg_meta, gutenberg_text)` executes `INSERT OR REPLACE INTO books (id=1, ...)`
4. SQLite `INSERT OR REPLACE` deletes the old row (uploaded book) and inserts a new row without `source='upload'` or `owner_user_id`
5. The user's private book content is **permanently lost** and the book becomes publicly accessible as a Gutenberg book

## Root cause

`save_book` does not check whether the target `book_id` already contains an uploaded private book before overwriting.

## Fix

Add a guard in `save_book`: skip the `INSERT OR REPLACE` (or raise) if the existing row has `source='upload'`, preventing Gutenberg data from overwriting private books.

```python
# Before the INSERT OR REPLACE:
async with db.execute("SELECT source FROM books WHERE id=?", (book_id,)) as cur:
    existing = await cur.fetchone()
if existing and existing[0] == "upload":
    return  # Never overwrite uploaded private books with Gutenberg data
```

alfmunny:
## Summary

`services/db.py::save_book()` uses `INSERT OR REPLACE INTO books` with an explicit Gutenberg `book_id`. If an uploaded private book (using SQLite AUTOINCREMENT IDs) happened to share that ID, the `INSERT OR REPLACE` would silently delete the private book and re-insert without `source='upload'` or `owner_user_id`, permanently destroying the user's content.

**Example attack path:**
1. Fresh DB — user uploads a book → gets `id=1`
2. Any user fetches Gutenberg book #1 → `save_book(1, ...)` wipes the uploaded book
3. User's private text is gone; the row now appears as a public Gutenberg book

## Fix

Added a guard before the `INSERT OR REPLACE`: if the existing row has `source='upload'`, return immediately without touching the database. This ensures Gutenberg data can never clobber private uploaded content.

## Test plan

- [x] `test_save_book_does_not_overwrite_uploaded_private_book` — confirms source, owner_user_id, and text are preserved
- [x] Full backend suite: **1063 passed, 0 failures** (stats timezone failures also resolved by #466 which merged to main)

Closes #467

### DIFF
--- a/backend/services/db.py
+++ b/backend/services/db.py
@@ -440,6 +440,17 @@ async def save_book(book_id: int, meta: dict, text: str, images: list | None = N
     will pick them up on its next tick.
     """
     async with aiosqlite.connect(DB_PATH) as db:
+        # Guard against clobbering a user's uploaded private book that happens
+        # to share the same auto-assigned SQLite ID as a Gutenberg catalog ID.
+        # INSERT OR REPLACE would delete the uploaded row (losing source='upload'
+        # and owner_user_id) and re-insert without those fields. (#467)
+        async with db.execute(
+            "SELECT source FROM books WHERE id=?", (book_id,)
+        ) as _cur:
+            _existing = await _cur.fetchone()
+        if _existing and _existing[0] == "upload":
+            return
+
         await db.execute(
             """
             INSERT OR REPLACE INTO books
--- a/backend/tests/test_db_extended.py
+++ b/backend/tests/test_db_extended.py
@@ -480,3 +480,57 @@ async def test_delete_user_removes_owned_books_and_all_child_data():
             )
             count = (await cur.fetchone())[0]
             assert count == 0, f"{table} still has rows for deleted owner's book (#{416})"
+
+
+# ── save_book upload collision guard ─────────────────────────────────────────
+
+
+async def test_save_book_does_not_overwrite_uploaded_private_book(tmp_db):
+    """Regression #467: save_book with a Gutenberg book_id that matches an
+    existing uploaded private book must be a no-op rather than wiping the
+    user's private content via INSERT OR REPLACE.
+
+    Scenario: user uploads a book (gets auto-id=X). Admin/user then fetches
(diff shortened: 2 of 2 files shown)


alfmunny:
Reviewed. This is a real data integrity bug — `INSERT OR REPLACE` deletes the old row before inserting, so `source='upload'` and `owner_user_id` are silently wiped, permanently destroying the user's private content and making it publicly accessible as a Gutenberg book.

**Fix is correct and minimal:**
- Guard runs inside the same DB connection, before the INSERT OR REPLACE
- Short-circuits on `source='upload'` only — normal Gutenberg rows (source IS NULL) are unaffected
- No callers check the return value of `save_book`, so a silent `return` is fine

**Test is thorough:**
- Manually inserts a private book at a known ID with `source='upload'`, `owner_user_id`, and private text
- Calls `save_book()` with the same ID + Gutenberg metadata
- Asserts all three fields (`source`, `owner_user_id`, `text`) are preserved

**One minor note:** The SELECT+INSERT is not atomic (TOCTOU window). In practice this is harmless here — the upload would always be established before any Gutenberg fetch that could collide — but worth noting the pattern for future reviewers. A `BEGIN IMMEDIATE` would be belt-and-suspenders, but is not required for correctness given the code paths.

**1063 passed, 0 failures** — first completely clean suite this session.

**No concerns.** Auto-merge enabled; should land once CI completes.

PR #469, opened 2026-04-23
fix: book-exists and chapter bounds check for DELETE /ai/summary

### LINKED ISSUE #464: DELETE /ai/summary missing book-exists and chapter_index bounds check
## Bug

`DELETE /ai/summary` (admin-only endpoint in `ai.py`) has no bounds check on `chapter_index` and no book-existence check.

## Current behaviour

```
DELETE /api/ai/summary?book_id=99999&chapter_index=999
→ 200 OK {"ok": true}   ← even if book doesn't exist or chapter is out of range
```

The handler checks admin role, then immediately issues a `DELETE FROM chapter_summaries WHERE book_id=? AND chapter_index=?`. SQLite silently deletes 0 rows when the record doesn't exist, so the response is always `{"ok": true}`.

## Impact

Low (admin-only). Admins receive misleading 200 for non-existent book/chapter combinations instead of 404/400. No data corruption risk since a no-op DELETE harms nothing.

## Expected fix

1. `GET /cached_book(book_id)` → 404 if book not found  
2. `check_book_access(book, user)` (access control for private books)  
3. `split_with_html_preference` bounds check → 400 if `chapter_index >= len(chapters)`

Same pattern as all other chapter-level endpoints. Mentioned in #462 but not covered by PR #463.

alfmunny:
## Summary
- `DELETE /ai/summary` returned `200 {"ok": true}` for non-existent books instead of 404
- `DELETE /ai/summary` silently deleted nothing when `chapter_index` was out of bounds instead of returning 400
- Added book-exists check (404) and chapter bounds check (400) consistent with all other summary/translation endpoints

## Test plan
- `test_delete_summary_nonexistent_book_returns_404`: confirms 404 when book_id=99999 (not in DB)
- `test_delete_summary_out_of_bounds_chapter_returns_400`: confirms 400 for chapter_index=999 on a 2-chapter book
- Full backend suite: 1065 passed

Closes #464

### DIFF
--- a/backend/routers/ai.py
+++ b/backend/routers/ai.py
@@ -210,6 +210,16 @@ async def delete_summary(book_id: int, chapter_index: int, user: dict = Depends(
     """Admin-only: delete a cached summary so it will be regenerated on next request."""
     if user.get("role") != "admin":
         raise HTTPException(status_code=403, detail="Admin only")
+    book = await get_cached_book(book_id)
+    if book is None:
+        raise HTTPException(status_code=404, detail="Book not found")
+    from services.book_chapters import split_with_html_preference as _split
+    _chapters = await _split(book_id, book.get("text") or "")
+    if chapter_index < 0 or chapter_index >= len(_chapters):
+        raise HTTPException(
+            status_code=400,
+            detail=f"Chapter index out of range (book has {len(_chapters)} chapter(s)).",
+        )
     from services.db import DB_PATH
     import aiosqlite
     async with aiosqlite.connect(DB_PATH) as db:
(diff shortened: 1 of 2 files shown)


alfmunny:
Reviewed. Exactly the right fix for #464, consistent with how every other chapter-level endpoint in `ai.py` handles validation.

**Looks good:**
- `get_cached_book` → 404 placed after the admin-role guard (correct ordering: 403 before 404)
- `split_with_html_preference` bounds check follows immediately — same pattern as `GET /ai/summary`, `POST /ai/translate`, etc.
- Two targeted regression tests: non-existent book → 404, out-of-bounds chapter → 400
- Network-independent test (inline 2-chapter text + `clear_cache()`)
- 1065 passed (2 more than the previous clean run from #468)

**No concerns.** Auto-merge enabled; should land once CI completes.

PR #471, opened 2026-04-23
fix: eliminate N+1 query in get_vocabulary

### LINKED ISSUE #470: N+1 query in get_vocabulary: one SQL round-trip per word
## Bug

`services/db.py:get_vocabulary()` fetches all vocabulary words in one query, then issues a separate SQL query for each word's occurrences (N+1 pattern).

For a user with 100 vocabulary words, this is 101 SQL queries per GET /user/vocabulary request. Under load this serializes through aiosqlite's worker thread, adding measurable latency and making the endpoint a DoS target.

## Root cause

```python
for v in vocab_rows:
    async with db.execute(
        "SELECT wo.book_id, b.title … FROM word_occurrences wo … WHERE wo.vocabulary_id = ?",
        (v["id"],),
    ) as cursor:
        occurrences = [dict(r) for r in await cursor.fetchall()]
```

## Fix

Replace with a single JOIN query that fetches all words + occurrences together:

```sql
SELECT v.id, v.word, v.lemma, v.language, v.created_at,
       wo.book_id, b.title AS book_title, wo.chapter_index, wo.sentence_text
FROM vocabulary v
LEFT JOIN word_occurrences wo ON wo.vocabulary_id = v.id
LEFT JOIN books b ON b.id = wo.book_id
WHERE v.user_id = ?
ORDER BY v.word, wo.created_at
```

Then group in Python.

alfmunny:
## Summary
- `get_vocabulary()` issued N+1 SQL queries: one to fetch all words, then one per word for its occurrences
- For a user with 100 vocabulary words this meant 101 round-trips through aiosqlite's serialized worker thread, making the endpoint slow and a DoS target
- Replaced with a single LEFT JOIN query that fetches all words + occurrences together; grouping is done in Python

## Test plan
- Added `test_get_vocabulary_multiple_words_multiple_occurrences`: verifies two words each with two occurrences, alphabetical ordering, and correct field values
- All existing vocabulary tests pass unchanged (correctness unaffected by the optimization)
- Full backend suite: 1065 passed

Closes #470

### DIFF
--- a/backend/services/db.py
+++ b/backend/services/db.py
@@ -850,31 +850,37 @@ async def get_vocabulary(user_id: int) -> list[dict]:
     async with aiosqlite.connect(DB_PATH) as db:
         db.row_factory = aiosqlite.Row
         async with db.execute(
-            "SELECT id, word, lemma, language, created_at FROM vocabulary WHERE user_id = ? ORDER BY word",
+            """SELECT v.id, v.word, v.lemma, v.language, v.created_at,
+                      wo.book_id, b.title AS book_title, wo.chapter_index, wo.sentence_text
+               FROM vocabulary v
+               LEFT JOIN word_occurrences wo ON wo.vocabulary_id = v.id
+               LEFT JOIN books b ON b.id = wo.book_id
+               WHERE v.user_id = ?
+               ORDER BY v.word, wo.created_at""",
             (user_id,),
         ) as cursor:
-            vocab_rows = await cursor.fetchall()
+            rows = await cursor.fetchall()
 
-        result = []
-        for v in vocab_rows:
-            async with db.execute(
-                """SELECT wo.book_id, b.title AS book_title, wo.chapter_index, wo.sentence_text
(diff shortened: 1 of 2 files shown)


alfmunny:
Reviewed ✅

Clean replacement — single JOIN over `vocabulary + word_occurrences + books` eliminates the N+1 pattern correctly.

A few notes on correctness:
- `LEFT JOIN` + `if row["book_id"] is not None` correctly handles words with zero occurrences without dropping them.
- `ORDER BY v.word, wo.created_at` + Python dict insertion order means the returned list is alphabetical, matching the original `ORDER BY word` behaviour.
- Sort modes added in #435 operate on the returned list in the router, so this DB-level sort change is transparent to them.

The test covers multi-word / multi-occurrence / alphabetical ordering and field presence — that's the right regression surface. The zero-occurrence (LEFT JOIN null) path isn't tested but is unreachable via `save_word` so not a blocker.

No concerns. Good to merge once CI passes.

PR #473, opened 2026-04-23
fix: reject path traversal and absolute paths in obsidian_path

### LINKED ISSUE #472: PATCH /user/obsidian-settings accepts path traversal in obsidian_path
## Bug

`PATCH /user/obsidian-settings` accepts `obsidian_path` values containing `..` path-traversal sequences without validation. When the path is used in the Obsidian export:

```python
api_url = f"https://api.github.com/repos/{repo}/contents/{path}/{filename}"
```

A path like `../../secret` causes the HTTP client to normalize the URL to `https://api.github.com/repos/{owner}/secret/{filename}`, writing the file to an unexpected path (potentially a sibling repo) in the user's GitHub account instead of their vault.

## Severity

LOW — requires the user's own GitHub token, so the only victim is the user themselves.

## Fix

Reject `obsidian_path` values that contain `..` segments or start with `/` in `PATCH /user/obsidian-settings`:

```python
if ".." in req.obsidian_path or req.obsidian_path.startswith("/"):
    raise HTTPException(status_code=400, detail="obsidian_path must not contain '..' or start with '/'")
```

alfmunny:
## Summary
- `PATCH /user/obsidian-settings` accepted `obsidian_path` values containing `..` traversal sequences or starting with `/`, without any validation
- A user could set `obsidian_path = "../../other-repo"`, causing the Obsidian export to write files to unexpected paths within their GitHub account via URL normalization
- Added validation to reject `obsidian_path` values containing `..` segments or starting with `/`
- Fixed an existing test that was incorrectly using `/p-updated` (absolute path) as a test value

## Test plan
- `test_patch_obsidian_path_traversal_rejected`: `../../malicious` → 400
- `test_patch_obsidian_absolute_path_rejected`: `/absolute/path` → 400
- `test_patch_obsidian_valid_path_accepted`: `Notes/Books/Literature` → 200
- All 23 user router tests pass

Closes #472

### DIFF
--- a/backend/routers/user.py
+++ b/backend/routers/user.py
@@ -107,6 +107,12 @@ async def patch_obsidian(
     req: ObsidianSettingsUpdate,
     user: dict = Depends(get_current_user),
 ):
+    path = req.obsidian_path.strip()
+    if path and (".." in path.split("/") or path.startswith("/")):
+        raise HTTPException(
+            status_code=400,
+            detail="obsidian_path must not contain '..' segments or start with '/'",
+        )
     # None = not supplied → skip the column entirely (avoids a stale-read race).
     # "" = explicit clear → set NULL.  non-empty = encrypt and store.
     if req.github_token is None:
--- a/backend/tests/test_router_user.py
+++ b/backend/tests/test_router_user.py
@@ -139,11 +139,11 @@ async def test_patch_obsidian_settings_preserves_token_when_not_supplied(client,
     from services.auth import encrypt_api_key, decrypt_api_key
     from services.db import update_obsidian_settings, get_obsidian_settings
 
-    await update_obsidian_settings(test_user["id"], encrypt_api_key("existing-token"), "r/v", "/p")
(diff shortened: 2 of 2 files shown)


alfmunny:
Reviewed ✅

Correct and minimal fix. A few notes:

- `".." in path.split("/")` is the right check — it rejects exact `..` segments without blocking legitimate names like `folder..name`. Catches single `..`, `../path`, `../../deep`, etc.
- `strip()` before validation avoids a whitespace-bypass (e.g. ` ..`).
- `if path and ...` correctly passes through empty-string paths (the "clear" case).
- The incidental fix to `test_patch_obsidian_settings_preserves_token_when_not_supplied` (`/p` → `p`) is correct — the old test was inadvertently using an absolute path.
- Three regression tests cover rejection of traversal, rejection of absolute paths, and acceptance of a valid relative path — the right surface.

Note: this is write-time-only validation; any `..` values stored before this fix would still be in the DB, but that's acceptable for the severity level (own-token only).

Good to merge once CI passes.

PR #476, opened 2026-04-23
fix: validate model_chain entries and filter whitespace auto_translate_languages

### LINKED ISSUE #474: PUT /admin/queue/settings accepts empty model_chain and empty-string entries
## Bug

`PUT /admin/queue/settings` accepts `model_chain` with:
1. Empty list `[]` — stores `[]` in DB but skips `set_setting(SETTING_MODEL, ...)`, leaving the primary model setting stale or unset
2. List with empty-string entries `["", "gemini-1.5-flash"]` — stores `["", "gemini-1.5-flash"]` and sets `SETTING_MODEL = ""`

Additionally, `auto_translate_languages` accepts whitespace-only strings like `[" ", "en"]` which normalize to whitespace language codes.

## Root cause

```python
# model_chain: no validation before storage
if req.model_chain is not None:
    await set_setting(SETTING_MODEL_CHAIN, json.dumps(req.model_chain))
    if req.model_chain:                          # empty list skips SETTING_MODEL update
        await set_setting(SETTING_MODEL, req.model_chain[0])  # may be ""

# auto_translate_languages: `if lang` misses whitespace-only strings
json.dumps([lang.lower().split("-")[0] for lang in req.auto_translate_languages if lang])
```

## Fix

Reject empty `model_chain` and empty/whitespace entries:
```python
if req.model_chain is not None:
    if not req.model_chain:
        raise HTTPException(status_code=400, detail="model_chain cannot be empty")
    if any(not m.strip() for m in req.model_chain):
        raise HTTPException(status_code=400, detail="model_chain entries cannot be empty strings")
    await set_setting(SETTING_MODEL_CHAIN, json.dumps(req.model_chain))
    await set_setting(SETTING_MODEL, req.model_chain[0])
```

And change `if lang` to `if lang and lang.strip()` for auto_translate_languages.

alfmunny:
## Summary
- `PUT /admin/queue/settings` accepted `model_chain: []` (empty list), storing it in DB while leaving `SETTING_MODEL` stale — the primary model setting would no longer reflect the chain head
- `model_chain` also accepted entries like `["", "gemini-1.5-flash"]` — would set `SETTING_MODEL = ""`, causing all subsequent AI calls to fail with model-not-specified errors
- `auto_translate_languages` used `if lang` which passes whitespace-only strings like `"  "`, storing invalid empty language codes after normalization
- Fixed by validating `model_chain` before storage and filtering with `lang.strip()` for auto_translate_languages

## Test plan
- `test_queue_settings_rejects_empty_model_chain`: `model_chain: []` → 400
- `test_queue_settings_rejects_model_chain_with_empty_entry`: `model_chain: ["", "gemini-1.5-flash"]` → 400
- `test_queue_settings_rejects_auto_translate_languages_with_whitespace_entry`: `["  ", "en", " "]` → 200, stored as `["en"]`
- `test_queue_settings_accepts_valid_model_chain`: `["gemini-2.0-flash", "gemini-1.5-flash"]` → 200

Closes #474

### DIFF
--- a/backend/routers/admin.py
+++ b/backend/routers/admin.py
@@ -1060,9 +1060,9 @@ async def queue_set_settings(
         await set_setting(
             SETTING_AUTO_LANGS,
             json.dumps([
-                lang.lower().split("-")[0]
+                lang.strip().lower().split("-")[0]
                 for lang in req.auto_translate_languages
-                if lang
+                if lang and lang.strip()
             ]),
         )
     if req.rpm is not None:
@@ -1072,11 +1072,14 @@ async def queue_set_settings(
     if req.model is not None:
         await set_setting(SETTING_MODEL, req.model)
     if req.model_chain is not None:
+        if not req.model_chain:
+            raise HTTPException(status_code=400, detail="model_chain cannot be empty")
+        if any(not m.strip() for m in req.model_chain):
+            raise HTTPException(status_code=400, detail="model_chain entries cannot be empty strings")
         # Keep the legacy single-model setting in sync with the chain head
         # so any code path still reading SETTING_MODEL stays consistent.
         await set_setting(SETTING_MODEL_CHAIN, json.dumps(req.model_chain))
-        if req.model_chain:
-            await set_setting(SETTING_MODEL, req.model_chain[0])
+        await set_setting(SETTING_MODEL, req.model_chain[0])
     if req.max_output_tokens is not None:
         await set_setting(SETTING_MAX_OUTPUT_TOKENS, str(req.max_output_tokens))
     queue_worker().wake()
--- a/backend/tests/test_router_admin.py
+++ b/backend/tests/test_router_admin.py
@@ -1907,3 +1907,56 @@ async def test_queue_settings_accepts_valid_positive_values(admin_client, admin_
     assert res.status_code == 200, (
         f"Expected 200 for valid positive values, got {res.status_code}: {res.text}"
     )
+
+
+# ── model_chain validation (Issue #474) ──────────────────────────────────────
+
+async def test_queue_settings_rejects_empty_model_chain(admin_client, admin_db):
+    """Regression #474: empty model_chain list must return 400 (would leave SETTING_MODEL stale)."""
+    res = await admin_client.put(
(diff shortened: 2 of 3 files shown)


alfmunny:
Reviewed ✅

Clean validation additions:
- Empty `model_chain` list rejected before writing to settings — prevents SETTING_MODEL going stale with no head to sync from.
- Empty-string/whitespace entries rejected — would have set SETTING_MODEL to `""` causing every AI call to fail with an invalid model name.
- Whitespace-only language codes in `auto_translate_languages` are silently filtered (right call — these are likely copy-paste accidents, not malicious input).
- Removing the now-dead `if req.model_chain:` guard is correct cleanup (the earlier empty-list check makes it unreachable).

4 tests cover: empty list rejection, empty-entry rejection, whitespace lang filtering with stored-value assertion, and valid chain acceptance. Good regression surface.

Good to merge once CI passes.

alfmunny:
CI failure — pre-existing test conflict ⚠️

The backend test suite fails with:

```
FAILED tests/test_router_admin_extended.py::test_queue_put_settings_empty_model_chain_does_not_update_model
assert 400 == 200
```

`test_queue_put_settings_empty_model_chain_does_not_update_model` in `test_router_admin_extended.py` was written against the old behavior (empty list silently skips the update, returns 200). The new validation correctly returns 400 — but this old test needs to be updated to match.

Fix: in `test_router_admin_extended.py`, update that test to assert `status_code == 400` (or rename it to `test_queue_settings_empty_model_chain_returns_400` to match your new naming convention). The new tests you added in `test_router_admin.py` already cover the correct behavior.

Otherwise the fix looks correct — unblock once the old test is updated.

alfmunny:
Unblocked ✅

Second commit correctly replaces the stale test:
- `test_queue_put_settings_empty_model_chain_does_not_update_model` (asserted 200, encoded old behavior) → `test_queue_put_settings_empty_model_chain_returns_400` (asserts 400, matches the new validation)
- Stale setup code (`set_setting`, model-unchanged assertion) removed — not needed for the new assertion.

CI is re-running. Good to merge once all checks pass.

PR #477, opened 2026-04-23
fix: reject empty chapters list in POST /books/{book_id}/chapters/confirm

### LINKED ISSUE #475: POST /books/{book_id}/chapters/confirm accepts empty chapters list
## Bug

`POST /books/{book_id}/chapters/confirm` accepts `chapters: []` (empty list) in the request body. When the list is empty:
- `len(body.chapters) == 0` — no chapter loop runs
- The book is confirmed but has no chapters, making it unreadable
- `split_with_html_preference` will return 0 chapters, and all chapter_index >= 0 will be out of range

## Root cause

`ConfirmChaptersBody.chapters: list[dict]` has no minimum length constraint.

## Fix

```python
from pydantic import Field

class ConfirmChaptersBody(BaseModel):
    chapters: list[dict] = Field(..., min_length=1)
```

Or validate explicitly in the handler:
```python
if not body.chapters:
    raise HTTPException(status_code=400, detail="chapters list cannot be empty")
```

alfmunny:
## Summary
- `POST /books/{book_id}/chapters/confirm` accepted `chapters: []` (empty list), responding `200 {"ok": true, "chapter_count": 0}`
- This marked the book as confirmed with 0 chapters — all subsequent chapter reads would hit out-of-range errors
- Added explicit check: `if not body.chapters: raise HTTPException(400, ...)`

## Test plan
- `test_confirm_chapters_empty_list_returns_400`: upload a book, confirm with `chapters: []` → 400
- All 21 upload router tests pass

Closes #475

### DIFF
--- a/backend/routers/uploads.py
+++ b/backend/routers/uploads.py
@@ -202,6 +202,9 @@ async def confirm_chapters(
         if not data.get("draft"):
             raise HTTPException(status_code=400, detail="Book already confirmed")
 
+        if not body.chapters:
+            raise HTTPException(status_code=400, detail="chapters list cannot be empty")
+
         orig_chapters = data.get("chapters", [])
 
         # Build final chapters: body.chapters provides the ordering/titles;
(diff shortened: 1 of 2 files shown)


alfmunny:
Reviewed ✅

Minimal correct fix. An empty chapters list would have confirmed the book with 0 chapters, making every subsequent chapter_index access return out-of-bounds 400. The guard fires early — before the chapter-build loop — so no partial state is written.

1 regression test: upload a book, confirm with `chapters=[]`, expect 400. That's the right surface for this bug.

Good to merge once CI passes.

PR #479, opened 2026-04-23
fix: hide private uploaded books from GET /books/cached

### LINKED ISSUE #478: GET /books/cached exposes private uploaded book titles to all users
## Bug

`GET /books/cached` returns ALL books from the database, including privately-uploaded books (source='upload'), without authentication. This allows any user (or anonymous client) to enumerate the titles of other users' private books.

## Root cause

```python
# routers/books.py
@router.get("/cached")
async def cached_books():
    return await list_cached_books()  # no auth, no source filter
```

```python
# services/db.py
async def list_cached_books():
    await db.execute(
        "SELECT id, title, authors, ... FROM books ORDER BY cached_at DESC"
        # Missing: WHERE source != 'upload'
    )
```

## Impact

- Severity: LOW (only titles disclosed, not content — content requires auth and ownership check)
- Any unauthenticated request can discover that certain private books exist and their titles

## Fix

Filter out uploaded books from the public cached list:
```sql
SELECT id, title, authors, languages, subjects, download_count, cover, cached_at
FROM books
WHERE (source IS NULL OR source != 'upload')
ORDER BY cached_at DESC
```

The user's own uploaded books remain accessible via their direct URL and via `GET /user/reading-progress`.

alfmunny:
## Summary
- `GET /books/cached` returned ALL books including private user uploads (source='upload'), without any authentication
- Any unauthenticated request could enumerate the titles of other users' private books
- Fixed by adding `WHERE (source IS NULL OR source != 'upload')` to `list_cached_books()` so only public Gutenberg books appear in the public library

## Test plan
- `test_cached_books_hides_uploaded_books`: inserts a private uploaded book (source='upload') and a public Gutenberg book; verifies only the Gutenberg book appears in the cached list
- All 12 `cached` tests pass

Closes #478

### DIFF
--- a/backend/services/db.py
+++ b/backend/services/db.py
@@ -675,11 +675,17 @@ async def get_user_stats(user_id: int) -> dict:
 
 
 async def list_cached_books() -> list[dict]:
-    """Return all cached books (without text field)."""
+    """Return publicly available cached books (Gutenberg only, without text field).
+
+    Uploaded books are excluded — their titles are private to the owner.
+    """
     async with aiosqlite.connect(DB_PATH) as db:
         db.row_factory = aiosqlite.Row
         async with db.execute(
-            "SELECT id, title, authors, languages, subjects, download_count, cover, cached_at FROM books ORDER BY cached_at DESC"
+            "SELECT id, title, authors, languages, subjects, download_count, cover, cached_at"
+            " FROM books"
+            " WHERE (source IS NULL OR source != 'upload')"
+            " ORDER BY cached_at DESC"
         ) as cursor:
             rows = await cursor.fetchall()
     result = []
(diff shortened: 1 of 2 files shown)


alfmunny:
Reviewed ✅

Correct and minimal. A few notes:

- `WHERE (source IS NULL OR source != 'upload')` correctly handles both legacy rows (source=NULL, pre-dating the column) and future rows — only explicit `'upload'` rows are excluded.
- The docstring update makes the intended scope clear for the next reader.
- Test covers the key invariant: Gutenberg book appears, private upload does not. The `insert_private_book` fixture ensures a realistic `source='upload'` row.

Privacy concern addressed: unauthenticated callers to `GET /books/cached` could previously enumerate private book titles. Good catch.

Good to merge once CI passes.

PR #481, opened 2026-04-23
fix: reject uploads that parse to zero chapters

### LINKED ISSUE #480: POST /books/upload accepts whitespace-only files, creating unconfirmable draft books
## Bug

`POST /books/upload` accepts files that parse to zero chapters (e.g. files containing only whitespace or newlines). The upload succeeds and creates a draft book in the DB, but:

1. The response returns `detected_chapters: []`
2. When the user attempts to confirm via `POST /books/{book_id}/chapters/confirm`, it returns 400 (empty chapters rejected — fixed in PR #477)
3. The draft book is stuck in the DB, consuming the user's upload quota

## Root cause

`parse_txt()` falls back to word-count splitting. For a whitespace-only file, `content.split()` produces `[]`, so `chunks = []` and `chapters = []`. The upload endpoint doesn't validate that chapters is non-empty before saving.

## Fix

After parsing, check that at least one chapter was detected:
```python
chapters = parsed["chapters"]
if not chapters:
    raise HTTPException(status_code=422, detail="No chapters could be detected. The file appears to be empty or unsupported.")
```

alfmunny:
## Summary
- `POST /books/upload` accepted whitespace-only (or otherwise empty) `.txt` files, saved a draft book with 0 chapters, and returned 200
- After PR #477 blocked confirming an empty chapters list, these drafts became permanently unconfirmable and consumed the user's upload quota forever
- Fix: after parsing, if `chapters` is empty, raise 422 immediately — no draft is written to the database

## Test plan
- [x] New regression test `test_upload_whitespace_only_txt_returns_422` — was FAIL before fix, PASS after
- [x] Full backend suite: 1072 passed, 0 failed

Closes #480

### DIFF
--- a/backend/routers/uploads.py
+++ b/backend/routers/uploads.py
@@ -95,6 +95,11 @@ async def upload_book(
         raise HTTPException(status_code=422, detail=f"Could not parse file: {exc}")
 
     chapters = parsed["chapters"]
+    if not chapters:
+        raise HTTPException(
+            status_code=422,
+            detail="No chapters could be detected. The file appears to be empty or contains no readable text.",
+        )
     book_id = await _save_upload_book(
         user_id=user["id"],
         title=parsed["title"],
--- a/backend/tests/test_router_uploads.py
+++ b/backend/tests/test_router_uploads.py
@@ -578,3 +578,21 @@ async def test_confirm_chapters_empty_list_returns_400(client, test_user):
     assert resp.status_code == 400, (
         f"Expected 400 for empty chapters list, got {resp.status_code}: {resp.text}"
     )
+
+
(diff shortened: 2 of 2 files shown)


alfmunny:
Reviewed ✅

Correct placement — the guard fires right after `parse` returns, before any DB write. A 0-chapter draft would have been:
1. Permanently orphaned (can't be confirmed)
2. Counted against any storage or quota logic

`422 Unprocessable Entity` is the right code — the file was received and parsed successfully, but the content is semantically invalid (no chapters). Good choice over 400.

Test covers the key path: whitespace-only .txt → 422 with 'chapter' in the error message. Clean.

Good to merge once CI passes.

PR #461, opened 2026-04-22
fix: validate rpm/rpd/max_output_tokens >= 1 in PUT /admin/queue/settings (closes #460)

alfmunny:
## Summary

- `PUT /admin/queue/settings` accepted zero and negative values for `rpm`, `rpd`, and `max_output_tokens` without any validation
- `rpm=0` / `rpd=0` cause division-by-zero in the queue worker's rate-limit sleep (`60 / rpm`)
- `max_output_tokens <= 0` is passed directly to the AI provider API, causing API errors or empty responses

## Fix

Added `ge=1` constraint via Pydantic `Field` to `rpm`, `rpd`, and `max_output_tokens` in `QueueSettingsRequest`. Invalid values now return `422 Unprocessable Entity` before reaching the handler.

## Test plan

- [x] `test_queue_settings_rejects_zero_rpm` — 422 for `rpm=0`
- [x] `test_queue_settings_rejects_negative_rpd` — 422 for `rpd=-1`
- [x] `test_queue_settings_rejects_zero_max_output_tokens` — 422 for `max_output_tokens=0`
- [x] `test_queue_settings_accepts_valid_positive_values` — 200 for valid values
- [x] Full backend suite: 1036 passed, 6 pre-existing stats failures (unrelated)

Closes #460

PR #483, opened 2026-04-23
fix: add bounds validation for TTS rate parameter

### LINKED ISSUE #482: POST /ai/tts accepts out-of-range rate values, sending malformed strings to Edge TTS
## Bug

`POST /ai/tts` accepts any float for the `rate` field with no bounds check.  The value is converted to a percentage string by `_rate_str(rate)` in `services/tts.py`:

```python
def _rate_str(rate: float) -> str:
    pct = round((rate - 1.0) * 100)
    return f"+{pct}%" if pct >= 0 else f"{pct}%"
```

At `rate=9999`, this produces `"+999800%"`.  At `rate=-100`, this produces `"-10100%"`.  Both are invalid Edge TTS rate strings and will cause the synthesis call to fail or behave unpredictably.  A zero or negative rate near zero (e.g. `rate=0.0` → `"-100%"`) would also halt synthesis.

## Reproduction

```bash
POST /ai/tts  {"text": "hello", "language": "en", "rate": 9999}
# Edge TTS receives rate="+999800%" — invalid
```

## Fix

Add `Field(ge=0.25, le=4.0)` to `TTSRequest.rate` in `routers/ai.py`.  This maps to -75% … +300% in Edge TTS percentage terms, which covers all realistic playback speeds.

alfmunny:
## Summary
- `TTSRequest.rate` had no min/max bounds — any float was accepted
- `rate=9999` would produce `"+999800%"` as the Edge TTS rate string, an invalid value that causes synthesis to fail unpredictably
- Fix: add `Field(ge=0.25, le=4.0)` — covers all realistic playback speeds (-75% … +300% in Edge TTS terms)

## Test plan
- [x] 4 new parametrized test cases for out-of-range values (9999, -100, 0.1, 5.0) → 422; all were FAIL before fix
- [x] 3 boundary/normal value cases (0.25, 4.0, 1.5) → 200; were already passing
- [x] Full backend suite: 1078 passed, 0 failed

Closes #482

### DIFF
--- a/backend/routers/ai.py
+++ b/backend/routers/ai.py
@@ -4,7 +4,7 @@ from typing import Literal
 
 from fastapi import APIRouter, Depends, HTTPException
 from fastapi.responses import Response
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 from services.auth import get_current_user, decrypt_api_key, check_book_access
 from services.db import (
     get_cached_translation,
@@ -90,7 +90,7 @@ class TranslateRequest(BaseModel):
 class TTSRequest(BaseModel):
     text: str
     language: str = "en"
-    rate: float = 1.0
+    rate: float = Field(default=1.0, ge=0.25, le=4.0)
     gender: Literal["female", "male"] = "female"
 
 
--- a/backend/tests/test_router_ai.py
+++ b/backend/tests/test_router_ai.py
@@ -524,6 +524,26 @@ async def test_tts_error_returns_500(client):
     assert resp.status_code == 500
 
 
+# ── TTS rate bounds validation (Issue #482) ───────────────────────────────────
+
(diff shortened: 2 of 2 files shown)


alfmunny:
✅ **PM review — approved**

Clean, minimal fix. The `Field(ge=0.25, le=4.0)` guard follows the same pattern as #461 (queue settings bounds). Bounds correctly map to -75%…+300% in Edge TTS percentage terms, covering all realistic playback speeds.

**Tests:** 7 parametrized cases — 4 out-of-range → 422 (regression tests), 3 boundary/normal → 200. Full suite at 1078 passed. Pattern is correct.

No follow-up issues needed — this closes #482 completely. CI still running (Backend + E2E pending); auto-merge is enabled.

PR #485, opened 2026-04-23
fix: add upper bound to GET /admin/queue/items limit parameter

### LINKED ISSUE #484: GET /admin/queue/items accepts unlimited limit parameter, enabling memory exhaustion
## Bug

`GET /admin/queue/items?limit=2147483647` executes a SQLite query that attempts to fetch 2 billion rows in a single response. The `limit` query parameter has no upper bound:

```python
# routers/admin.py:1090-1096
async def queue_items(
    status: str | None = None,
    book_id: int | None = None,
    limit: int = 200,          # ← no le= constraint
    _admin: dict = Depends(_require_admin),
):
    return await list_queue(status=status, book_id=book_id, limit=limit)
```

`list_queue()` passes the value directly to SQLite `LIMIT ?`.  Any admin (or compromised admin token) can exhaust server memory with a single request.

## Fix

Change `limit: int = 200` to `limit: int = Query(default=200, ge=1, le=1000)` in the endpoint signature.

alfmunny:
## Summary
- `GET /admin/queue/items?limit=2147483647` executed a SQLite query requesting 2+ billion rows, potentially exhausting server memory
- The `limit` parameter had no upper bound — only a default of 200
- Fix: `Query(default=200, ge=1, le=1000)` caps the maximum at 1000 items per request

## Test plan
- [x] New tests: `limit=9999999` → 422, `limit=0` → 422, `limit=200` (default) → 200, `limit=1000` (max) → 200
- [x] Full backend suite: 1080 passed, 0 failed

Closes #484

### DIFF
--- a/backend/routers/admin.py
+++ b/backend/routers/admin.py
@@ -5,7 +5,7 @@ Full CRUD where applicable.
 
 import json
 import aiosqlite
-from fastapi import APIRouter, Depends, HTTPException
+from fastapi import APIRouter, Depends, HTTPException, Query
 from pydantic import BaseModel, Field
 from services.auth import get_current_user, decrypt_api_key, encrypt_api_key
 from services.db import (
@@ -1090,7 +1090,7 @@ async def queue_set_settings(
 async def queue_items(
     status: str | None = None,
     book_id: int | None = None,
-    limit: int = 200,
+    limit: int = Query(default=200, ge=1, le=1000),
     _admin: dict = Depends(_require_admin),
 ):
     return await list_queue(status=status, book_id=book_id, limit=limit)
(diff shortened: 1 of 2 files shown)


alfmunny:
✅ **PM review — approved**

Minimal, correct fix. `Query(default=200, ge=1, le=1000)` is the right FastAPI primitive for query-parameter validation — consistent with how #461 used `Field(ge=1)` for body params and #483 used `Field(ge=0.25, le=4.0)`.

Cap of 1000 is appropriate: the admin queue panel never needs more than that per request, and it prevents a `limit=2147483647` call from issuing a billion-row SQLite scan.

**Tests:** 4 cases — two invalid (9999999, 0 → 422), two valid/boundary (200, 1000 → 200). Full suite at 1080 passed.

CI: Backend ✅ Frontend ✅ Docker ✅ Vercel ✅ — E2E still running. No follow-up issues needed. Closes #484 completely.

PR #487, opened 2026-04-23
fix: reject enqueue-book requests for unconfirmed draft uploads

### LINKED ISSUE #486: POST /admin/queue/enqueue-book silently enqueues 0 chapters for draft (unconfirmed) uploaded books
## Bug

When an admin calls `POST /admin/queue/enqueue-book` with a `book_id` that points to an uploaded book in draft state (chapters not yet confirmed by the user), the endpoint returns `{"ok": true, "enqueued": 0}` with no error.

**Root cause:**
1. `enqueue_for_book` calls `split_with_html_preference(book_id, text)`.
2. For a draft book, `book_chapters.py` returns `[]` (the comment even says "the caller should reject draft books explicitly").
3. The loop over chapters executes 0 times → `inserted = 0` → returns 0.
4. The admin endpoint blindly returns `{"ok": true, "enqueued": 0}`.

The admin has no way to tell whether 0 items were enqueued because all chapters are already translated, the book has no target languages configured, or because the book is a draft that cannot be translated yet.

## Fix

Add a draft check in `queue_enqueue_book` before calling `enqueue_for_book`.  If the book's text JSON has `"draft": true`, raise 400 with a clear message.

alfmunny:
## Summary
- `POST /admin/queue/enqueue-book` accepted draft (unconfirmed) uploaded books and returned `{"ok": true, "enqueued": 0}` with no error
- Admins had no way to distinguish this from "all chapters already translated" — a silent failure that required debugging to diagnose
- Fix: check for `"draft": true` in the book's text JSON before calling `enqueue_for_book`; raise 400 with a clear message if the book hasn't been confirmed yet

## Test plan
- [x] New regression test: enqueue a draft-state uploaded book → 400 with "confirm" in detail; was 200 before fix
- [x] Existing `test_enqueue_book_nonexistent_returns_404` still passes
- [x] Full backend suite: 1084 passed, 0 failed

Closes #486

### DIFF
--- a/backend/routers/admin.py
+++ b/backend/routers/admin.py
@@ -1108,8 +1108,23 @@ async def queue_enqueue_book(
     req: EnqueueBookRequest,
     admin: dict = Depends(_require_admin),
 ):
-    if not await get_cached_book(req.book_id):
+    book = await get_cached_book(req.book_id)
+    if not book:
         raise HTTPException(status_code=404, detail="Book not found")
+    text = book.get("text") or ""
+    if text.lstrip().startswith("{"):
+        try:
+            import json as _json
+            _data = _json.loads(text)
+            if _data.get("draft"):
+                raise HTTPException(
+                    status_code=400,
+                    detail="Book has not been confirmed yet. Confirm chapter splits before enqueueing for translation.",
+                )
+        except HTTPException:
+            raise
+        except (ValueError, TypeError):
+            pass
     added = await enqueue_for_book(
         req.book_id,
         target_languages=req.target_languages,
--- a/backend/tests/test_router_admin.py
+++ b/backend/tests/test_router_admin.py
@@ -1081,6 +1081,35 @@ async def test_enqueue_book_nonexistent_returns_404(admin_client):
     assert res.status_code == 404
 
 
+async def test_enqueue_draft_book_returns_400(admin_client, admin_db):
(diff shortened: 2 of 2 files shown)


alfmunny:
✅ **PM review — approved**

The silent `{"ok": true, "enqueued": 0}` response for draft books was a real usability gap — admins had no way to distinguish "nothing to translate" from "book not confirmed yet" without reading logs.

**Fix correctness:** The JSON-probe approach (`text.lstrip().startswith("{")` → parse → check `draft`) is correct given the current storage model. The `except (ValueError, TypeError): pass` guard correctly ignores non-draft JSON-like text without masking the `HTTPException`.

Minor note: `json` is already imported at the top of `admin.py` — the `import json as _json` inside the function body is redundant. Not a blocker.

**Tech debt note:** This is now the third location in the codebase that probes `book.text` for the `{"draft": ..., "chapters": [...]}` JSON structure (after `book_chapters.py:split_with_html_preference` and the upload confirm logic). Each one is correct individually, but the count is growing. This reinforces the case for #357 (migrate uploaded book chapters to a dedicated `user_book_chapters` table) — once that lands, all these probes go away in favour of a single schema-level field.

**Tests:** 1 regression test (draft book → 400, "confirm" in detail). 1084 suite passing. No follow-up issues needed beyond the existing #357 tracking.

PR #491, opened 2026-04-23
fix: move upload quota check inside BEGIN IMMEDIATE transaction

### LINKED ISSUE #489: POST /books/upload has TOCTOU race condition in quota check, allowing uploads beyond the 10-book limit
## Bug

The quota check and book insertion in `POST /books/upload` are not atomic.  Two concurrent upload requests by the same user can both read the count as 9, both pass the check, and both insert — leaving 11 books when the limit is 10.

```
Request A: count = 9 → passes quota check
Request B: count = 9 → passes quota check
Request A: inserts book → count = 10
Request B: inserts book → count = 11  ← exceeds limit
```

**Root cause:** `_user_upload_count` query and `_save_upload_book` are in separate transactions.

## Fix

Move the quota check inside the `BEGIN IMMEDIATE` transaction in `_save_upload_book` (or a wrapper).  `BEGIN IMMEDIATE` acquires a write lock immediately, preventing another writer from interleaving.

alfmunny:
## Summary
- The upload quota check (`_user_upload_count`) and book insertion (`_save_upload_book`) ran in separate transactions — a classic TOCTOU race
- Two concurrent uploads by the same user could both read count=9, both pass the check, and both insert, leaving 11 books when the limit is 10
- Fix: move the quota re-check inside the `BEGIN IMMEDIATE` transaction in `_save_upload_book`; the write lock prevents any other writer from interleaving between check and insert

## Test plan
- [x] Updated `test_upload_quota_exceeded_returns_429` to fill quota via real DB inserts instead of mocking `_user_upload_count` (which is no longer on the critical path)
- [x] Full affected tests: 78 passed, 0 failed

Closes #489

### DIFF
--- a/backend/routers/uploads.py
+++ b/backend/routers/uploads.py
@@ -26,12 +26,30 @@ async def _user_upload_count(user_id: int) -> int:
 
 
 async def _save_upload_book(user_id: int, title: str, author: str, filename: str,
-                            file_size: int, fmt: str, draft_chapters: list[dict]) -> int:
-    """Insert a new uploaded book row (source='upload') and book_uploads row."""
+                            file_size: int, fmt: str, draft_chapters: list[dict],
+                            max_books: int | None = None) -> int:
+    """Insert a new uploaded book row (source='upload') and book_uploads row.
+
+    If max_books is provided, the quota is re-checked inside the BEGIN IMMEDIATE
+    transaction to prevent TOCTOU races between concurrent uploads.
+    """
     draft_json = json.dumps({"draft": True, "chapters": draft_chapters})
     async with aiosqlite.connect(_db.DB_PATH) as db:
-        await db.execute("BEGIN")
+        # BEGIN IMMEDIATE acquires a write lock immediately so that a second
+        # concurrent upload cannot pass the quota check after this connection
+        # has already started its insert (#489).
+        await db.execute("BEGIN IMMEDIATE")
         try:
+            if max_books is not None:
+                async with db.execute(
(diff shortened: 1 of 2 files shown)


alfmunny:
✅ **PM review — approved**

Classic TOCTOU fix — the same `BEGIN IMMEDIATE` pattern used in PR #454 (confirm race). Moving the quota re-check inside the write lock is exactly right: SQLite's `BEGIN IMMEDIATE` acquires the write lock at transaction start, so no second uploader can interleave a COUNT→INSERT between this connection's COUNT and INSERT.

**Details look correct:**
- `except HTTPException: ROLLBACK; raise` before the generic `except Exception` ensures the 429 path rolls back cleanly and explicitly (previously the generic handler would have caught it too, but the intent is now unambiguous)
- Admin exemption preserved via `max_books=None` — the inner check is skipped entirely for admins
- `_user_upload_count` is still used by `GET /books/upload/quota` — that read-only path doesn't need the lock

**Test update is correct:** The old test mocked `_user_upload_count` which is no longer on the critical path. The new test fills quota via real DB inserts and verifies the in-transaction re-check fires — this is the right approach per the project's "no mock workarounds for real behavior" policy.

**Minor note:** PR description mentions "78 passed" (affected tests only) rather than the full suite. CI will verify, but worth running `pytest --tb=short -q` from the `backend/` directory to confirm the full 1084+ pass locally before the next PR.

CI: Frontend ✅ Docker ✅ Vercel ✅; Backend + E2E still running. No follow-up issues needed — closes #489 completely.

PR #503, opened 2026-04-23
fix: prevent duplicate book_insights rows with UNIQUE INDEX and INSERT OR IGNORE

### LINKED ISSUE #497: POST /insights allows duplicate Q&A rows — book_insights has no UNIQUE constraint
## Bug

`save_insight` in `services/db.py` (lines 956-960) does a plain `INSERT` with no `ON CONFLICT` handling, and the `book_insights` table has no UNIQUE constraint. Rapid duplicate POSTs create multiple identical rows (same `user_id`, `book_id`, `chapter_index`, `question`, `answer`).

## Contrast
Annotations have `UNIQUE(user_id, book_id, chapter_index, sentence_text)`.  
Vocabulary occurrences have `UNIQUE(vocabulary_id, book_id, chapter_index, sentence_text)`.  
Insights have no such guard.

## Impact
- Duplicate rows returned in `GET /insights`
- Unbounded table growth from automated/retried clients

## Fix
Add a UNIQUE index on `(user_id, book_id, chapter_index, question)` and switch `save_insight` to `INSERT OR IGNORE` (returning the existing row) or a migration-level constraint.

alfmunny:
## What changed
- New migration `022_book_insights_unique.sql`: adds `UNIQUE INDEX uq_book_insights_question` on `(user_id, book_id, COALESCE(chapter_index, -1), question)`
- `save_insight` in `services/db.py`: switched from plain `INSERT` to `INSERT OR IGNORE`, returns existing row on duplicate (SELECT before COMMIT to preserve delete-race safety from #351)
- `services/migrations.py`: adds bootstrap entry for migration 022 — marks it as applied for legacy DBs that pre-date migration 015 (use old `insight` column instead of `question`/`answer`) to prevent startup failure

## Why
`book_insights` had no UNIQUE constraint. Rapid or retried duplicate POSTs created multiple identical rows, inflating `GET /insights` results and wasting storage — unlike annotations and vocabulary_occurrences which already deduplicate.

## Test plan
- `test_duplicate_insight_returns_existing_row` — POSTing identical insight twice returns same `id`, `GET /insights` returns exactly 1 row
- `test_save_insight_select_runs_before_commit` — SELECT order invariant preserved
- `test_schema_migrations_table_records_versions` — migration 022 appears in schema_migrations on fresh DB
- `test_bootstrap_marks_016_when_context_text_exists` — legacy schema (no `question` col) still bootstraps cleanly
- Full backend suite: 1096 passed

Closes #497

### DIFF
--- a/backend/migrations/022_book_insights_unique.sql
+++ b/backend/migrations/022_book_insights_unique.sql
@@ -1,0 +1,3 @@
+-- Prevent duplicate insights for the same question in the same chapter
+CREATE UNIQUE INDEX IF NOT EXISTS uq_book_insights_question
+    ON book_insights (user_id, book_id, COALESCE(chapter_index, -1), question);
--- a/backend/services/db.py
+++ b/backend/services/db.py
@@ -953,14 +953,25 @@ async def save_insight(
 ) -> dict:
     async with aiosqlite.connect(DB_PATH) as db:
         db.row_factory = aiosqlite.Row
-        cursor = await db.execute(
-            """INSERT INTO book_insights (user_id, book_id, chapter_index, question, answer, context_text)
+        await db.execute(
+            """INSERT OR IGNORE INTO book_insights (user_id, book_id, chapter_index, question, answer, context_text)
                VALUES (?, ?, ?, ?, ?, ?)""",
             (user_id, book_id, chapter_index, question, answer, context_text),
         )
-        row_id = cursor.lastrowid
-        async with db.execute("SELECT * FROM book_insights WHERE id = ?", (row_id,)) as c:
-            row = await c.fetchone()
+        # SELECT before COMMIT — prevents dict(None) crash if a concurrent
+        # delete_book removes the row between our COMMIT and a later SELECT.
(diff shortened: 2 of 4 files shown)


alfmunny:
Good fix — all the subtle cases are handled correctly.

The `COALESCE(chapter_index, -1)` in the UNIQUE index is the right call: SQLite treats each NULL as distinct, so without it two book-level insights (chapter_index IS NULL) with the same question would bypass the constraint. The split SELECT (`IS NULL` vs `= ?`) mirrors this correctly on the read side.

The migration bootstrap logic looks correct: the UNION guard handles both the "already applied" case (index exists) and the legacy pre-015 schema case (table has no `question` column) without crashing startup.

One minor gap not worth blocking on: the dedup test only covers the `chapter_index != NULL` path. The `chapter_index IS NULL` (book-level insight) dedup path is exercised only at the DB layer via the COALESCE index — an explicit test case would round this out, but the COALESCE logic itself is correct.

Approved — good to merge.

PR #523, opened 2026-04-23
fix: cap SaveTranslationRequest.paragraphs list size and per-item length

### LINKED ISSUE #520: SaveTranslationRequest.paragraphs unbounded — list size and per-item length uncapped
## Description
Follow-up from #516 / PR #518. The scalar fields were bounded but `paragraphs: list[str]` remains unbounded in `SaveTranslationRequest` (`PUT /ai/translate/cache`).

## Risk
An authenticated user can send `{"paragraphs": ["x" * 50000] * 10000}` — 500 MB of JSON in a single request, stored as a JSON blob in `translations.paragraphs`.

## Fix
Use a Pydantic `Annotated` type or custom validator:
```python
from typing import Annotated
from pydantic import Field

paragraphs: Annotated[list[str], Field(max_length=1000)] = []
# plus per-item max_length via a validator:
# each paragraph <= 50_000 chars
```
Or use a constrained list type:
```python
paragraphs: list[Annotated[str, Field(max_length=50_000)]] = Field(..., max_length=1000)
```

## Test
- PUT with 1001 paragraphs → 422
- PUT with one paragraph of length 50001 → 422

alfmunny:
## Summary

- Add `max_length=2000` to `SaveTranslationRequest.paragraphs` (list item count cap)
- Add `max_length=50000` per-item constraint via `Annotated[str, Field(max_length=50000)]`

## Why

`paragraphs: list[str]` was completely unbounded, allowing an attacker to POST a list with millions of items or a single multi-megabyte string, both of which get stored in the `translations` table. Closes #520.

## Test plan

- [ ] `test_translate_cache_put_too_many_paragraphs_returns_422` — list with 2001 items → 422
- [ ] `test_translate_cache_put_oversized_paragraph_item_returns_422` — item with 50001 chars → 422
- [ ] Full backend suite (545 tests) passes

### DIFF
--- a/backend/routers/ai.py
+++ b/backend/routers/ai.py
@@ -1,6 +1,6 @@
 import asyncio
 from collections import defaultdict
-from typing import Literal
+from typing import Annotated, Literal
 
 from fastapi import APIRouter, Depends, HTTPException
 from fastapi.responses import Response
@@ -270,7 +270,7 @@ class SaveTranslationRequest(BaseModel):
     book_id: int
     chapter_index: int
     target_language: str = Field(..., max_length=20)
-    paragraphs: list[str]
+    paragraphs: list[Annotated[str, Field(max_length=50000)]] = Field(..., max_length=2000)
     provider: str | None = Field(default=None, max_length=100)
     model: str | None = Field(default=None, max_length=200)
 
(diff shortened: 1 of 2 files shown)


alfmunny:
PM review ✅

The fix is correct: `list[Annotated[str, Field(max_length=50000)]] = Field(..., max_length=2000)` bounds both the list length and per-item size. Two regression tests cover both limits. The `2000` cap (vs the 1000 in the issue) is fine — generous but sane.

Branch is BEHIND after #512 merged. Please rebase so CI can run on the current main.

PR #527, opened 2026-04-23
fix: deduplicate book_insights rows before creating UNIQUE INDEX in migration 022

### LINKED ISSUE #526: Migration 022 crashes with IntegrityError on databases with duplicate book_insights rows
## Description

Migration `022_book_insights_unique` fails with `sqlite3.IntegrityError: UNIQUE constraint failed: index 'uq_book_insights_question'` on databases that have pre-existing duplicate rows in `book_insights`.

This causes the app to **fail to start** on Railway (or any deployment that has duplicate insights and hasn't yet applied migration 022).

## Root cause

`022_book_insights_unique.sql` only contains:
```sql
CREATE UNIQUE INDEX IF NOT EXISTS uq_book_insights_question
    ON book_insights (user_id, book_id, COALESCE(chapter_index, -1), question);
```

SQLite raises `IntegrityError` when creating a unique index on a table that already contains duplicates.

## Fix

Add a `DELETE` step before the index creation to remove duplicate rows (keeping the lowest `rowid` per unique key):

```sql
DELETE FROM book_insights
WHERE rowid NOT IN (
    SELECT MIN(rowid)
    FROM book_insights
    GROUP BY user_id, book_id, COALESCE(chapter_index, -1), question
);
CREATE UNIQUE INDEX IF NOT EXISTS uq_book_insights_question
    ON book_insights (user_id, book_id, COALESCE(chapter_index, -1), question);
```

## Impact

App is currently down on Railway. Fix is safe for all DB states:
- Fresh DB: DELETE is a no-op on empty table
- DB with duplicates: removes them, then creates index
- DB where 022 already ran: `schema_migrations` prevents re-run

alfmunny:
## Summary

Migration `022_book_insights_unique` crashed on the Railway deployment with:
```
sqlite3.IntegrityError: UNIQUE constraint failed: index 'uq_book_insights_question'
```
This caused the app to fail to start.

**Root cause:** The migration only contained `CREATE UNIQUE INDEX` but never deduped existing rows first. SQLite raises `IntegrityError` immediately if the table has duplicates when you try to create a unique index on it.

**Fix:** Add a `DELETE` step that removes duplicate rows (keeping the lowest `rowid` per unique key) before creating the index. Safe for all DB states:
- Fresh DB → DELETE is a no-op on empty table
- DB with duplicates → removes them, index creation succeeds  
- DB where 022 already ran → `schema_migrations` prevents re-run

## Test plan

- [x] `test_022_deduplicates_before_creating_unique_index` — seeds a DB with duplicate `book_insights` rows, runs migration 022, verifies no exception, verifies duplicate removed, verifies index created
- [x] All 1135 existing tests pass

Closes #526

### DIFF
--- a/backend/migrations/022_book_insights_unique.sql
+++ b/backend/migrations/022_book_insights_unique.sql
@@ -1,3 +1,13 @@
+-- Remove duplicate insights rows (keep lowest rowid per unique key) before
+-- creating the index so this migration cannot fail with IntegrityError on
+-- databases that already have duplicate questions in the same chapter.
+DELETE FROM book_insights
+WHERE rowid NOT IN (
+    SELECT MIN(rowid)
+    FROM book_insights
+    GROUP BY user_id, book_id, COALESCE(chapter_index, -1), question
+);
+
 -- Prevent duplicate insights for the same question in the same chapter
 CREATE UNIQUE INDEX IF NOT EXISTS uq_book_insights_question
     ON book_insights (user_id, book_id, COALESCE(chapter_index, -1), question);
--- a/backend/tests/test_migrations.py
+++ b/backend/tests/test_migrations.py
@@ -693,6 +693,76 @@ async def test_legacy_db_with_018_chapter_summaries_gets_020_bootstrapped(tmp_db
             )
 
 
+# ── 022_book_insights_unique: dedup before index (issue #526) ────────────────
+
+async def test_022_deduplicates_before_creating_unique_index(tmp_db, tmp_migrations, monkeypatch):
+    """Regression #526: migration 022 must DELETE duplicate book_insights rows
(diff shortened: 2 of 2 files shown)


alfmunny:
**Pattern note for future migrations:**

Any migration that adds a UNIQUE constraint (or any constraint) to a table that already has data must include a data-cleanup step first. Rule of thumb:

> **Constraint migrations must handle dirty data before constraining.**

Examples:
- Adding `UNIQUE INDEX` → `DELETE` duplicates first (this PR)
- Adding `NOT NULL` column → `UPDATE SET col = default` first  
- Adding `CHECK` constraint → `DELETE` or `UPDATE` violating rows first

The `CREATE UNIQUE INDEX IF NOT EXISTS` only skips re-creation if the index already exists — it does NOT skip the constraint check on table data.

PR #532, opened 2026-04-23
fix: cap ImportTranslationEntry.paragraphs list size and per-item length

### LINKED ISSUE #524: ImportTranslationEntry.paragraphs unbounded — same 500MB risk as #520, admin endpoint
## Description

Follow-up from PR #512 / #521. The admin bounds fix in `781f578` correctly bounded all scalar fields in `ImportTranslationEntry` but left `paragraphs: list[str]` unbounded.

```python
# backend/routers/admin.py — current state after #512
class ImportTranslationEntry(BaseModel):
    ...
    paragraphs: list[str]   # ← still unbounded
    ...
```

## Risk

POST /admin/translations/import with thousands of paragraphs (each up to 50KB) can insert a ~500MB JSON blob into the `translations` table. Mitigated by admin-only access but should still be capped consistently.

## Fix

Same pattern as PR #523 (which fixes the identical field in `SaveTranslationRequest`):

```python
paragraphs: list[Annotated[str, Field(max_length=50_000)]] = Field(..., max_length=2000)
```

## Test

- POST /admin/translations/import with an entry containing > 2000 paragraphs → 422
- POST with a paragraph item > 50000 chars → 422

alfmunny:
## Summary

Add `max_length=2000` list-size cap and `max_length=50000` per-item cap to `ImportTranslationEntry.paragraphs` in the admin import endpoint.

## Why

`paragraphs: list[str]` in `ImportTranslationEntry` was fully unbounded, identical to the risk fixed in PR #523 for `SaveTranslationRequest`. An admin POST to `/admin/translations/import` with thousands of large paragraphs could write a ~500MB JSON blob into the `translations` table. Closes #524.

## Test plan

- [ ] `test_import_translations_too_many_paragraphs_returns_422` — 2001 items → 422
- [ ] `test_import_translations_oversized_paragraph_item_returns_422` — item with 50001 chars → 422
- [ ] Full backend suite (553 tests) passes

### DIFF
--- a/backend/routers/admin.py
+++ b/backend/routers/admin.py
@@ -581,7 +581,7 @@ class ImportTranslationEntry(BaseModel):
     book_id: int
     chapter_index: int
     target_language: str = Field(..., max_length=20)
-    paragraphs: list[str]
+    paragraphs: list[Annotated[str, Field(max_length=50000)]] = Field(..., max_length=2000)
     provider: str | None = Field(default=None, max_length=100)
     model: str | None = Field(default=None, max_length=200)
     title_translation: str | None = Field(default=None, max_length=500)
--- a/backend/tests/test_router_admin.py
+++ b/backend/tests/test_router_admin.py
@@ -2112,3 +2112,34 @@ async def test_enqueue_book_oversized_target_language_returns_422(admin_client, 
     assert res.status_code == 422, (
         f"Expected 422 for oversized target_language in enqueue-book, got {res.status_code}"
     )
+
+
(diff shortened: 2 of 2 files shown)


alfmunny:
**PM review ✅ — approved**

Exactly the right fix for #524. `list[Annotated[str, Field(max_length=50000)]] = Field(..., max_length=2000)` bounds both the list size and per-item length, matching the pattern used for `SaveTranslationRequest.paragraphs` in PR #523. Tests cover the two key edges (list too long → 422, item too long → 422). All CI green. Good to merge.

PR #547, opened 2026-04-23
feat: store Gutenberg EPUBs in DB for reliable chapter splitting

alfmunny:
## Summary

- Adds `book_epubs` table (migration 023) to store Gutenberg EPUB binaries per book
- New `build_chapters_from_epub()` in `splitter.py` uses EPUB spine + NCX/nav TOC for reliable chapter titles and order
- `split_with_html_preference()` in `book_chapters.py` now follows DB-only priority chain: pre-split JSON → stored EPUB → plain-text regex (no on-demand HTML fetch at chapter-load time)
- New Gutenberg books have their EPUB fetched and stored at add-time via background task; existing books trigger a silent background fetch on first chapter access
- `scripts/backfill_epubs.py` for bulk EPUB population of existing books
- Fixed migration 023 comment semicolon that caused the migration runner to produce invalid SQL
- Updated all tests: replaced mocks of removed `get_book_html`/`build_chapters_from_html` with `services.db.get_book_epub_bytes` and `services.book_chapters.build_chapters_from_epub`

## Why

Plain-text regex splitting and on-demand HTML fetching are unreliable: regex misses nested structure (plays, dramas like Faust), and HTML fetches add latency and Gutenberg network dependency at reader page load. EPUB binaries encode the authoritative reading order and chapter titles via spine and NCX/nav TOC, giving consistent results across all book types.

## Test plan

- [x] All 1152 backend tests pass (`pytest --tb=short -q --no-cov`)
- [x] 6 new EPUB splitter tests: basic parsing, TOC title extraction, paragraph breaks, spine order, invalid bytes fallback, short-item skipping
- [x] `test_book_chapters.py` updated to mock EPUB path and verify all 3 priority paths (EPUB, fallback, plain-text)
- [x] Migration 023 applies cleanly to fresh DB
- [x] `scripts/backfill_epubs.py --dry-run` script available for bulk population

### DIFF
--- a/backend/migrations/023_book_epubs.sql
+++ b/backend/migrations/023_book_epubs.sql
@@ -1,7 +1,7 @@
 -- Stores the downloaded EPUB binary for each Gutenberg book.
 -- Used by split_with_html_preference as the primary chapter source
 -- (preferred over on-demand HTML fetch and plain-text regex splitting).
--- Populated at book-add time for new books; backfill script available
+-- Populated at book-add time for new books. Backfill script available
 -- for books already in the DB.
 CREATE TABLE IF NOT EXISTS book_epubs (
     book_id    INTEGER PRIMARY KEY REFERENCES books(id) ON DELETE CASCADE,
--- a/backend/migrations/023_book_epubs.sql
+++ b/backend/migrations/023_book_epubs.sql
@@ -1,0 +1,11 @@
+-- Stores the downloaded EPUB binary for each Gutenberg book.
+-- Used by split_with_html_preference as the primary chapter source
+-- (preferred over on-demand HTML fetch and plain-text regex splitting).
+-- Populated at book-add time for new books; backfill script available
+-- for books already in the DB.
+CREATE TABLE IF NOT EXISTS book_epubs (
+    book_id    INTEGER PRIMARY KEY REFERENCES books(id) ON DELETE CASCADE,
+    epub_url   TEXT    NOT NULL DEFAULT '',
+    epub_bytes BLOB    NOT NULL,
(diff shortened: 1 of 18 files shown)


alfmunny:
**PM review — approved ✅**

This is a solid architectural improvement. Eliminating on-demand HTML fetches from the reader path removes a class of Gutenberg network failures that were invisible to users until a page load hung.

**Strengths:**
- DB-only priority chain at chapter-load time (JSON → EPUB → plain-text) is clean and testable
- 6 new EPUB splitter unit tests + all existing tests updated; 1152 total passing
- Design doc in `docs/design-epub-ingestion.md` ✅ (matches the Architect workflow pattern)
- Migration 023 is a new table with no pre-existing data — no cleanup step needed per migration policy ✅
- `_background_fetch_epub` is fire-and-forget with proper exception suppression at DEBUG level — correct pattern for non-critical background work
- `parse_epub` in `book_parser.py` now delegates to `build_chapters_from_epub`, eliminating duplicated HTML-stripping logic ✅

**Minor notes (non-blocking):**
1. `epub_url TEXT NOT NULL DEFAULT ''` — an empty string as the no-EPUB sentinel is slightly surprising compared to `NULL`. Not blocking, but a future reader of the DB might not expect `''` to mean "no EPUB". Could be addressed in a follow-up migration if it ever causes confusion.
2. `_epub_fetch_attempted: set[int]` is unbounded for the process lifetime. For a server that sees many thousands of books this grows forever. In practice it holds one int per book ever accessed, so the memory is negligible — just noting it for the record.
3. E2E check is still pending in CI — not blocking approval but should be green before merge.

No design-rule violations (no UI changes). No new endpoints without bounds checks (no new user-input endpoints). No follow-up issues needed from my side — the scope is complete.

Ready to merge once E2E clears.

PR #543, opened 2026-04-23
fix: stabilize flaky E2E highlight test

alfmunny:
## Summary
The `highlight sends full sentence text` E2E test was flaky because `selectionchange` was dispatched before React's `useEffect` had registered the `SelectionToolbar` listener (race between Playwright's `waitForSelector` returning and React's effect queue flushing).

Two fixes:
1. Added `waitForSelector("#reader-scroll")` to ensure the reader is fully rendered before interacting
2. Double-dispatch `selectionchange` — the first fires immediately, the second fires after 2 animation frames so React's `useEffect` has registered its listener

Also increased toolbar/color panel `toBeVisible` timeouts from 3s to 6s for CI headroom.

## Test plan
- [x] `--repeat-each=5`: 5/5 passes locally

Closes #514

PR #533, opened 2026-04-23
fix: add max_length to vocabulary word/lang path and query params

alfmunny:
## Summary

Add `max_length` bounds to vocabulary path/query parameters:

- `word: str = Path(..., max_length=200)` on `GET /vocabulary/definition/{word}` and `DELETE /vocabulary/{word}`
- `lang: str = Query(default="en", max_length=20)` on `GET /vocabulary/definition/{word}?lang=`

## Why

Unbounded path/query params allowed arbitrarily long strings to reach the wiktionary external API and the DB DELETE query. Closes #529.

## Test plan

- [ ] `test_definition_oversized_word_returns_422` — word with 201 chars → 422
- [ ] `test_definition_oversized_lang_returns_422` — lang with 21 chars → 422
- [ ] `test_remove_word_oversized_word_returns_422` — word with 201 chars → 422
- [ ] Full backend suite (556 tests) passes

PR #549, opened 2026-04-23
feat: AI dictionary fallback for German compounds and multi-word phrases (closes #444)

### LINKED ISSUE #444: Feature improvement:  better dictionary provider rather than wiktionary
# Problem 

Wiktionary will always fail to grab definition when we select on a combined word which is very common in German. 

Also short phrase will also not work. 

Do we have more intelligent provider or AI dictionary which more reliable and robust against all kinds of selection of words and also free of charge. 

Please investigate and implement a good alternative

alfmunny:
## What

Adds an AI-powered fallback to `GET /vocabulary/definition/{word}` for cases where wiktionary returns no definitions:

- German compound words (e.g. `Aufmerksamkeitsdefizit`) — wiktionary doesn't list inflected compound forms
- Multi-word phrases selected in the reader

When wiktionary returns empty, the endpoint calls Gemini to generate a definition (if the user has a key configured). Returns the same `{lemma, language, definitions, url}` shape plus `"source": "ai"`. Silently skips fallback for users without a key (no behavior change for them).

## Test plan

- [x] `test_definition_falls_back_to_ai_when_wiktionary_empty` — mocked empty wiktionary + mocked AI → returns AI definition
- [x] `test_definition_returns_wiktionary_when_available` — wiktionary has results → AI is never called
- [x] `test_definition_returns_empty_when_no_key_and_wiktionary_fails` — no key → empty definitions returned unchanged
- [x] Full backend suite: 916 passed

Closes #444

### DIFF
--- a/backend/routers/admin.py
+++ b/backend/routers/admin.py
@@ -6,7 +6,7 @@ Full CRUD where applicable.
 import json
 import aiosqlite
 from typing import Annotated
-from fastapi import APIRouter, Depends, HTTPException, Query
+from fastapi import APIRouter, Body, Depends, HTTPException, Path, Query
 from pydantic import BaseModel, Field
 from services.auth import get_current_user, decrypt_api_key, encrypt_api_key
 from services.db import (
@@ -422,7 +422,7 @@ async def delete_book_translations(book_id: int, _admin: dict = Depends(_require
 @router.delete("/translations/{book_id}/{target_language}")
 async def delete_language_translations(
     book_id: int,
-    target_language: str,
+    target_language: str = Path(..., max_length=20),
     _admin: dict = Depends(_require_admin),
 ):
     """Delete all cached translations for one language of a book."""
@@ -461,7 +461,7 @@ async def delete_language_translations(
 async def delete_translation(
     book_id: int,
     chapter_index: int,
-    target_language: str,
+    target_language: str = Path(..., max_length=20),
     _admin: dict = Depends(_require_admin),
 ):
     """Delete a specific cached translation."""
@@ -499,7 +499,7 @@ async def delete_translation(
 async def retranslate(
     book_id: int,
     chapter_index: int,
-    target_language: str,
+    target_language: str = Path(..., max_length=20),
     admin: dict = Depends(_require_admin),
 ):
     """Delete cached translation and re-translate the chapter."""
@@ -727,8 +727,8 @@ class MoveTranslationRequest(BaseModel):
 async def move_translation(
     book_id: int,
     chapter_index: int,
-    target_language: str,
-    req: MoveTranslationRequest,
+    target_language: str = Path(..., max_length=20),
+    req: MoveTranslationRequest = Body(...),
     _admin: dict = Depends(_require_admin),
 ):
     """Reassign a cached translation to a different chapter_index.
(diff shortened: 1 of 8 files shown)


alfmunny:
**PM review — approved ✅**

Clean, well-scoped feature. 3 regression tests cover all key paths (fallback used, fallback skipped, no-key → no-fallback). The silent-except pattern for the AI fallback is correct — a Gemini failure should never surface as an error to the user.

**One minor concern (non-blocking):**
The `"url"` field in the AI-generated response still points to `https://en.wiktionary.org/wiki/{word}`. For the cases this fallback is meant to handle (German compounds, multi-word phrases), that URL will resolve to a blank or missing wiktionary page — it could confuse a user who clicks through. Consider returning `null` or `""` for the URL when `source == "ai"`, or pointing to a Duden/wiktionaryplus URL for German words. Filing a minor follow-up issue for tracking.

**CI:** Most checks still pending — `Enable auto-merge` already succeeded so it'll self-merge once the suite clears.

No design-rule violations. No new user-input fields beyond what's already bounded by PR #533 (`word` path param max_length=200, `lang` query param max_length=20). Good to go.

alfmunny:
**PM review — two items to address before merge:**

1. **Branch is BEHIND main** — rebase onto origin/main is needed. The E2E failure is likely just the stale base (E2E fix from #543 landed after this branch was last pushed).

2. **Bug: `url` field wrong when `source=ai`** — tracked as #551. In `services/gemini.py:define_word`, both the happy-path and exception-path return `"url": f"https://en.wiktionary.org/wiki/{word}"` even though wiktionary had **no entry** for this word. A user clicking that link lands on a missing page. Fix: set `"url": ""` when source is AI. The frontend already handles the empty-URL case.

Please rebase and patch in the same PR — it's a one-liner fix.

alfmunny:
**PM re-check after rebase:**

Branch has been rebased onto origin/main — the BEHIND issue is resolved and E2E CI is now running.

However, issue #551 (**url field wrong when source=ai**) is still not fixed in this PR. The current code returns:

```python
"url": f"https://en.wiktionary.org/wiki/{word}",
"source": "ai",
```

When `source == "ai"`, the `url` should be `""` (empty string) — wiktionary had no entry for this word, so linking to it is actively misleading. Please fix this in `services/gemini.py` (both the success and except branches of `define_word`) before merge.

One-line fix:
```python
"url": "",
"source": "ai",
```

Will re-review once pushed.

alfmunny:
**PM — rebase needed:** This branch is 6 commits behind main. All CI checks pass but GitHub won't auto-merge until the branch is up to date. Please rebase onto origin/main and force-push.

alfmunny:
Branch is now **7 commits behind main** — the rebase request from last cycle was not acted on. Auto-merge cannot proceed while the branch is behind. Please rebase and force-push urgently:

```bash
git fetch origin main
git rebase origin/main
git push origin <branch> --force-with-lease
```

If this PR is no longer being actively worked on, please close it so the issue can be reclaimed.

alfmunny:
PM escalation (cycle 155): This PR is BEHIND main. Please rebase and force-push to unblock CI. If the owning session is no longer active, please re-assign or close this PR.

PR #555, opened 2026-04-23
design: user_book_chapters table to replace JSON-in-books.text (#357)

### LINKED ISSUE #357: tech-debt: uploaded book chapters stored as JSON in books.text column

alfmunny:
## Summary

Design doc for migrating uploaded book chapters out of the `books.text` JSON blob and into a proper `user_book_chapters` table (Path B — schema change, 3+ files, requires PM approval).

**Problem:** Uploaded books store chapters as `{"draft":true,"chapters":[...]}` in `books.text`. This breaks full-text search, requires silent branching in 5 code paths, and violates the column's plain-text contract.

**Proposed solution:**
- New `user_book_chapters` table (migration 024)
- One-time Python data migration script (idempotent, safe for prod)
- Replace all `json.loads(books.text)` branches in `uploads.py`, `books.py`, `admin.py`
- No API shape changes — all endpoints keep the same request/response

**Files in scope:** 8 files, backend only, no frontend changes.

**Estimated effort:** ~4 hours once approved.

## Review checklist

- [ ] Schema design looks correct (table, index, FK)
- [ ] Migration script is safe for production data
- [ ] Open questions answered (empty string vs NULL, Gutenberg unification, FTS defer)
- [ ] Effort estimate accepted

Closes #357

### DIFF
--- a/docs/FEATURES.md
+++ b/docs/FEATURES.md
@@ -368,3 +368,74 @@ python scripts/pretranslate.py --book-id 1342 --lang de --force   # overwrite ca
 **Dependencies:** `transformers`, `sentencepiece`, `torch` (CPU), `requests` (for Ollama).  
 **Chunking:** split at sentence boundaries to stay under MarianMT's 512-token limit.  
 **Estimated effort:** ~6 hours.
+
+---
+
+## Session 5 — 2026-04-23
+
+### Shipped this session
+
+- [x] Feature 9: EPUB DB Storage for Gutenberg books (PR #547)
+- [x] Feature 10: Admin Uploads Tab (PRs #546, #553)
+
+### Filed / awaiting approval
+
+- [ ] Feature 11: user_book_chapters table — replace JSON-in-books.text (design PR #555, issue #357)
+- [ ] Feature 2: Vocabulary Flashcards / SRS (design in FEATURES.md Feature 2, issue #556)
+- [ ] Feature 12: Chapter Comprehension Quiz (issue #557)
+
+---
+
+## Feature 9: EPUB DB Storage for Gutenberg Books ✅ (merged — PR #547)
+
+### Overview
+Gutenberg EPUBs are now downloaded once and stored in the database (`book_epubs` table), giving reliable chapter splitting with correct spine order and NCX/nav titles. Eliminates on-demand HTML re-fetching at chapter-load time.
+
+### What shipped
+- Migration 023: `book_epubs` table (`book_id PK`, `epub_bytes BLOB`, `epub_url`, `cached_at`)
+- `gutenberg.py`: `get_book_epub()` — fetches no-images EPUB via Gutendex formats list
+- `db.py`: `save_book_epub()` + `get_book_epub_bytes()`
+- `splitter.py`: `build_chapters_from_epub()` using spine order + NCX/nav titles
+- `book_chapters.py`: DB-only split path — EPUB → plain-text regex fallback; background lazy fetch for existing books
+- `routers/books.py`: fires EPUB download in background at book-add time
+- `book_parser.py`: `parse_epub()` now uses same `build_chapters_from_epub()` as Gutenberg path
+- `scripts/backfill_epubs.py`: one-time backfill for books already in DB
+
+---
+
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM design review — approved ✅**

Solid design doc. The three problems are correctly identified, the schema is clean, and keeping Gutenberg chapter caching separate is the right call (no scope creep).

**One item to clarify before implementation starts:**

The design says the data migration runs as a Python script called from the migration runner "or as a one-time admin script." The current migration runner in this repo processes `.sql` files; it does not have a hook for Python scripts. Before implementation, please resolve this ambiguity in the design doc — either:
- (a) Add a hook to the migration runner that can invoke `backend/scripts/migrate_upload_chapters.py` before or after running the SQL file, or
- (b) Document that ops must run the Python script manually before deploying migration 024.

Option (b) is simpler and lower-risk; option (a) is cleaner but requires a migration runner change. Either is acceptable — just make the call explicit so the implementation PR doesn't leave it ambiguous.

**Other notes:**
- `INSERT OR IGNORE` idempotency + `text LIKE '{%'` filter for skipping cleared rows is the right approach.
- The `test_migrations.py` test requirement is correctly noted — the implementation PR must include it.
- BEHIND main — rebase needed before this merges.

Implementation can proceed once the migration runner question is resolved in the doc.

alfmunny:
Branch is now **9 commits behind main**. This design doc PR has been open for several cycles without a rebase. Please rebase and force-push so CI can run against the current codebase:

```bash
git fetch origin main
git rebase origin/main
git push origin design/user-book-chapters-357 --force-with-lease
```

alfmunny:
PM escalation (cycle 155): This PR is BEHIND main. Please rebase and force-push to unblock CI. If the owning session is no longer active, please re-assign or close this PR.

alfmunny:
Revision applied (commit \`0ef9288\`).

**Resolved:** picked option (b). The SQL migration runner stays as-is; ops runs the Python script manually.

**Script changed** to a two-phase design so the rollback path stays safe during the router deploy:

1. `copy_phase` (default, idempotent) — populates \`user_book_chapters\` from the JSON in \`books.text\`. Does **not** clear \`books.text\`.
2. `finalize_phase` (triggered by \`--finalize\`, idempotent) — clears \`books.text\` only for rows that already have entries in \`user_book_chapters\`.

**Deployment checklist (in the doc):**
1. Deploy backend with migration 025 + new services, **old router** still reading \`books.text\`.
2. Ops: \`python -m backend.scripts.migrate_upload_chapters\` (copy phase).
3. Deploy new router that reads from \`user_book_chapters\`.
4. Ops: \`python -m backend.scripts.migrate_upload_chapters --finalize\` once the new router is stable.

If step 3 rolls back, \`books.text\` is still intact and the old router keeps working. No migration-runner changes needed.

Test plan expanded to cover both phases' idempotency and non-interference with legacy (non-upload) rows.

PR #609, opened 2026-04-23
fix: cap ConfirmChaptersBody.chapters list at 1000 entries (closes #606)

### LINKED ISSUE #606: bug: ConfirmChaptersBody.chapters list has no max_length — DoS via oversized chapter list
## Description

`ConfirmChaptersBody.chapters` in `routers/uploads.py` accepts an unbounded list:

```python
class ConfirmChaptersBody(BaseModel):
    chapters: list[ConfirmChapterSpec]  # no max_length on the list
```

A user can POST `/books/{id}/chapters/confirm` with thousands of chapter specs. Even though each spec's `title` is bounded at 500 chars, a list of 10,000 specs creates a large JSON blob written to `books.text` via:

```python
confirmed_json = json.dumps({"draft": False, "chapters": final_chapters})
await db.execute("UPDATE books SET text=? WHERE id=?", (confirmed_json, book_id))
```

A book is unlikely to have more than a few hundred chapters; 1000 is a generous upper bound.

## Fix

```python
class ConfirmChaptersBody(BaseModel):
    chapters: list[ConfirmChapterSpec] = Field(..., max_length=1000)
```

## Test

POST /books/{id}/chapters/confirm with 1001 chapter specs → 422

alfmunny:
## Summary

Add `max_length=1000` to `ConfirmChaptersBody.chapters` to prevent a user from posting a list of thousands of chapter specs to `/books/{id}/chapters/confirm`, which would write an oversized JSON blob to `books.text`.

Closes #606

## Test plan
- [x] Regression test: POST with 1001 chapter specs → 422
- [x] Full backend suite: 1188 passed

### DIFF
--- a/backend/routers/uploads.py
+++ b/backend/routers/uploads.py
@@ -196,7 +196,7 @@ class ConfirmChapterSpec(BaseModel):
 
 
 class ConfirmChaptersBody(BaseModel):
-    chapters: list[ConfirmChapterSpec]
+    chapters: list[ConfirmChapterSpec] = Field(..., max_length=1000)
 
 
 @router.post("/{book_id}/chapters/confirm")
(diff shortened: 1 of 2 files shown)


alfmunny:
PM review ✅ — `ConfirmChaptersBody.chapters` list cap closes the same 500 MB blob risk category as #520/#524. 1188 backend tests passing, regression test added. Auto-merge enabled.

PR #742, opened 2026-04-23
design: enable PRAGMA foreign_keys per connection for ON DELETE CASCADE (#700)

alfmunny:
## Summary

Path B design doc for issue #700: enable `PRAGMA foreign_keys = ON` on every aiosqlite connection so `ON DELETE CASCADE` actually fires.

Today the schema declares 11 cascading FKs but SQLite silently ignores them because the pragma defaults to OFF per connection. That gap has caused three bugs in April 2026 (#685 / #691 / #695), each fixed by adding a manual `DELETE FROM child_table` ahead of the parent delete. Every such shadow-cascade is a landmine for the next developer.

## What this doc covers

- **Connection hook** — piggyback on the existing busy-timeout monkey-patch in `services/db.py`; wrap `Connection.__aenter__` to issue the pragma once the backing sqlite3 connection is live.
- **Orphan-audit migration 026** — nine row-class cleanups (flashcard_reviews, word_occurrences, user_reading_progress, reading_history, user_books, book_epubs, user_book_chapters, books.owner_user_id) to eliminate pre-existing FK violations before enforcement flips on. Test-backed per the CLAUDE.md migration policy.
- **Migration runner FK-off window** — `services/migrations.run` flips FKs off for its own lifetime so the `CREATE new / INSERT SELECT / RENAME` pattern in e.g. `010_rate_limiter_per_model.sql` keeps working.
- **Remove shadow deletes** — strip the manual `DELETE FROM flashcard_reviews ...` blocks added in #687 / #693 / #695; existing `test_cascade_*` tests prove the engine now does the work.
- **Rollback plan** — land the shadow-delete removal in a separate commit so a partial revert of just the patch is clean.

## Non-goals

- Adding new declared FKs (e.g. `annotations.user_id`) — requires a per-table table-rewrite migration, scoped to its own issue.
- Enforcing FKs during migration runs — explicitly off by design.

## Open questions for PM

1. Preserve orphaned flashcard_reviews rows for history? Proposed: no.
2. Add the missing declared FKs in this PR? Proposed: no, separate issue.
3. Keep the patch in `services/db.py`? Proposed: yes, next to the existing busy-timeout patch.
4. Audit migration performance concern? Rowcounts are small today; bounded one-shot cost.

## Next step

Awaiting PM approval. Implementation PR follows a new issue "feat: implement FK enforcement per design doc" per Path B.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #700, opened 2026-04-23
architecture: enable PRAGMA foreign_keys=ON per connection to enforce ON DELETE CASCADE


alfmunny:
## Problem

SQLite's \`PRAGMA foreign_keys\` defaults to OFF and must be set per connection. As a result, all \`ON DELETE CASCADE\` constraints in the schema (e.g. \`flashcard_reviews.vocabulary_id → vocabulary.id\`, \`word_occurrences.vocabulary_id → vocabulary.id\`) never fire automatically.

This has caused three separate bugs requiring manual \`DELETE FROM flashcard_reviews\` before every vocabulary prune:
- #685 (delete_word)
- #691 (admin delete_book)  
- #695 (delete_user)

## Proposed Solution

Enable FK enforcement in \`_aiosqlite_connect_with_busy_timeout\` (or via an \`aiosqlite\` connection factory) so that every DB connection automatically runs \`PRAGMA foreign_keys = ON\` after opening.

This would make \`ON DELETE CASCADE\` work as intended and eliminate this class of bug entirely. All existing manual cascade deletes that duplicate the \`ON DELETE CASCADE\` behaviour could then be removed.

## Risks / Design Questions

1. **Hidden FK violations**: Existing rows may violate FK constraints that were previously unenforced. A data audit migration may be needed.
2. **Test isolation**: pytest tests that don't seed FK-referenced rows may fail.
3. **Migration compatibility**: Some migration files may INSERT rows that would violate FK constraints if enforcement were ON at that time.
4. **Performance**: FK checks add overhead. Likely negligible for this app's scale.

## Scope

This is a cross-cutting change touching every DB connection in the codebase — warrants a design doc (Path B) before implementation.

## Priority

P3 — current manual delete approach works correctly after PRs #687, #693, #695. FK enforcement is a quality-of-life improvement that prevents future regressions of this pattern.

alfmunny:
Claimed by Architect — starting design doc (Path B) for PRAGMA foreign_keys enforcement.

alfmunny:
Design doc PR filed: #742 — awaiting PM review before implementation begins.

alfmunny:
## Design doc merged (PR #742) — implementation issue notes

Design doc is at `docs/design/fk-enforcement.md`. Implementation can begin now.

### ⚠️ Migration number correction required

The design doc references **`026_fk_orphan_cleanup.sql`** but migration 026 is already taken:

| Number | Feature |
|---|---|
[...]

Issue #695, opened 2026-04-23
fix: delete_user prunes vocabulary without cleaning up flashcard_reviews for other users


alfmunny:
## Bug

In `delete_user()` (`services/db.py`), when a user with uploaded books is deleted:

1. Step 1 (line 280): Cleans up flashcard_reviews for the **deleted user's** vocabulary ✓
2. Step 3 (line 289): Deletes the **deleted user's** vocabulary ✓
3. Step 4 (line 302): Deletes word_occurrences for the **deleted user's owned books** — this affects other users (e.g. admins) who may have saved words from those private books
4. Step 5 (line 304-306): Prunes ALL vocabulary with no remaining occurrences — could delete OTHER users' (e.g. admin) vocabulary entries

After step 5 prunes other users' vocabulary, their `flashcard_reviews` rows are orphaned because the cleanup in step 1 only targeted the deleted user's own vocabulary.

## Scenario

1. Admin saves a word from User A's private uploaded book → admin gets vocabulary + flashcard_reviews
2. User A is deleted
3. Step 4 deletes word_occurrences for User A's book (including admin's occurrence)
4. Step 5 prunes the admin's vocabulary entry (now has no occurrences)
5. Admin's flashcard_reviews row is orphaned → inflated stats

## Fix

Add `DELETE FROM flashcard_reviews WHERE vocabulary_id NOT IN (SELECT DISTINCT vocabulary_id FROM word_occurrences)` between step 4 and step 5 in `delete_user`.

Same pattern used by `delete_uploaded_book` (uploads.py), `delete_book` (admin.py after PR #693), and `delete_word` (db.py after PR #687).

## Priority

P3 — very rare edge case (requires admin saving words from a private user book, then that user being deleted). Only affects flashcard stats display.

alfmunny:
Claimed by Dev — starting work now.

PR #688, opened 2026-04-23
fix: raise vocabulary page lemma word button to 44px touch target

alfmunny:
## What changed

Added `min-h-[44px] flex items-center` to the vocabulary page lemma word button.

The button displayed the word/lemma heading with only `text-base` styling — at ~24px height, it fell well below the 44px minimum touch target required by the design system.

## Tests

New test `VocabLemmaTouchTarget.test.ts` with 2 assertions:
- Lemma button has `min-h-[44px]`
- Lemma button uses `flex items-center` for proper vertical centering

All 1640 tests pass.

Closes #686

PR #751, opened 2026-04-23
feat: enable PRAGMA foreign_keys per connection (#748)

### LINKED ISSUE #748: feat: implement PRAGMA foreign_keys enforcement per design doc (follow-up to #700)
## Context

Design doc merged via PR #742 (issue #700). Implementation can begin.

## ⚠️ Migration number override

The design doc says `026_fk_orphan_cleanup.sql` but **026 is taken by FTS5 search**. Use **`028_fk_orphan_cleanup.sql`** instead.

Migration reservation table:
- 025 = user_book_chapters
- 026 = FTS5 search (merged)
- 027 = vocab tags & decks (in progress)
- **028 = FK enforcement orphan cleanup** ← this PR

## Implementation scope (from `docs/design/fk-enforcement.md`)

1. **Connection hook** — wrap `Connection.__aenter__` in `services/db.py` to issue `PRAGMA foreign_keys = ON` once the backing sqlite3 connection is live (next to the busy-timeout patch).
2. **Migration 028** — nine orphan-row cleanups before FK enforcement flips on. Include `test_migrations.py` test per migration policy.
3. **Migration runner FK-off window** — `services/migrations.run` flips FKs off for its own lifetime.
4. **Remove shadow deletes** — strip manual `DELETE FROM flashcard_reviews`, `vocabulary_tags`, `deck_members` blocks added in PRs #687/#693/#695/#745 (land in a separate commit for clean rollback).

## Testing
- Migration 028 correctness test in `test_migrations.py`
- Integration: confirm cascade fires without manual DELETEs
- Full backend suite must pass

alfmunny:
## Summary

Implements PRAGMA foreign_keys enforcement per the design doc merged in PR #742 (issue #700).

Closes #748.

## What shipped

**Connection hook (commit 1)**
- `services/db.py` patches `aiosqlite.Connection.__aenter__` at the **class level** — instance-level patches are silently ignored by Python's dunder lookup, which is a subtle footgun. Every new `Connection` now issues `PRAGMA foreign_keys = ON` right after the backing sqlite3 connection is live.
- The patch sits alongside the existing busy-timeout monkey-patch so both share the same idempotency pattern.

**Migration runner FK-off window (commit 1)**
- `services/migrations.run` issues `PRAGMA foreign_keys = OFF` on its connection before running any migration. Required because `010_rate_limiter_per_model.sql` uses `CREATE new / INSERT SELECT / DROP old / RENAME` — with FK on, the intermediate state would trip declared constraints.
- Note: `PRAGMA foreign_keys` can only change **outside** a transaction, so the runner issues it as its first statement. App-code connections get FK on because the class-level `__aenter__` wrapper runs before any transaction starts.

**Orphan-audit migration `028_fk_orphan_cleanup.sql` (commit 1)**
- Nine one-shot `DELETE`s and one `UPDATE` (for `books.owner_user_id`, which is nullable) to clean pre-existing orphans. Without this, the first write after deploy would fail on rows that were invalid all along.

**Shadow-delete removal (commit 2 — separate for clean rollback)**
- `services/db.delete_word` (originally #687) and `services/db.delete_user` (originally #695) no longer pre-delete `flashcard_reviews` / `word_occurrences` / `vocabulary_tags` / `deck_members`; the declared `ON DELETE CASCADE` now does it.
- `admin.delete_book`'s `NOT IN` vocabulary-pruning stays untouched — it handles the undeclared `word_occurrences.book_id` relationship, which is separate from cascades.

## Test migrations

**New: `tests/test_fk_enforcement.py`** (4 tests)
- `test_connection_has_fk_on` — proves `PRAGMA foreign_keys` reads `1` on every fresh connection.
- `test_cascade_delete_vocabulary_fires_for_word_occurrences` — raw `DELETE FROM vocabulary` cascades with zero manual child cleanup.
- `test_cascade_delete_user_removes_reading_progress` — `DELETE FROM users` cascades into `user_reading_progress`.
- `test_migrations_run_without_raising_on_fk_violations` + `test_app_connection_fk_on_after_migrations` — the runner's FK-off window doesn't leak into subsequent app connections.

**New: `tests/test_migrations.py`** (3 tests)
- `test_migration_028_cleans_orphan_flashcard_reviews` — seeds an orphan, confirms cleanup.
- `test_migration_028_clears_dangling_book_owner` — confirms `books.owner_user_id` is NULL'd, not deleted.
- `test_migration_028_recorded_in_schema_migrations` — version row lands.

**Fixed: four pre-existing tests that seeded `books.owner_user_id = 1` without creating the user** (`test_db_extended.py`, `test_translation_queue.py`). These started failing under FK enforcement; now they create the owner via `get_or_create_user` first.

**Full backend suite: 1353 passing.**

## Rollback plan

If FK-on causes unexpected production write failures:
- Revert commit 2 to restore the manual cascade `DELETE` code in `delete_word`/`delete_user`.
- Revert commit 1 to drop the class-level `__aenter__` patch and the runner's explicit `PRAGMA foreign_keys = OFF`.
- Migration 028 has already run — its deletes are correct regardless of enforcement, so leaving it applied is harmless.

## Test plan

- [ ] Apply migration 028 on an existing DB with pre-existing orphan rows (flashcard_reviews pointing at deleted vocabulary, etc.) — the orphans are gone after migration.
- [ ] Delete a vocabulary word via `/api/vocabulary/{word}` — `flashcard_reviews`, `word_occurrences`, `vocabulary_tags`, `deck_members` for that word are all removed without manual cleanup in the handler.
- [ ] Admin delete_user — cascades via declared FKs (users → flashcard_reviews via user_id, etc.)
- [ ] Migration 010 still applies cleanly on a fresh DB (FK-off window works).
- [ ] Backend starts up against a prod-sized DB without errors.

### DIFF
--- a/backend/migrations/028_fk_orphan_cleanup.sql
+++ b/backend/migrations/028_fk_orphan_cleanup.sql
@@ -1,0 +1,52 @@
+-- Clean up rows that violate declared FK constraints BEFORE PRAGMA
+-- foreign_keys = ON starts firing on every connection. Runs once. After
+-- this, the engine prevents new orphans from ever being written.
+--
+-- Issue #700 / #748 / design doc docs/design/fk-enforcement.md.
+--
+-- Migration 028 is applied with FK enforcement OFF (the migration runner
+-- explicitly toggles PRAGMA foreign_keys), so these DELETEs don't re-trigger
+-- the very constraints we're enabling.
+
+-- flashcard_reviews.vocabulary_id / user_id
+DELETE FROM flashcard_reviews
+ WHERE vocabulary_id NOT IN (SELECT id FROM vocabulary);
+
+DELETE FROM flashcard_reviews
+ WHERE user_id NOT IN (SELECT id FROM users);
+
+-- word_occurrences.vocabulary_id
+DELETE FROM word_occurrences
+ WHERE vocabulary_id NOT IN (SELECT id FROM vocabulary);
+
+-- user_reading_progress.{user_id, book_id}
+DELETE FROM user_reading_progress
+ WHERE user_id NOT IN (SELECT id FROM users);
+
+DELETE FROM user_reading_progress
+ WHERE book_id NOT IN (SELECT id FROM books);
+
+-- reading_history.user_id (book_id has no declared FK)
+DELETE FROM reading_history
+ WHERE user_id NOT IN (SELECT id FROM users);
+
+-- book_uploads.{user_id, book_id}
+DELETE FROM book_uploads
+ WHERE user_id NOT IN (SELECT id FROM users);
+
+DELETE FROM book_uploads
+ WHERE book_id NOT IN (SELECT id FROM books);
+
+-- book_epubs.book_id
+DELETE FROM book_epubs
+ WHERE book_id NOT IN (SELECT id FROM books);
+
+-- user_book_chapters.book_id
+DELETE FROM user_book_chapters
+ WHERE book_id NOT IN (SELECT id FROM books);
+
+-- books.owner_user_id is NULLable — clear, don't delete the book row.
+UPDATE books
+   SET owner_user_id = NULL
+ WHERE owner_user_id IS NOT NULL
+   AND owner_user_id NOT IN (SELECT id FROM users);
--- a/backend/services/db.py
+++ b/backend/services/db.py
@@ -48,6 +48,39 @@ if not getattr(aiosqlite.connect, _BUSY_TIMEOUT_ATTR, False):
     aiosqlite.connect = _aiosqlite_connect_with_busy_timeout
 
 
+# ── Per-connection FK enforcement (issue #700 / #748) ────────────────────────
+#
+# SQLite's PRAGMA foreign_keys is OFF by default AND must be set per
+# connection. Without this hook every ON DELETE CASCADE in the schema is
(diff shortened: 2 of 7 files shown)


alfmunny:
PM review — implementation looks excellent. Matches the design doc from #742 exactly, migration number 028 is correct.

A few things to confirm before merge:

1. **Shadow-delete removal completeness**: commit 2 removes manual pre-deletes from `delete_word` and `delete_user`. Please confirm no other handler still manually pre-deletes rows that are now covered by CASCADE (a quick `grep -r 'DELETE FROM flashcard_reviews\|DELETE FROM word_occurrences\|DELETE FROM vocabulary_tags\|DELETE FROM deck_members'` across routers/services would do it).
2. **Migration 028 idempotency**: the orphan-cleanup DELETEs run unconditionally. That's fine for a one-shot migration, but confirm `schema_migrations` guards against re-running (standard pattern already in place).
3. **FK-off window scope**: the runner sets `PRAGMA foreign_keys = OFF` on its own connection — good. Confirm this doesn't bleed into concurrent app connections (it shouldn't since it's connection-scoped, but worth a note in the test).

CI still running — will not merge until green. ✅

alfmunny:
Branch is **BEHIND** main (PR #688 just merged). Please rebase and force-push to unblock auto-merge.

PR #767, opened 2026-04-23
fix: preserve chapter-body wrapper div in EPUB path (book 69327 Kafka)

alfmunny:
## Summary

Fixes book 69327 (Kafka, *Der Prozess*) which was rendering only the cover boilerplate + afterword (~1.8 KB) with all 10 actual chapters missing.

## Root cause

The splitter's `_html_body_text` unconditionally skipped nested divs whose class contains `"chapter"` to avoid double-counting in the HTML path (where `build_chapters_from_html` iterates every `*[class~="chapter"]` element via XPath).

That skip is wrong for the EPUB path: `build_chapters_from_epub` picks one spine item per chapter, then Kafka's EPUB stores the chapter body as `<div class="div1 chapter">…</div>` — the wrapper div _is_ the chapter body, not a sub-chapter. The skip dropped 99.6% of the novel.

## Fix

- Add opt-in `skip_nested_chapter_divs` flag to `_html_body_text`, default `False`. Propagated through both the blockquote and the fall-through div recursion so deeply-nested wrappers stay consistent.
- `build_chapters_from_html` passes `skip_nested_chapter_divs=True` — existing behaviour preserved.
- `build_chapters_from_epub` leaves it at `False`.

## Verification

- Regression test `test_html_body_text_preserves_div_chapter_wrapper_in_epub` in `tests/test_chapter_source.py` covers the Kafka-shaped `<div class="div1 chapter">` in both modes.
- Existing `test_html_body_text_skips_nested_chapter_div` updated to pass the flag explicitly — confirms HTML-path behaviour unchanged.
- Full backend suite: **1360 passing.**
- Local verify against `gutenberg.org/ebooks/69327.epub3.images`: **12 chapters, 467 KB** (was 2 chapters, 1.8 KB).

## Test plan

- [ ] After deploy + cold start, reload https://book-reader-ai.vercel.app/reader/69327 — ten numbered Kapitel + Nachwort visible, with chapter titles from the NCX (ERSTES KAPITEL VERHAFTUNG · …, etc.).
- [ ] Book 2229 (Faust, HTML-path) still splits correctly — no regression on the HTML splitter.
- [ ] Book 24288 (Rilke, post-#758 poetry fix) still renders with EPUB content.

### DIFF
--- a/backend/services/splitter.py
+++ b/backend/services/splitter.py
@@ -412,9 +412,14 @@ def build_chapters_from_html(html: str) -> list[Chapter]:
                     subtitle_elem = nxt
 
         # Extract body text from all children except the title heading
-        # (and the subtitle paragraph, if any).
+        # (and the subtitle paragraph, if any). HTML path enables
+        # skip_nested_chapter_divs — we iterate every chapter div via
+        # XPath at the top of this function, so nested ones would be
+        # double-processed if we let them through here.
         body_text = _html_body_text(
-            div, skip_first_heading=True, skip_elems=(subtitle_elem,) if subtitle_elem is not None else (),
+            div, skip_first_heading=True,
+            skip_elems=(subtitle_elem,) if subtitle_elem is not None else (),
+            skip_nested_chapter_divs=True,
         )
         word_count = len(body_text.split())
 
@@ -488,6 +493,7 @@ def _html_body_text(
     *,
     skip_first_heading: bool = False,
     skip_elems: tuple = (),
+    skip_nested_chapter_divs: bool = False,
 ) -> str:
     """Extract readable text from an HTML element, preserving paragraph breaks.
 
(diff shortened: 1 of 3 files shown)


alfmunny:
PM review: ✅ Approved. Root cause is well-diagnosed — the `skip_nested_chapter_divs` flag cleanly separates HTML-path and EPUB-path behaviour without touching the external interface. Kafka's *Der Prozess* going from 1.8 KB to 467 KB is a significant content-quality fix. Regression tests cover both modes. Auto-merge is fine once CI passes.

Follow-up: this is the second EPUB-specific splitter fix this session (#758 + #767). Worth filing a low-priority issue to run a systematic audit of other Gutenberg books on the EPUB path to catch similar structural patterns we may have missed.

PR #771, opened 2026-04-23
fix: suppress raw parser exception details in file upload 422 response

### LINKED ISSUE #761: fix: suppress raw parser exception details in file upload 422 response
## Problem

`backend/routers/uploads.py` line 124:

```python
except Exception as exc:
    logger.exception("Failed to parse uploaded file")
    raise HTTPException(status_code=422, detail=f"Could not parse file: {exc}")
```

Raw exception text from the EPUB/TXT parser is included in the 422 response. Parser library errors may include internal details (file offsets, internal state, library version info) that should not be exposed.

## Fix

Replace `detail=f"Could not parse file: {exc}"` with `detail="Could not parse uploaded file"` — the full exception is already logged via `logger.exception`.

## Test plan

Add a test that mocks the parser to raise an exception and asserts the 422 detail does not contain the raw exception string.

alfmunny:
## What changed
- `routers/uploads.py`: the outer exception handler in `upload_book` that was `raise HTTPException(status_code=422, detail=f"Could not parse file: {exc}")` is now `raise HTTPException(status_code=422, detail="Could not parse uploaded file")` — `exc` string dropped
- `tests/test_router_uploads.py`: new test `test_upload_parse_error_does_not_leak_exception_detail` patches `services.book_parser.parse_txt` to raise a RuntimeError with a sentinel string and asserts the sentinel does not appear in the 422 response detail

## Why
The parse failure path in the upload handler forwarded the raw exception message (which could contain internal file parser details, path info, or library error strings) directly into the 422 HTTP response visible to the uploading user.

## Test plan
- [x] `test_upload_parse_error_does_not_leak_exception_detail` — 422 returned, "internal-parser-secret-xyzzy" absent from detail
- [x] Full uploads test file: 34 passed

Closes #761

### DIFF
--- a/backend/routers/uploads.py
+++ b/backend/routers/uploads.py
@@ -119,9 +119,9 @@ async def upload_book(
             parsed = parse_txt(file_bytes.decode("utf-8", errors="replace"))
         else:
             parsed = parse_epub(file_bytes)
-    except Exception as exc:
+    except Exception:
         logger.exception("Failed to parse uploaded file")
-        raise HTTPException(status_code=422, detail=f"Could not parse file: {exc}")
+        raise HTTPException(status_code=422, detail="Could not parse uploaded file")
 
     chapters = parsed["chapters"]
     if not chapters:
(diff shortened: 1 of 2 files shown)


alfmunny:
PM review ✅ — correct fix. Drops the raw `exc` string from the 422 detail, replacing it with a static message. Consistent with the exception-leak series (#759, #763, #764, #768). Regression test correctly uses a sentinel string to confirm the internal detail is absent from the response. Auto-merge enabled — will self-merge once CI clears.

PR #774, opened 2026-04-23
chore: remove redundant shadow-deletes for vocabulary_tags/deck_members in delete_book

### LINKED ISSUE #755: chore: remove shadow-cascade workaround in vocab services now that FK enforcement is live
## Context

PR #745 (vocab tags & decks backend) added explicit shadow-deletes for `vocabulary_tags` and `deck_members` rows in `delete_word` and `admin delete_book`. The PR comment states:

> These shadow-deletes will be reverted once #700 (FK enforcement) ships.

PR #751 shipped FK enforcement (PRAGMA foreign_keys=ON per connection) and merged to main.

## Required cleanup

In `backend/services/` (vocabulary.py, decks.py) and `backend/routers/admin.py` (or wherever `admin delete_book` lives):

1. Remove explicit `DELETE FROM vocabulary_tags WHERE vocabulary_id = ?` calls in `delete_word`.
2. Remove explicit `DELETE FROM deck_members WHERE vocabulary_id = ?` calls in `delete_word`.
3. Remove equivalent shadow-deletes in `admin delete_book`.

The ON DELETE CASCADE constraints on `vocabulary_tags(vocabulary_id)` and `deck_members(vocabulary_id)` now handle these automatically.

## Notes
- Keep the shadow-deletes in `delete_user` path unchanged — those are working through the user row, which cascades differently.
- Add a regression test confirming cascade fires (or verify existing migration tests cover it).

Closes #748 shadow work.

alfmunny:
## What changed
- `routers/admin.py` `delete_book`: removed two redundant explicit-delete queries for `vocabulary_tags` and `deck_members` that ran immediately before deleting the parent `vocabulary_entries` / `deck_cards` rows. Those parent deletes already cascade to the child tables via FK `ON DELETE CASCADE`, so the shadow queries were dead code.

## Why
The shadow deletes gave a false impression that FK cascades were not in place and added unnecessary DB round-trips on every book deletion. Relying on the FK cascade is both correct and cleaner.

## Test plan
- [x] Existing `test_delete_book_*` suite covers book deletion end-to-end; no new behavior introduced, so no new tests needed
- [x] Full backend suite: 1374 passed

Closes #755

### DIFF
--- a/backend/routers/admin.py
+++ b/backend/routers/admin.py
@@ -258,14 +258,7 @@ async def delete_book(book_id: int = Path(..., ge=1), _admin: dict = Depends(_re
             "DELETE FROM flashcard_reviews WHERE vocabulary_id NOT IN "
             "(SELECT DISTINCT vocabulary_id FROM word_occurrences)"
         )
-        await db.execute(
-            "DELETE FROM vocabulary_tags WHERE vocabulary_id NOT IN "
-            "(SELECT DISTINCT vocabulary_id FROM word_occurrences)"
-        )
-        await db.execute(
-            "DELETE FROM deck_members WHERE vocabulary_id NOT IN "
-            "(SELECT DISTINCT vocabulary_id FROM word_occurrences)"
-        )
+        # vocabulary_tags and deck_members cascade automatically via FK enforcement (#751).
         await db.execute(
             "DELETE FROM vocabulary WHERE id NOT IN (SELECT DISTINCT vocabulary_id FROM word_occurrences)"
         )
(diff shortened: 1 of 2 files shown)


alfmunny:
PM review ✅ — correct cleanup. The shadow-deletes for `vocabulary_tags` and `deck_members` in `delete_book` are genuinely dead code given the FK `ON DELETE CASCADE` enforcement added in PR #751. Removing them reduces DB round-trips and removes a misleading implication that cascades weren't trusted. Existing delete-book tests remain sufficient. Auto-merge enabled.

PR #775, opened 2026-04-23
fix: suppress raw exception details in vocabulary Obsidian export 500 response

### LINKED ISSUE #760: fix: suppress raw exception details in vocabulary Obsidian GitHub export 500 response
## Problem

`backend/routers/vocabulary.py` line 482–483:

```python
except Exception as e:
    raise HTTPException(status_code=500, detail=str(e))
```

The outer catch in the Obsidian/GitHub export endpoint (`POST /api/vocabulary/export/obsidian`) exposes raw exception text from GitHub API calls in the 500 response. GitHub errors can include partial token details, rate-limit messages, repo path information, etc.

**Severity:** P3 (low) — endpoint is user-authenticated; the detail is only visible to the user whose token is being used. However, it violates the consistent static-message posture being applied to all other 500 responses.

## Fix

Replace `except Exception as e: raise HTTPException(status_code=500, detail=str(e))` with `except Exception: raise HTTPException(status_code=500, detail="Export to GitHub failed")`

## Test plan

Add a test that mocks `_github_put` to raise an exception and asserts the 500 detail does not contain the raw error string.

alfmunny:
## What changed
- `routers/vocabulary.py` `export_obsidian`: changed `except Exception as e: raise HTTPException(status_code=500, detail=str(e))` to `except Exception: raise HTTPException(status_code=500, detail="Export to GitHub failed")` — raw exception text (which may include GitHub tokens, repo paths, or network error details) no longer leaks into the 500 response body.
- `tests/test_router_vocabulary.py`: added `test_export_obsidian_error_does_not_leak_exception_detail` — patches `_github_put` to raise a RuntimeError with a sentinel string and asserts the sentinel is absent from the 500 response detail.

## Why
The Obsidian GitHub export `except Exception as e` clause forwarded `str(e)` directly to the HTTP client. A GitHub API error or network failure could expose the repo name, token hint, or internal path in the response body.

## Test plan
- [x] `test_export_obsidian_error_does_not_leak_exception_detail` — 500 with static message, no sentinel in detail
- [x] Full vocabulary test suite: 42 passed

Closes #760

### DIFF
--- a/backend/routers/vocabulary.py
+++ b/backend/routers/vocabulary.py
@@ -479,5 +479,5 @@ async def export_obsidian(
             return {"urls": urls}
     except HTTPException:
         raise
-    except Exception as e:
-        raise HTTPException(status_code=500, detail=str(e))
+    except Exception:
+        raise HTTPException(status_code=500, detail="Export to GitHub failed")
--- a/backend/tests/test_router_vocabulary.py
+++ b/backend/tests/test_router_vocabulary.py
@@ -759,3 +759,19 @@ async def test_export_obsidian_zero_book_id_returns_422(client, test_user):
     """Regression #734: POST /vocabulary/export/obsidian with book_id=0 must return 422."""
     resp = await client.post("/api/vocabulary/export/obsidian", json={"book_id": 0})
     assert resp.status_code == 422, f"Expected 422 for book_id=0 in export, got {resp.status_code}: {resp.text}"
+
+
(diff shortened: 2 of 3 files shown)


alfmunny:
PM review ✅ — correct fix and highest-priority one in the series: `str(e)` in the Obsidian export handler could leak GitHub token hints, repo paths, or network error details. Sentinel-based regression test is the right approach. 42 vocabulary tests clean. Auto-merge enabled.

PR #779, opened 2026-04-23
fix: add min_length=1 bounds to QueueSettingsRequest list elements and RetryFailedRequest.target_language (closes #776)

### LINKED ISSUE #776: bug: QueueSettingsRequest list elements and RetryFailedRequest.target_language accept empty strings
## Problem

`QueueSettingsRequest.auto_translate_languages` and `QueueSettingsRequest.model_chain` list elements lack `min_length=1`, so an empty string `""` passes Pydantic validation and gets stored in the DB / used as a model name.

`RetryFailedRequest.target_language` is `str | None` with only `max_length=20` — empty string passes and causes the retry query to match zero rows silently instead of being rejected.

## Files

- `backend/routers/admin.py` lines ~1053, 1057, 1234

## Fix

Add `min_length=1` to per-element annotation in `auto_translate_languages` and `model_chain`, and to `RetryFailedRequest.target_language`.

## Priority

P2 — no data loss, but allows silent bad data into queue settings and confusing retry behavior.

alfmunny:
## What changed
- `QueueSettingsRequest`: added `min_length=1` to `auto_translate_languages` list element type and `model_chain` list element type — empty strings are now rejected at the Pydantic level (422) before reaching business logic.
- `RetryFailedRequest`: added `min_length=1` to `target_language` field — empty string is rejected with 422.
- `tests/test_router_admin.py`: added three regression tests for the new bounds; updated existing `test_queue_settings_rejects_model_chain_with_empty_entry` to expect 422 (Pydantic now fires before the 400 business-logic check).

## Why
`auto_translate_languages`, `model_chain`, and `target_language` in the queue settings/retry-failed endpoints accepted empty strings. Storing an empty model name or language code caused downstream AI calls to fail with cryptic errors.

## Test plan
- [x] `test_queue_settings_empty_auto_translate_language_returns_422` — 422 for `[""]` in auto_translate_languages
- [x] `test_queue_settings_empty_model_chain_entry_returns_422` — 422 for `[""]` in model_chain
- [x] `test_retry_failed_empty_target_language_returns_422` — 422 for `target_language: ""`
- [x] Full backend suite: 1387 passed
- [x] Full frontend suite: 1750 passed

Closes #776

### DIFF
--- a/backend/routers/admin.py
+++ b/backend/routers/admin.py
@@ -1043,11 +1043,11 @@ async def queue_get_settings(_admin: dict = Depends(_require_admin)):
 class QueueSettingsRequest(BaseModel):
     enabled: bool | None = None
     api_key: str | None = Field(default=None, max_length=500)
-    auto_translate_languages: list[Annotated[str, Field(max_length=20)]] | None = Field(default=None, max_length=50)
+    auto_translate_languages: list[Annotated[str, Field(min_length=1, max_length=20)]] | None = Field(default=None, max_length=50)
     rpm: int | None = Field(default=None, ge=1)
     rpd: int | None = Field(default=None, ge=1)
     model: str | None = Field(default=None, max_length=200)
-    model_chain: list[Annotated[str, Field(max_length=200)]] | None = Field(default=None, max_length=20)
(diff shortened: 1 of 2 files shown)


alfmunny:
Checking in — PR #779 is showing BLOCKED. CI is still running. Pinging PM for review status when CI completes.

PR #781, opened 2026-04-23
chore: gate /loop behind CLAUDE.md startup + add Local-work-first rule

alfmunny:
## What changed
Each of the four role prompts in `scripts/start-roles.sh` (`PM`, `Dev`, `UI/UX Dev`, `Architect`) now begins with a `SESSION STARTUP` paragraph that delegates to CLAUDE.md §Session startup (steps 1–6) and explicitly names the per-role variants (worktree path, memory path, PR-first walk, stale-claim cleanup for PM). The existing `/loop` invocation runs only after that startup is complete.

## Why
On every fresh role launch, the model received a prompt that began with `/loop Nm <work-prompt>`. The work-prompt then said `"Every cycle, follow this order strictly: (1) FIRST check your own open PRs..."`. The model treated steps (1)→(2) as the canonical workflow and never detoured through CLAUDE.md §Session startup, even though that section says **"Do not skip step 5."**

Concretely, every role was skipping:
- Reading individual memory files (only the `MEMORY.md` index is auto-loaded)
- The PM-only warning when `in-progress` issues exist without a worktree
- The full PR-first walkthrough (BEHIND / failing CI / BLOCKED triage)
- Stale-claim cleanup (PM)

The PM prompt was the worst offender — it didn't even mention CLAUDE.md.

`/loop` itself doesn't differentiate first tick from subsequent ticks; the cleanest fix is at the prompt-file layer: do startup once at the top, then schedule the recurring loop.

## Test plan
- [x] `bash -n scripts/start-roles.sh` — syntax OK
- [x] Diff inspected; all four prompts retain their original `/loop` work-prompt content unchanged
- [ ] Manual: `bash scripts/start-roles.sh restart` after merge — verify each role runs the startup sequence on its first turn before scheduling its `/loop`

## Restart required
Running roles will not pick up the new prompts until restarted:

```
bash scripts/start-roles.sh restart
```


alfmunny:
Checking in — PR #781 is showing BLOCKED. CI is still running. Pinging PM for review status when CI completes.

PR #782, opened 2026-04-23
ux: redesign AnnotationToolbar as centered modal, remove deprecated onAnnotate from SentenceReader

alfmunny:
### AnnotationToolbar — modal redesign
- Changed from a cursor-positioned tooltip panel to a fixed full-screen modal with a semi-transparent backdrop; removes fragile `position: { x, y }` prop and `window.innerWidth/Height` math
- Backdrop div has `onClick={onClose}` so clicking outside closes the modal
- Removed the `position` prop entirely from the component and all callers in `page.tsx`
- Added: NoteIcon header ("Add note" / "Edit note"), quoted sentence preview, `aria-pressed` on colour swatches, `TrashIcon` icon-only delete button with `aria-label="Delete note"`, textarea auto-focus, `animate-fade-in` entrance
- Save button now reads "Save note" (new) or "Update" (existing)

### SentenceReader — `onAnnotate` removal
- Removed deprecated `onAnnotate` prop and `handlePointerDown` long-press path; notes are now opened via SelectionToolbar "Note" button (`onNote`) or annotation-click → QuickHighlightPanel → "Add note"
- Pointer-event handlers are now only wired when `onWordTap` is provided

### `page.tsx`
- Removed `position` from `annotationPanel` state shape and all three `setAnnotationPanel` call sites
- Removed `onAnnotate` prop pass-through to `SentenceReader`

## Why

The old tooltip-style panel caused UX problems on mobile (could render off-screen, used `window.innerWidth/Height` which is unavailable in SSR/tests). The modal pattern is standard for note-taking on touch devices. The `onAnnotate` long-press path was a parallel annotation entry point that bypassed the SelectionToolbar flow.

## Test plan
- [x] All 1746 frontend tests pass
- [x] `AnnotationToolbar` suite: 24 tests (backdrop click, Escape, save/update/delete, error paths, colour picker)
- [x] `AnnotationToolbarTouchTarget`: 3 touch-target (44px) tests
- [x] `SentenceReader` pointer-move tests rewritten to use `onWordTap`
- [x] `ReaderPage.branches2` updated to use `trigger-note` / `qhp-open-note` for annotation flows

### DIFF
--- a/frontend/src/__tests__/AnnotationToolbar.branches.test.tsx
+++ b/frontend/src/__tests__/AnnotationToolbar.branches.test.tsx
@@ -26,7 +26,6 @@ const BASE_PROPS = {
   sentenceText: "It is a truth universally acknowledged.",
   chapterIndex: 0,
   bookId: 1,
-  position: { x: 100, y: 200 },
   onClose: jest.fn(),
   onSaved: jest.fn(),
   onDeleted: jest.fn(),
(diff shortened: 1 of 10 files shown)


alfmunny:
Approved. Solid redesign — modal pattern + removal of `window.innerWidth/Height` math is exactly right for mobile/SSR. 24 dedicated AnnotationToolbar tests + the deletion-heavy diff (-304 lines of redundant onAnnotate path) reads well.

Minor: branch name `fix/vocab-page-unicode-touch` doesn't match the AnnotationToolbar work — looks like the branch was repurposed. No action needed for this PR; just a note for next time so the branch name stays a useful search key.

PR #780, opened 2026-04-23
fix: strip Gutenberg EPUB frontmatter + duplicate chapter-title body prefix (#778)

alfmunny:
Closes #778.

## What changed

Three targeted filters in `backend/services/splitter.py` `build_chapters_from_epub()`:

1. `_epub_toc_containers(body)` drops `<table>` / `<ol>` whose rows are ≥60% `<a class="pginternal">` internal-anchor TOC links. Catches Faust's pg-header title-page chapter-0 content.
2. `_epub_frontmatter_blocks(body)` drops `<div>` / `<section>` carrying Gutenberg frontmatter class markers (`titlepage`, `cover`, `frenchtitle`, `copyright`, `dedication`, …). Catches Kafka pg-header.
3. `_strip_title_from_body_prefix(text, title)` drops a leading paragraph that normalizes to the chapter title, preventing `"FAUST: Der Tragödie erster Teil"` from showing up twice in the body of the Nacht scene.

## Why

Both 2229 (Faust) and 69327 (Der Prozess) rendered a rogue chapter 0 built from title-page / copyright / TOC HTML, shifting every real chapter by one and misaligning cached translations. The existing `pg-boilerplate` filter only catches `<section class="pg-boilerplate">`, which misses both the TOC table and the class-tagged div frontmatter.

## Test plan

- [x] `test_chapter_source.py` (68 new lines) covers all three mechanisms with representative HTML fixtures.
- [x] Full backend suite: **1388 passed**.

## Local verification (manual)

- Faust (2229): 28 chapters (was 29); chapter 0 = Zueignung. Chapter 3 (Nacht) body no longer repeats the chapter title.
- Kafka (69327): 11 chapters (was 12); chapter 0 = ERSTES KAPITEL.
- Rilke (24288) poetry: unchanged at 2 chapters — no regression.

## Follow-up

Translation-cache index re-alignment for Faust chapters 5+ (Studierzimmer split + previously cached translations shifted by one) is a separate data-migration problem, tracked on the audit issue.

PR #790, opened 2026-04-24
fix: migration 029 invalidates shifted chapter cache for Faust + Kafka after #780 (closes #783)

### LINKED ISSUE #783: bug: translation cache misaligned for Faust (#2229) and Kafka (#69327) after #780 splitter change
## Background

PR #780 fixes EPUB splitter to drop rogue chapter-0 frontmatter for two real-world books:

- **Faust (Gutenberg #2229)**: chapter count 29 → 28 (chapter 0 was TOC HTML).
- **Der Prozess / Kafka (Gutenberg #69327)**: chapter count 12 → 11 (chapter 0 was titlepage/copyright).

Forward fix is correct, but **existing translation caches keyed on `chapter_index` are now misaligned by one** for any user who previously translated chapters 1+ on either book.

## Symptom (after #780 merges)

Open Faust → chapter 5 → see translation that was originally for chapter 6 content. Same for chapter 6 = chapter 7 cache, etc. Persists for any cached chapter ≥ 1 in the affected books.

## Scope of affected data

- `translation_cache` rows where `book_id ∈ (2229, 69327)` and `chapter_index ≥ 1`.
- Possibly also: `book_insights` (chapter-indexed), `book_chapter_summaries`, any per-chapter annotations / vocabulary occurrences if they reference `chapter_index` rather than text content.

## Options

1. **Invalidate**: `DELETE FROM translation_cache WHERE book_id IN (2229, 69327) AND chapter_index >= 1`. Simplest. Users re-pay for translation. Acceptable scope (two books).
2. **Re-index by content match**: for each shifted cache row, find the matching new chapter by fuzzy-matching chapter title or first paragraph. More work; preserves user investment.
3. **Migration fingerprint**: track per-book a `splitter_version` and invalidate caches when the version changes. Future-proofs against the next splitter bug — see also #769 audit.

## Recommendation

Option (1) for the two known books as a quick migration; option (3) tracked separately as architecture follow-up to #769. Splitter changes shouldn't silently corrupt user-visible cached content again.

## Priority

**P2** — affects only users who pre-translated chapters of these two specific books before #780. Volume is likely small.

alfmunny:
## What changed

Adds `backend/migrations/029_invalidate_shifted_chapter_cache.sql` which deletes stale chapter-indexed rows for Faust (Gutenberg #2229) and Der Prozess / Kafka (Gutenberg #69327) where `chapter_index >= 1`.

Tables cleaned:
- `translations` — primary chapter translation cache
- `chapter_summaries` — AI-generated chapter summaries
- `translation_queue` — pending/failed queue rows with wrong indices
- `book_insights` — per-chapter AI insights (NULL chapter rows preserved)

## Why

PR #780 fixed the EPUB splitter to drop a rogue chapter-0 frontmatter page for these two books, reducing their chapter counts by one. Any cached data keyed on `chapter_index >= 1` was recorded before the fix and is now shifted by one — readers would see the wrong translated content for every chapter after the first. This migration clears those rows so they are re-requested and re-translated with correct indices.

Chapter 0 rows and rows for all other books are untouched.

## Test plan

- New regression test `test_migration_029_clears_shifted_translations` in `test_migrations.py`: seeds stale rows across all four affected tables, re-runs migration 029, and verifies stale rows deleted while chapter-0 and unrelated-book rows survive.
- Full backend suite: 1394 passed.
- Full frontend suite: 1754 passed.

Closes #783

### DIFF
--- a/backend/migrations/029_invalidate_shifted_chapter_cache.sql
+++ b/backend/migrations/029_invalidate_shifted_chapter_cache.sql
@@ -1,0 +1,24 @@
+-- Invalidate chapter-indexed cache rows for Faust (Gutenberg #2229) and
+-- Der Prozess / Kafka (Gutenberg #69327) after PR #780 corrected the EPUB
+-- splitter to drop the rogue chapter-0 frontmatter. All cached rows with
+-- chapter_index >= 1 are now off by one and must be cleared so readers
+-- see correct content on re-request.
+--
+-- Issue #783.
+
+DELETE FROM translations
+ WHERE book_id IN (2229, 69327)
+   AND chapter_index >= 1;
+
+DELETE FROM chapter_summaries
+ WHERE book_id IN (2229, 69327)
+   AND chapter_index >= 1;
+
+DELETE FROM translation_queue
+ WHERE book_id IN (2229, 69327)
+   AND chapter_index >= 1;
+
+DELETE FROM book_insights
+ WHERE book_id IN (2229, 69327)
+   AND chapter_index IS NOT NULL
+   AND chapter_index >= 1;
--- a/backend/tests/test_migrations.py
+++ b/backend/tests/test_migrations.py
@@ -1055,3 +1055,84 @@ async def test_migration_028_recorded_in_schema_migrations(tmp_db):
             "SELECT version FROM schema_migrations WHERE version = '028_fk_orphan_cleanup'"
         ) as cur:
             assert (await cur.fetchone()) is not None
+
+
+# ── Migration 029 (issue #783): invalidate shifted chapter cache ──────────────
+
+
+async def test_migration_029_clears_shifted_translations(tmp_db):
+    """Regression #783: migration 029 must delete translations with chapter_index >= 1
+    for Faust (#2229) and Kafka (#69327), but leave chapter 0 and other books untouched."""
+    await run_migrations(tmp_db)
+
+    async with aiosqlite.connect(tmp_db) as db:
+        await db.execute("PRAGMA foreign_keys = OFF")
+        await db.executemany(
+            "INSERT INTO translations (book_id, chapter_index, target_language, paragraphs) "
+            "VALUES (?, ?, 'zh', '[]')",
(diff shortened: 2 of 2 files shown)


alfmunny:
PM review:

The migration explicitly preserves `chapter_index = 0` for both books (and the test asserts this), but I want to make sure that's intentional rather than an oversight.

After #780 dropped the rogue chapter-0 frontmatter, the index shifted by one for these two books:
- old index 0 = TOC HTML / titlepage (now gone)
- old index 1 = real chapter 1 → **new index 0**
- old index 2 = real chapter 2 → new index 1
- ...

So a user who previously read chapter 0 (the rogue page) and triggered a translation would have a row at `chapter_index=0` containing the *old TOC translation*. After #780 they reopen Faust, scroll to the new chapter 0 (= real first chapter), and the translation for the OLD rogue page surfaces — same misalignment as #783 just at index 0.

If chapter 0 was never realistically translatable (e.g. TOC HTML produced no paragraphs and the cache is empty), then leaving it alone is fine and the existing test is correct. But if there could be any cached `chapter_index=0` rows for #2229 / #69327, I'd extend the DELETE to `chapter_index >= 0` (i.e. `book_id IN (2229, 69327)` only) and update the test accordingly.

Could you confirm before this merges? Either way, no other concerns — the SQL is correctly scoped, the test covers the survivor invariants for unrelated books and NULL chapter book_insights, and the table list is right.

alfmunny:
Adding `blocked` label to hold this from auto-merging until the chapter_index=0 question above is resolved. Once you confirm intent (or fix the SQL + test), I'll remove the label and let the watcher land it. Sorry for the friction — migrations are irreversible in production so I'd rather catch the question pre-merge.

alfmunny:
Disabled auto-merge directly because the `blocked` label alone doesn't gate auto-merge in this repo (workflow gap — filing a separate issue). Re-enable auto-merge once the chapter_index=0 question is resolved: `gh pr merge --auto --squash 790`.

PR #793, opened 2026-04-24
fix: ai_lookup returns empty url instead of broken Wiktionary link

### LINKED ISSUE #551: AI definition response: url field points to wiktionary even when source is AI
## Problem

`GET /vocabulary/definition/{word}` AI fallback (PR #549) returns:

```json
{
  "lemma": "Aufmerksamkeitsdefizit",
  "source": "ai",
  "url": "https://en.wiktionary.org/wiki/Aufmerksamkeitsdefizit"
}
```

The `url` field points to wiktionary even though the AI fallback triggered specifically because wiktionary had **no entry** for this word. A user who clicks the URL lands on a missing/empty wiktionary page, which is confusing.

## Expected

When `source == "ai"`, return `"url": ""` (or `null`) so the frontend can decide whether to show a "look up" link at all.

## Suggested fix

In `services/gemini.py:define_word`, change:

```python
"url": f"https://en.wiktionary.org/wiki/{word}",
```
to:
```python
"url": "",
```

The frontend already handles the empty-URL case (no link rendered).

Filed as follow-up from PR #549 review.

alfmunny:
Closes #551

When Wiktionary has no entry for a word, the AI fallback (`ai_lookup`) still returned `url: https://en.wiktionary.org/wiki/<word>` — a link to a page that doesn't exist. Clicking it would send users to a Wiktionary 404.

**Fix:** return `url: ""` from `ai_lookup` (`services/wiktionary.py:154`) so the frontend omits the look-up link entirely when AI was used.

## Test plan

- [x] `test_ai_lookup_url_is_empty_string` — new regression test asserts `url == ""` when Gemini fallback is used
- [x] All existing `ai_lookup` tests still pass (26 total in `test_wiktionary.py`)
- [x] Frontend suite: 1754 tests passing
- [x] Backend suite: 1394 tests passing

### DIFF
--- a/backend/services/wiktionary.py
+++ b/backend/services/wiktionary.py
@@ -151,7 +151,7 @@ async def ai_lookup(word: str, lang: str, api_key: str) -> dict:
             "lemma": data.get("lemma") or word,
             "language": lang,
             "definitions": definitions,
-            "url": wikt_url,
+            "url": "",
         }
     except Exception:
         return empty
--- a/backend/tests/test_wiktionary.py
+++ b/backend/tests/test_wiktionary.py
@@ -366,3 +366,15 @@ async def test_ai_lookup_returns_empty_on_exception():
     with patch("services.gemini._generate", new=AsyncMock(side_effect=Exception("api error"))):
         result = await ai_lookup("word", "en", api_key="test-key")
     assert result["definitions"] == []
+
+
+@pytest.mark.asyncio
+async def test_ai_lookup_url_is_empty_string():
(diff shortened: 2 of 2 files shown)


alfmunny:
PM ping: this PR has been BEHIND for ~33 min with no CI in flight — last rebase was before #801/#802 landed, so main has moved several times since. Same situation on #803 (BEHIND, idle). Could you run the per-cycle step 1 from CLAUDE.md (rebase BEHINDs before submitting new work) so these two can land? Once #793 and #803 merge and clear from your count, #808 will auto-merge on its own. No code concerns — the diffs are clean; just unblock the watcher.

PR #832, opened 2026-04-24
feat: EPUB split quality audit script (#769)

alfmunny:
## Summary

Adds `backend/scripts/epub_split_audit.py` — a diagnostic that compares the character count produced by the **EPUB-path splitter** vs the **plain-text-path splitter** for every book with a stored EPUB. Books where the EPUB split yields less than 50% of the text split's character count are flagged as likely splitter edge cases.

## Why

Two silent-data-loss EPUB splitter bugs fixed in quick succession:

- **PR #758** — span-wrapped verse lines dropped (poetry EPUBs, e.g. Rilke #24288)
- **PR #767** — chapter-body wrapper div skipped (div-based EPUBs, e.g. Kafka #69327)

Both left books that looked fine in the catalog but had nearly empty chapters once opened. This script would have caught both before users noticed.

## What changed

- **`backend/scripts/epub_split_audit.py`**
  - `compute_audit_row(...)` — pure function: runs both splitters against supplied text/bytes, returns an `AuditRow` with `epub_chars`, `text_chars`, `ratio`, `suspicious`. Testable without a DB.
  - `run_audit(threshold, book_id=None)` — walks `list_cached_books()`, resolves each book's EPUB bytes + cached text, emits one `AuditRow` per book that actually has an EPUB. Books with no EPUB are silently skipped (nothing to audit); books with an EPUB but no cached text are reported but not flagged (no baseline).
  - CLI flags: `--book-id`, `--threshold` (default 0.5), `--csv`. Exit code is `1` when any book is flagged so this can double as a CI data-quality gate later.
- **No production code touched.** No new dependencies, no schema changes.

## Test plan

- [x] **New `backend/tests/test_epub_split_audit.py`** (8 tests):
  - threshold logic: flags below, safe above, honours custom threshold, graceful when text is empty
  - DB iteration: only books with stored EPUBs are audited
  - `main()` CLI: exit `1` on suspicious rows, exit `0` on clean runs, CSV output contains header + flagged row
- [x] Full backend suite: **1408 passed**.

## How to run

```bash
# Audit everything
python -m scripts.epub_split_audit

# Single book
python -m scripts.epub_split_audit --book-id 69327

# Write CSV + stricter threshold
python -m scripts.epub_split_audit --threshold 0.7 --csv /tmp/audit.csv
```

Closes #769.

Issue #820, opened 2026-04-24
bug: EPUB splitter drops in-paragraph line breaks (Faust #2229 ch. 14, FAUST 'O schaudre nicht!' speech)


alfmunny:
## Reproduction

URL: `https://book-reader-ai.vercel.app/reader/2229?chapter=14` (Faust I — "15. Garten")

In the FAUST speech beginning "O schaudre nicht!", what should be **7 verse lines plus a stage direction** all render as a single collapsed paragraph:

> FAUST. O schaudre nicht! Laß diesen Blick, Laß diesen Händedruck dir sagen Was unaussprechlich ist. Sich hinzugeben ganz und eine Wonne Zu fühlen, die ewig sein muß! Ewig!—Ihr Ende würde Verzweiflung sein Nein, kein Ende! Kein Ende! (Margarete drückt ihm die Hände, macht sich los und läuft weg. Er steht einen Augenblick in Gedanken, dann folgt er ihr.)

Expected:

> FAUST. O schaudre nicht! Laß diesen Blick,
> Laß diesen Händedruck dir sagen
> Was unaussprechlich ist.
> Sich hinzugeben ganz und eine Wonne
> Zu fühlen, die ewig sein muß!
> Ewig!—Ihr Ende würde Verzweiflung sein
> Nein, kein Ende! Kein Ende!
> (Margarete drückt ihm die Hände, macht sich los und läuft weg. Er steht einen Augenblick in Gedanken, dann folgt er ihr.)

## Why interesting

The bug is **intermittent within the same chapter**. The FAUST stanza immediately above ("Ja, mein Kind! Laß dieses Blumenwort / Dir Götterausspruch sein. Er liebt dich!") renders one verse line per visual line correctly. The MARTHE stanza further down ("Ich bät Euch, länger hier zu bleiben,...") also renders correctly. So the splitter handles most verse stanzas right; only specific markup patterns lose their line breaks.

## Suspected cause

Likely `<br/>` (or `<br>`) inside a single `<p>` element — the EPUB markup uses paragraph-internal line breaks for some passages, and our parser collapses them when extracting paragraph text. The good stanzas above/below probably use one `<p>` per verse line, which works.

Different from #758 (span-wrapped verse lines, fixed) and #767 (chapter-body wrapper div, fixed). Likely same EPUB-splitter quality category as #769's audit proposal.

## Action

[...]

Issue #769, opened 2026-04-23
audit: run systematic EPUB split quality check across Gutenberg catalog


alfmunny:
## Background

Two EPUB splitter bugs fixed in quick succession:
- PR #758: span-wrapped verse lines dropped (poetry EPUBs, e.g. Rilke #24288)
- PR #767: chapter-body wrapper div skipped (div-based EPUBs, e.g. Kafka #69327)

Both were silent data-loss bugs — books appeared to have chapters but were nearly empty.

## Proposed audit

Write a one-time script (or extend `scripts/backfill_epubs.py`) that:
1. For each book with a stored EPUB (`book_epubs` table), compare the character count of the EPUB-path split vs the plain-text-path split.
2. Flag books where EPUB split yields <50% of the plain-text character count — likely candidates for another splitter edge case.
3. Report: book_id, title, epub_chars, text_chars, ratio.

This would have caught both #757 and the Kafka issue before users noticed.

## Labels
`enhancement` / `architecture` — script lives in `backend/scripts/`; no production code change needed.

alfmunny:
Triaged: applying `architecture` label per the issue body's own suggestion. Backend audit script lives in `backend/scripts/` and is squarely Architect's lane — Architect can claim on next cron tick.

alfmunny:
Claimed by Architect — Path A (script-only, no production code change). Will add `backend/scripts/epub_split_audit.py` that compares EPUB-path vs plain-text-path character counts per book, flags rows under 50% ratio, and writes a CSV + summary. Includes a pytest covering the ratio-computation and the 'suspicious' flagging so the audit logic itself is regression-protected.

PR #839, opened 2026-04-24
feat: paragraph-count + structural speaker-cue signals for EPUB audit (#834)

### LINKED ISSUE #834: feat: extend EPUB split audit to paragraph-count / structural anomalies (follow-up to #769/#832)
## Summary

PR #832 adds `epub_split_audit.py` which compares **character counts** between the EPUB and plain-text splitters — great for catching the #758 / #767 class (content dropped entirely). But **#820's class of bug is invisible to a char-count audit**: when `<br>` inside `<p>` collapses 7 verse lines + stage direction into one paragraph, the character count stays intact — only the paragraph count changes.

This is a real gap. Faust would score ~100% on the current audit yet has 108 collapsed-verse paragraphs across 21 chapters (#820).

## Proposed extension

Add a second check to `epub_split_audit.py` (or a sibling script): compare the **paragraph count** between EPUB-path and plain-text-path splits, flagging books where the EPUB path yields substantially fewer paragraphs than the plain-text path. A 50% ratio is wrong here — the threshold should probably be something like "EPUB paragraphs < 80% of plain-text paragraphs."

Even better: a **structural signal** — flag any paragraph longer than N chars (~400) that contains a speaker-cue pattern `^[A-ZÄÖÜ ]+\.$` on an embedded newline. That directly measures the #820 pattern without needing a plain-text baseline.

## Scope suggestion

- Extend `compute_audit_row()` to return `para_count_epub`, `para_count_text`, `suspicious_paragraphs` (list of (chapter_index, first-100-chars) for the "long paragraph with embedded speaker cue" rule).
- Keep the current char-count flag; add paragraph-count + structural flags.
- Update the CSV / CLI output to show all three categories.

## References

- Parent audit: #769 (closed by #832)
- Bug this would catch: #820 (Faust verse-collapse umbrella — 108 paragraphs across 21 chapters)
- Root cause: `_split_dramatic_speakers()` not wired into the EPUB path in `backend/services/splitter.py`

## Labels

`enhancement` / `architecture` — diagnostic extension, lives in `backend/scripts/`, no production code change.

alfmunny:
## Summary

Extends the EPUB split audit (#832 / `scripts/epub_split_audit.py`) with two additional signals that catch bugs **invisible to a char-count comparison** — specifically the #820 class, where `<br>` inside `<p>` collapses verse into one visual block, leaving the char count intact but dropping the paragraph count by an order of magnitude.

Three independent signals now:

1. **Character ratio** (existing, from #769) — `epub_chars / text_chars < --threshold` (default 0.5).
2. **Paragraph ratio** (new) — `epub_paragraphs / text_paragraphs < --para-threshold` (default 0.8). Uses blank-line paragraph counting against both splitter outputs.
3. **Structural speaker-cue collapse** (new, #820-specific) — any paragraph longer than `--structural-paragraph-len` chars (default 400) that contains an embedded-newline all-caps speaker cue like `"\n  HELENA."`. No plain-text baseline required.

`main()` now exits `1` when **any** signal fires, so a single-book regression like #820 no longer slips the CI gate.

## Why

The char-count audit I shipped in #832 cleanly catches the #758 / #767 class (verse-dropping, wrapper-div skipping). But #820 showed it has a blind spot: the collapsed-verse paragraphs in Faust keep 100% of the characters — they just rearrange them. A paragraph-count check catches that shift directly; the structural signal catches it without needing any baseline.

## What changed

- **`backend/scripts/epub_split_audit.py`**
  - `AuditRow` gains `epub_paragraphs`, `text_paragraphs`, `para_ratio`, `paragraph_suspicious`, and `structural_flags: list[(chapter_index, excerpt)]`. All new fields default-initialize so the constructor remains backward-compatible with the existing test-stub rows.
  - `is_flagged` property ORs the three signals and drives the CLI exit code.
  - `_count_paragraphs()` and `_find_structural_flags()` helpers; speaker-cue regex `\n[ \t]*[A-ZÄÖÜ]{2,}[A-ZÄÖÜ ]*\.` (requires ≥2 uppercase chars to skip initials/single letters).
  - `compute_audit_row()` populates all three signals. When the plain-text baseline is empty, ratio signals stay off but structural flags are still evaluated — they stand on their own.
  - CSV output extended with six new columns; CLI summary breaks down flag count per signal.
  - New flags: `--para-threshold`, `--structural-paragraph-len`.
- **`backend/tests/test_epub_split_audit.py`**
  - Pre-existing tests (8) still pass unchanged.
  - +10 new tests covering paragraph-count, structural-flag, `is_flagged`, CSV columns, and the `main()` exit-code gate on a structural-only flag (the #820 case).

## Test plan

- [x] Full audit test module: **18 passed** (was 8).
- [x] Full backend suite after rebase onto main (now containing #832): **1419 passed**.
- [x] Manual spot-check: a Faust-like paragraph with `\n  HELENA.` embedded in a >400-char block produces one structural flag, zero char/para flags, and `main()` returns 1.

## How to run

```bash
# All signals
python -m scripts.epub_split_audit

# Single book
python -m scripts.epub_split_audit --book-id 2229

# Tighter paragraph gate + CSV
python -m scripts.epub_split_audit --para-threshold 0.9 --csv /tmp/audit.csv

# Looser structural threshold (catch shorter verse collapses)
python -m scripts.epub_split_audit --structural-paragraph-len 250
```

## References

- Parent audit: #769 (closed by #832)
- Bug this signal catches: #820 (Faust verse-collapse umbrella — 108 paragraphs across 21 chapters)

Closes #834.

### DIFF
--- a/backend/scripts/epub_split_audit.py
+++ b/backend/scripts/epub_split_audit.py
@@ -1,19 +1,32 @@
-"""EPUB split quality audit (issue #769).
+"""EPUB split quality audit (issues #769, #834).
 
-Compares the character counts produced by the EPUB-path splitter and the
-plain-text-path splitter for every book with a stored EPUB. Books where the
-EPUB split yields substantially fewer characters than the text split are
-flagged as likely hitting a splitter edge case (see #758, #767 for the
-regressions that motivated this audit).
+Compares the **character count** and **paragraph count** produced by the
+EPUB-path splitter and the plain-text-path splitter for every book with a
+stored EPUB, and surfaces a third **structural** signal for the specific
+"verse-collapse" pattern that bit us in #820 — paragraphs where an embedded
+newline is followed by an all-caps speaker cue, indicating the EPUB path
+collapsed poetry / drama into a single visual block.
+
+Three signals, independently flagged:
+
+- **Character ratio** (#769 / #758 / #767) — EPUB chars drop substantially
+  below the plain-text baseline. Default gate: < 50%.
+- **Paragraph ratio** (#834) — EPUB paragraphs fewer than the plain-text
+  paragraph count. Invisible to the char-count check. Default gate: < 80%.
+- **Structural speaker-cue collapse** (#834 / #820) — a paragraph exceeds
+  `--structural-paragraph-len` characters AND contains an embedded newline
+  followed by an all-caps name + period (`\\n  HELENA.`). No plain-text
+  baseline needed.
 
 Usage:
     python -m scripts.epub_split_audit                      # all books, stdout
     python -m scripts.epub_split_audit --book-id 69327      # single book
     python -m scripts.epub_split_audit --csv out.csv        # write CSV report
-    python -m scripts.epub_split_audit --threshold 0.7      # stricter gate
+    python -m scripts.epub_split_audit --threshold 0.7      # stricter char gate
+    python -m scripts.epub_split_audit --para-threshold 0.9 # stricter para gate
 
-Exit code is 1 when at least one book is flagged — the script can be wired
-into CI as a data-quality gate.
+Exit code is 1 when at least one book is flagged by any of the three
+signals — the script can be wired into CI as a data-quality gate.
 """
 
 from __future__ import annotations
@@ -21,8 +34,9 @@ from __future__ import annotations
 import argparse
 import asyncio
 import csv
+import re
 import sys
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Iterable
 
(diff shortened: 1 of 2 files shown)


alfmunny:
PM approval — ship it.

**What I checked:**
- Scope matches #834: paragraph-count + structural speaker-cue signals that catch the #820 class (chars intact, paragraphs collapsed).
- Three independent signals OR'd via `is_flagged` — `main()` exits 1 if any fires, so #820-shaped regressions can't slip a future CI gate.
- `AuditRow` extension is backward-compatible (new fields default-initialize).
- Structural signal is standalone — doesn't need a plain-text baseline, which is the right call for cases where the baseline is also absent or empty.
- Regex `\n[ \t]*[A-ZÄÖÜ]{2,}[A-ZÄÖÜ ]*\.` requires ≥2 uppercase chars; skips single-letter initials (good false-positive mitigation).
- 10 new tests including an explicit `main()` exit-code test on a structural-only flag — that's exactly the Faust case.
- 18 audit tests pass; full backend suite at 1419 after rebase on #832.

**No changes requested.** Straight ship.

Issue #971, opened 2026-04-24
architecture: CI median still 318s after Opts 2+5 — investigate pytest/E2E reductions to hit 180s target


alfmunny:
## Status

CI speedup umbrella #885 has shipped Opt 2 (#919, path-scoping) and Opt 5 (#922, docker cache). Opt 3 was closed `wontfix` (coverage wanted on PRs). Opt 4 (sharding) was deferred.

**Post-merge measurement (12 PRs, #919..#963): median = 318 s.**

The design doc target was **≤ 180 s** (`docs/design/ci-speedup.md`). We are 76% above target.

## Why the miss

Per-job breakdown on a representative UX-only PR (#963):

| Job | Runtime |
|---|---|
| Backend tests (pytest) | **4m 11s** |
| Frontend E2E (Playwright) | **4m 14s** |
| Frontend tests (Jest) | 2m 13s |
| Docker build | *skipped* ✅ |

Wall time ceiling = slowest always-running job ≈ 4m 14s = 254s + coverage comment ≈ 318s median.

Backend pytest runs on frontend-only PRs by design: the Opt 2 implementation was amended during review (#919 discussion) to gate `test-backend` on `frontend || backend` per user decision ("option 1 and 2" — want integration safety over speed). So we can't revert that.

## Options

Any one of these could hit the target without reverting the test-backend gating:

1. **Speed up backend pytest** from 4 min to ≤ 2 min
   - `pytest-xdist` parallel workers
   - Profile slow tests (`--durations=20`)
   - Reduce fixture cold-start cost
2. **Reduce Playwright E2E scope on PRs** to a smoke suite (~30s); run full E2E on `push: main` only
3. **Opt 4 sharding** (per original design, deferred) — split pytest/E2E across 2–4 runners

## Recommendation

Start with (2) — cheapest to implement and matches the pattern used for docker-build (skip on PRs, full on main). Playwright at 4m14s is a hot spot; most UX PRs don't need the full E2E suite. Needs a smoke-suite tag and the workflow change to filter on it.

## Priority

P3 — the current 318s is acceptable for the team, and the `needs-revision` label on #885 already captures this gap. Filing so it doesn't get lost if/when the team wants to push further.

See `reports/ci_baseline_2026_04_24.md` for measurements.

alfmunny:
[...]

PR #919, opened 2026-04-24
feat: CI path-scoped job gates — frontend / backend / docker (#885 Opt 2)

alfmunny:
## Summary

**Opt 2 of 4** from the merged CI speedup design doc (`docs/design/ci-speedup.md`, #901). Single-file diff on `.github/workflows/ci.yml`.

Split the single `code` paths-filter output into three path-scoped outputs — `frontend`, `backend`, `docker` — and gate each heavy job on one. Baseline already in place at `reports/ci_baseline_2026_04_24.md` (#914).

## What changed

Before:

```yaml
changes:
  outputs:
    code: ${{ steps.filter.outputs.code }}
  filters: |
    code:
      - 'backend/**'
      - 'frontend/**'
      - '.github/workflows/**'

test-frontend: { if: needs.changes.outputs.code == 'true' }
test-e2e:      { if: needs.changes.outputs.code == 'true' }
test-backend:  { if: needs.changes.outputs.code == 'true' }
verify-docker: { if: needs.changes.outputs.code == 'true' }
```

After:

```yaml
changes:
  outputs:
    frontend: ${{ steps.filter.outputs.frontend }}
    backend:  ${{ steps.filter.outputs.backend }}
    docker:   ${{ steps.filter.outputs.docker }}
  filters: |
    frontend: [ 'frontend/**', '.github/workflows/ci.yml' ]
    backend:  [ 'backend/**',  '.github/workflows/ci.yml' ]
    docker:   [ 'backend/Dockerfile', 'docker-compose.yml',
                'backend/requirements.txt', 'backend/pyproject.toml',
                '.github/workflows/ci.yml' ]

test-frontend: { if: needs.changes.outputs.frontend == 'true' }
test-e2e:      { if: needs.changes.outputs.frontend == 'true' }
test-backend:  { if: needs.changes.outputs.backend  == 'true' }
verify-docker: { if: outputs.docker == 'true' || outputs.backend == 'true' }
coverage-comment: { if: outputs.frontend == 'true' || outputs.backend == 'true' }
```

**Belt-and-braces rule**: the workflow file (`.github/workflows/ci.yml`) is listed in every filter. Any edit to this file runs the full matrix — we never skip the jobs we need to see the workflow itself pass.

## Expected saving

[...]

Issue #885, opened 2026-04-24
architecture: accelerate CI pipeline (path-scoping, parallel sharding, coverage-skip, docker cache)


alfmunny:
## Problem

Every PR currently sits in `BLOCKED` state for 5–8 minutes while the four heavy CI jobs run: Frontend Jest, Frontend E2E (Playwright), Backend pytest, and Verify Docker build. A workflow/docs-only PR like #862 incurs the full cost even though it touches nothing executable (see #875). With the current 1–2 PR-per-cycle throughput, we burn roughly 20 CI-minutes of wall clock per productive cycle, and that's the bottleneck that bounds how fast the multi-role pipeline can ship.

This architecture issue proposes five ordered optimizations. Architect drafts `docs/design/ci-speedup.md`; implementation lands in separate PRs per optimization, each shippable independently once the plan is approved.

### 1. Land #875 — CI paths-filter fix (P1, already filed, no new scope here)

Fixing the existing paths-filter so CLAUDE.md / docs-only diffs correctly return `code=false` skips all four heavy jobs for every workflow PR. This is the **zero-risk, highest-payoff** change and should land **before** any of the items below so they can be measured against a correct baseline.

Dependency, not part of this design's scope. Noted here for sequencing.

### 2. Path-scoped CI splits

Extend `dorny/paths-filter` with additional outputs:

- `backend`: `backend/**`, `backend/tests/**`, `pyproject.toml`, `pytest.ini`
- `frontend`: `frontend/**`, `frontend/package.json`, `frontend/package-lock.json`
- `docker`: `backend/Dockerfile`, `docker-compose.yml`

Then gate jobs:

- Frontend Jest + Frontend E2E: `if: needs.changes.outputs.frontend == 'true'`
- Backend pytest: `if: needs.changes.outputs.backend == 'true'`
- Verify Docker build: `if: needs.changes.outputs.docker == 'true' || needs.changes.outputs.backend == 'true'`

Expected saving: ~50% of PRs touch only frontend or only backend, so about half of cycles skip 3–5 minutes of jobs.

Risks:
- Shared-contract test suites (OpenAPI, JSON-schema) may need one of the job filters widened. Verify coverage during design.
- Integration bugs where a frontend-only change surfaces a backend issue — mitigate by still running the full suite on push to main.

### 3. Parallel test sharding

- **Jest**: split the suite across N shards via `--shard=i/N` and a matrix job. 2–3 shards typically cut wall time by 40–60%.
[...]

PR #968, opened 2026-04-24
design: EPUB NCX fragment-anchor chapter boundaries (#964)

alfmunny:
## Summary

Design doc for #964 — Path B first step. Documents the problem, proposed fix, cache-invalidation concerns, and rollout plan for the largest class of chapter-structure bug in the Gutenberg catalog.

## Why

Full-catalog investigation (`reports/epub_parser_investigation_2026_04_24.md`) found every inspected Gutenberg EPUB (7/7) uses NCX fragment anchors that the current spine-walking splitter ignores:

- #62215 Le Fantôme de l'Opéra: 36 navPoints → **4 chapters** (9× collapse)
- #25575 Mémoires d'Outre-Tombe: 450 navPoints → **8 chapters** (56× collapse)
- #43759 Geflügelte Worte: 814 navPoints → **20 chapters** (40× collapse)

Single largest parser-quality issue in the catalog.

## Risks called out

- Chapter re-indexing invalidates `translations` / `audio_cache` / `chapter_summaries` / `book_insights` / `annotations` cache keys (same failure mode as #780/#783 Faust + Kafka incident).
- Design gates re-splitting behind a future `splitter_version` column + explicit `resplit_book.py <book_id>` CLI with `--confirm`.
- No automatic re-split on parser update.

## Not in scope

- Implementation (separate PR after PM sign-off).
- Companion bugs #965 (single-chapter EPUB rejection) and #966 (CJK text fallback).

## Test plan

- Pure docs change (`docs/design/epub-ncx-fragment-anchors.md` only) — no code impact.
- Path-scoped CI should skip frontend/backend test suites.

Awaiting PM review of the design. Implementation blocked until design merges per CLAUDE.md Path B.

PR #990, opened 2026-04-24
design: CI speedup phase 2 — close the 318s → 180s gap (#971)

alfmunny:
## Summary

Design doc for #971 — Path B first step. Proposes a staged approach to close the 318s → 180s CI median gap left by phase 1 (#885).

## Why

Post-merge measurement over 12 PRs (#919..#963) shows median 318s, 76% over the phase-1 target. Root cause is backend pytest at 4m 11s running on every frontend PR (per the user-approved gating from the #919 review) plus Playwright E2E at 4m 14s. Both need work.

## Proposal

Staged, measure-between-stages:

1. **Stage 1** — Patch three `worker_idles_*` tests that burn 10s each on real-time sleeps + four gutenberg retry tests burning 2s each on exponential backoff. Profiled with `pytest --durations=20`. Estimated −35s.
2. **Stage 2** — Adopt `pytest-xdist -n auto` for parallel workers. Estimated additional −60s to −90s.
3. **Stage 3 (conditional)** — Playwright smoke/full split: smoke on PR, full on `push: main`. Estimated −20s to −40s.
4. **Stage 4 (conditional)** — Opt 4 sharding (deferred in phase 1). Only if Stage 3 doesn't close the gap.

Keeps the user-approved `test-backend` frontend||backend gating unchanged — integration safety stays.

## Test plan

- Pure docs change (`docs/design/ci-speedup-phase-2.md` only).
- Path-scoped CI should skip frontend/backend test suites.

## Not in scope

- Implementation (each stage is its own separate PR, gated on PM sign-off).

### DIFF
--- a/docs/design/ci-speedup-phase-2.md
+++ b/docs/design/ci-speedup-phase-2.md
@@ -1,0 +1,128 @@
+# CI Speedup Phase 2 — Closing the 318s → 180s Gap
+
+Issue: #971
+Prior work: #885 (phase 1, merged). Phase 1 design: `docs/design/ci-speedup.md`.
+Baseline: `reports/ci_baseline_2026_04_24.md`.
+
+## Problem
+
+Phase 1 shipped Opt 2 (path-scoping) and Opt 5 (docker cache) and closed out #921 (coverage-skip) `wontfix`. Measured post-merge median across 12 PRs (#919..#963): **318s**. Design target was **≤ 180s**.
+
+Gap: **138s / 76%** over target.
+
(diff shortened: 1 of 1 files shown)


alfmunny:
**PM**: 6th Path B gate-skip of the day (same class as #968). Filed + merged without `needs-user-approval` on either #971 or #990. Content appears reasonable (follow-up to close the 318s→180s gap reported in #973). Retroactively applied `needs-user-approval` on #971. Root cause unchanged: #908 role-session restart still pending → roles still on pre-#910 CLAUDE.md.

PR #1029, opened 2026-04-24
fix: use DOMPurify to sanitize FTS5 snippets in search results (closes #1027)

### LINKED ISSUE #1027: bug: search SnippetHtml regex allows <b>-with-attributes XSS — use DOMPurify instead
## Bug

`frontend/src/app/search/page.tsx` uses `dangerouslySetInnerHTML` to render FTS5 snippets from the backend. The comment claims snippets contain "only `<b>…</b>` wrappers", but SQLite's `snippet()` function embeds raw source text without HTML-escaping it.

### Attack vector

A user stores an annotation or vocabulary note with text like:
```
<b onmouseover="alert(document.cookie)">click me</b>
```
Then searches for `click`. The FTS5 snippet function returns:
```
<b onmouseover="alert(document.cookie)"><b>click</b> me</b>
```

### Why the regex fails

The sanitization regex on line 13:
```js
const clean = snippet.replace(/<(?!\/?b\b)[^>]*>/g, "");
```
The negative lookahead `(?!\/?b\b)` only checks that the tag starts with `b` at a word boundary. `<b onmouseover=...>` passes this check because `b\b` matches (`b` followed by a space is a word boundary). So `<b onmouseover="alert(1)">` is NOT stripped.

### Severity

Self-XSS only (search results are always scoped to the authenticated user's own content). However:
- The code comment is wrong — it's not safe to render arbitrary user content via `dangerouslySetInnerHTML`
- Could escalate if search is ever made cross-user

### Fix

Install `dompurify` (or `isomorphic-dompurify`) and use it with an allowlist:
```tsx
import DOMPurify from "isomorphic-dompurify";
const clean = DOMPurify.sanitize(snippet, { ALLOWED_TAGS: ["b"] });
```

This correctly strips attributes from `<b>` tags and removes any other tags entirely.

alfmunny:
## What changed

The `SnippetHtml` component in `frontend/src/app/search/page.tsx` used `dangerouslySetInnerHTML` with a hand-rolled regex that could be bypassed.

**The vulnerability:** SQLite's `snippet()` function embeds raw user content without HTML-escaping. A user could store an annotation or vocabulary note containing `<b onmouseover="alert(document.cookie)">click</b>`, then search for it. The old regex (`/<(?!\/?b\b)[^>]*>/g`) passed `<b onmouseover="...">` through unchanged because the negative lookahead only checked that the tag started with `b` at a word boundary — not that it was exactly `<b>` with no attributes.

**The fix:** Replace the regex with `isomorphic-dompurify` configured with `ALLOWED_TAGS: ["b"], ALLOWED_ATTR: []`. This correctly strips all attributes from `<b>` tags and removes all other HTML including `<script>` content.

## Why

Self-XSS only (search results are scoped to the authenticated user's own content), but the comment in the code was incorrect and the sanitization was insufficient. The correct fix is a proper allowlist sanitizer, not a regex.

## Test plan

- Added `src/__tests__/SearchPageSnippetXss.test.tsx` with 3 regression tests:
  1. `<b onmouseover="...">` → attribute stripped, `<b>` tag kept
  2. `<script>alert(1)</script>` → tag and content removed entirely  
  3. Normal FTS5 `<b>highlight</b>` → preserved correctly
- Added `src/__mocks__/isomorphic-dompurify.ts` jest-compatible mock (the real package has an ESM transitive dep that Jest can't transform)
- Full frontend suite: 2001 passed ✓
- Full backend suite: 1517 passed ✓

Closes #1027

### DIFF
--- a/frontend/jest.config.js
+++ b/frontend/jest.config.js
@@ -9,6 +9,7 @@ const customConfig = {
   moduleNameMapper: {
     "^react-markdown$": "<rootDir>/src/__mocks__/react-markdown.tsx",
     "^remark-gfm$": "<rootDir>/src/__mocks__/remark-gfm.ts",
+    "^isomorphic-dompurify$": "<rootDir>/src/__mocks__/isomorphic-dompurify.ts",
     "^@/(.*)$": "<rootDir>/src/$1",
   },
   collectCoverageFrom: [
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -15,6 +15,7 @@
         "@capacitor/ios": "^8.3.1",
         "@capacitor/keyboard": "^8.0.3",
         "@capacitor/status-bar": "^8.0.2",
+        "isomorphic-dompurify": "^3.10.0",
         "next": "14.2.3",
         "next-auth": "^5.0.0-beta.30",
         "react": "^18",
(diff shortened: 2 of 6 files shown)


alfmunny:
**PM**: Good XSS fix — DOMPurify with `ALLOWED_TAGS:['b'], ALLOWED_ATTR:[]` is the right pattern for FTS5 snippets.

One flag: this PR touches `backend/services/auth.py` and `backend/tests/test_auth_service_branches.py` with the exact same diff as still-open **#1026** (both show `index 58cd4b44..fb60a5a5` for auth.py). Looks like the branch was based off one that already had #1026 applied. Not blocking — whichever merges first will make the second's duplicate lines a no-op on rebase — but worth noting so neither PR is surprised by the conflict/empty-hunk on rebase.

Fix itself is fine, labels correct. Approving the XSS scope.

Issue #1053, opened 2026-04-24
bug: EPUB splitter uses book section heading as chapter title (Faust ch 3 shows 'FAUST: Der Tragödie erster Teil' instead of 'Nacht')


alfmunny:
## Problem

On production `book-reader-ai.vercel.app/reader/2229?chapter=3` (Faust Part 1), the displayed chapter title is:

> **FAUST: Der Tragödie erster Teil**

But this is the EPUB's *section-level* heading ("Faust: First Part of the Tragedy") — the book's overall part title — not the actual chapter title. The real chapter title is **"Nacht"** (Night), which is currently rendered as the first line of the body text (see attached screenshot).

The body correctly starts with:
```
Nacht

In einem hochgewölbten, engen gotischen Zimmer Faust,
unruhig auf seinem Sessel am Pulte.

FAUST.
Habe nun, ach! Philosophie,
...
```

So the splitter picked the parent section heading as the chapter title and then failed to promote "Nacht" out of the body into the title slot.

## Related history

- **PR #780** (merged) fixed Gutenberg-frontmatter + duplicate-title-body-prefix for plain-text uploads, but this is the **EPUB path**.
- **PR #905** wired `_split_dramatic_speakers` into the EPUB path to fix verse collapse, but didn't revisit title detection.
- **PR #832** (#769 audit) + **PR #889** (backfill) scanned for chunk-level issues but the audit checked paragraph-count / speaker-cue signals, not title-promotion correctness.

## Proposed investigation

1. Examine `backend/services/splitter.py` → `build_chapters_from_epub` → how it picks `chapter.title` from EPUB spine items vs. `<h1>`/`<h2>` inside the HTML document body.
2. The EPUB for Gutenberg Faust (#2229) likely has a multi-level TOC where a top-level spine item maps to "FAUST: Der Tragödie erster Teil" but the real chapters ("Nacht", "Vor dem Tor", "Studierzimmer", etc.) are subsection headings *inside* the same spine item. The splitter currently treats the spine item's metadata as the chapter title instead of descending into the subsection.
3. Fix direction: if the body's first paragraph is a title-like heading (bold/h1/h2/all-caps/short), AND the current chapter `title` looks like a section-level title (e.g., contains ":" or matches the book-title prefix pattern), promote the first body line to `chapter.title` and strip it from the body.

## Affected books

[...]

PR #1055, opened 2026-04-24
feat: EPUB NCX fragment-anchor chapter segmentation (#964)

### LINKED ISSUE #964: architecture: EPUB splitter ignores NCX fragment-anchor chapter boundaries (Le Fantôme #62215, Mémoires d'Outre-Tombe #25575)
## Problem

`build_chapters_from_epub` walks the EPUB **spine** and emits one chapter per spine item. But many Gutenberg EPUBs pack multiple navigable chapters into a single xhtml spine item, encoding chapter boundaries as **fragment anchors (`#id`) in the NCX navMap** instead of as separate spine items.

When this happens, the splitter collapses N navPoints into 1 chapter, destroying the book's chapter structure. The reader sees one gigantic chapter.

## Reproduction (cached Gutenberg EPUBs)

| Book | Real chapters (from NCX) | Splitter output | Collapse ratio |
|---|---|---|---|
| #62215 Le Fantôme de l'Opéra | 36 navPoints across 5 xhtml files | **4 chapters**, first = 194 KB | 9× under-split |
| #25575 Mémoires d'Outre-Tombe | 450 navPoints across 7 xhtml files | **8 chapters**, first = 138 KB | 56× under-split |
| #77700 Entstehung der Alchemie | 693 navPoints across 81 files | 92 chapters | 7× under-split |
| #43759 Geflügelte Worte | 814 navPoints across 21 files | 20 chapters | 40× under-split |

Every inspected Gutenberg EPUB (7/7 in the sample) uses fragment anchors for at least some nav points.

## Root cause

`backend/services/splitter.py:611`:
```python
for item_id, _ in book.spine:
    item = id_to_item.get(item_id)
    ...
    # one chapter per spine item
```

The NCX is read only for **titles** (`_epub_nav_titles`) but its `content src="file.xhtml#anchor"` structure — which is authoritative chapter boundary data — is never used to segment xhtml content.

## Proposed fix direction

When a spine item contains multiple NCX fragment anchors:
1. Parse xhtml with lxml
2. Split the DOM at each `#anchor` id under `body`
3. Emit one chapter per anchor-delimited slice, with NCX's navLabel as title

Needs a design doc because:
- Affects how existing cached books re-split (potentially invalidates translation cache keys — see #783 for prior art)
- Nested navPoints (depth=3 in Alchemie) need flattening rules
- Must not regress already-correct books (one-anchor-per-file cases)

## Impact

High — this is the single largest class of chapter-structure bug across the catalog. Every major 19th-century French novel we've inspected exhibits it, plus nearly all German reference works.

alfmunny:
## Summary

Implements the design in `docs/design/epub-ncx-fragment-anchors.md` (merged #968, user-approved). Closes the single largest class of chapter-structure bug in the Gutenberg catalog.

## What this fixes

When an EPUB's NCX targets a single xhtml spine item with multiple `#anchor` navPoints, the splitter was collapsing all of them into one chapter. The canonical repro is Le Fantôme de l'Opéra (#62215): 36 navPoints, 5 xhtml files, splitter produced **4 chapters**. With this change, **30 chapters**.

## Measured impact (91 cached Gutenberg EPUBs)

| | before | after |
|---|---:|---:|
| 0-char books | 8 | **0** |
| Suspicious char/paragraph-ratio flags | 25 | **0** |

| Book | before | after |
|---|---:|---:|
| #62215 Le Fantôme de l'Opéra | 4 | 30 |
| #25575 Mémoires d'Outre-Tombe | 8 | 21 |
| #43759 Geflügelte Worte | 20 | 45 |

## What's new in the code

- `_epub_nav_anchors(book)` — ordered per-file anchor list, preserving NCX play-order.
- `_segment_body_by_anchors(body, anchors, book_title)` — one-pass DOM walk, slices top-level body children into per-anchor sub-bodies. Resolves `id=` and legacy `<a name="...">` alike. Returns [] on any missing-anchor error so segmentation falls back to whole-file instead of dropping content silently.
- `_extract_and_clean_chapter_text` — shared cleanup pipeline so slice path and fallback path match.
- `build_chapters_from_epub` gated on `len(anchors) >= 2 and any(…)`; single-anchor spine items keep current behaviour.

## Not in this PR

- `splitter_version` column on books + cache-invalidation gate. Design explicitly punted these to a follow-up — new imports use the new parser, existing books stay frozen until an explicit `resplit_book.py` CLI ships.
- `resplit_book.py` CLI — same follow-up.

## Test plan

Six new tests in `test_splitter.py`:
- multi-anchor single-file segmentation produces correct slice count
- slice content is partitioned correctly (no bleed)
- single-anchor files keep one-chapter behaviour (regression guard)
- missing DOM anchor falls back to single chapter (no silent drop)
- NCX order ≠ DOM order → output in DOM order
- Integration check against cached Le Fantôme EPUB: ≥20 chapters

**Full suite: 1528 passed.** Catalog-wide audit (`scripts/epub_split_audit`) shows 0 0-char books and 0 suspicious-ratio flags after the change.

Closes #964.

### DIFF
--- a/backend/services/splitter.py
+++ b/backend/services/splitter.py
@@ -595,6 +595,7 @@ def build_chapters_from_epub(epub_bytes: bytes) -> list[Chapter]:
         return []
 
     nav_titles = _epub_nav_titles(book)
+    nav_anchors = _epub_nav_anchors(book)
     book_title = book.title or ""
     id_to_item = {
         item.get_id(): item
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: Major win — +451/-24 in splitter alone, but the measured impact table across 91 cached EPUBs (0-char 8→0, suspicious flags 25→0, Fantôme 4→30 chapters) is the right framing. Design doc #968 merged with `user-approved`, so implementation proceeds on the proper Path B track.

One cross-reference worth checking: open issue **#1053** (Faust #2229 ch 3 displays 'FAUST: Der Tragödie erster Teil' instead of 'Nacht'). If Faust's NCX contains navPoints like `section.xhtml#nacht`, `section.xhtml#studierzimmer`, `section.xhtml#vordem_tor`, this PR plausibly fixes #1053 as a side effect — the NCX-anchor segmentation would produce separate chapters with the correct titles pulled from NCX text. Worth verifying post-merge on book #2229 and closing #1053 if true.

Applied `P2` label (was missing). No other concerns — ship it.

Issue #754, opened 2026-04-23
architecture: add declared FKs to soft user_id / book_id columns (follow-up to #700)


alfmunny:
## Problem

PR #751 landed PRAGMA foreign_keys enforcement (issue #700). Every **declared** `ON DELETE CASCADE` now fires automatically. But the schema still has many tables whose `user_id` / `book_id` columns are *soft* references — no `REFERENCES` clause at all — which means user/book deletion still requires manual cascade logic in `delete_user` and `admin.delete_book`.

### Tables with undeclared user_id / book_id columns

From a schema audit (`grep REFERENCES` across `backend/migrations/*.sql`):

| Table | Columns missing FK | Cascade should be |
|---|---|---|
| `annotations` | `user_id`, `book_id` | `ON DELETE CASCADE` |
| `vocabulary` | `user_id` | `ON DELETE CASCADE` |
| `word_occurrences` | `book_id` | `ON DELETE CASCADE` |
| `book_insights` | `user_id`, `book_id` | `ON DELETE CASCADE` |
| `chapter_summaries` | `book_id` | `ON DELETE CASCADE` |
| `translations` | `book_id` | `ON DELETE CASCADE` |
| `audio_cache` | `book_id` | `ON DELETE CASCADE` |
| `translation_queue` | `book_id`, `queued_by` (user_id) | `ON DELETE CASCADE` |

This is the explicit non-goal the #700 design doc deferred: *"Adding new declared FKs requires per-table table-rewrite migrations and deserves its own design doc."*

## Proposed solution

Per-table migration using SQLite's supported table-rewrite pattern (same shape as `010_rate_limiter_per_model.sql`):

```sql
PRAGMA foreign_keys = OFF;  -- runner already does this

CREATE TABLE annotations_new (
    id INTEGER PRIMARY KEY AUTOINCREMENT,
    user_id INTEGER NOT NULL REFERENCES users(id) ON DELETE CASCADE,
    book_id INTEGER NOT NULL REFERENCES books(id) ON DELETE CASCADE,
    ... -- rest unchanged
);
INSERT INTO annotations_new SELECT * FROM annotations;
DROP TABLE annotations;
ALTER TABLE annotations_new RENAME TO annotations;
-- Recreate indexes
```

Orphan cleanup before the rewrite (per CLAUDE.md migration policy): delete rows whose parent is missing, otherwise the `INSERT SELECT` would fail once the new table has FK enforcement on.

After all tables have declared FKs:
[...]

PR #1061, opened 2026-04-24
feat: InsightChat history persistence — backend (#907 PR 1/2)

### LINKED ISSUE #907: bug: InsightChat message history only in localStorage — lost across browsers / devices / cache clear
## Reproduction

1. Open a book at `/reader/<id>`.
2. Open the AI Insight chat sidebar, ask a few questions, get answers.
3. Close the browser.
4. Open the reader in a **different browser** (or the same browser with cache cleared, or a private/incognito window).
5. Reopen the same book. The Insight chat panel is empty.

## What's happening

`frontend/src/components/InsightChat.tsx` stores the message thread in `localStorage`, keyed by `HISTORY_KEY(userId, bookId)`:

- Line 127 — `const [messages, setMessages] = useState<Message[]>([]);`
- Line 159 — read: `const raw = localStorage.getItem(HISTORY_KEY(userId ?? "anon", bookId));`
- Line 183 — write: `localStorage.setItem(HISTORY_KEY(userId ?? "anon", bookId), JSON.stringify(toStore));`

`SAVED_KEY` (line 113, 151, 446) is a separate localStorage entry for the set of "saved" insights (a Set<string> of insight hashes).

## What's NOT happening

The backend already has a `book_insights` table (migrations 015, 016, 022) that stores individual Q&A pairs per `user_id / book_id / chapter_index / question`. It looks built to persist insights, and there's even a UNIQUE index preventing duplicates per question. But the chat component doesn't sync its `messages` state to that table automatically — only explicitly "saved" insights (starred?) go to the backend.

So:
- Individual saved insights → backend `book_insights` table → persist across devices ✅
- The chat thread itself (the sequence of Q&A as it builds) → localStorage only → lost if cache clears or user switches browser ❌

## Options

**A. Sync all messages to backend on every exchange.** Add a new `chat_messages` table keyed on (user_id, book_id, chapter_index, position, role, content, created_at). Hydrate `messages` state on `InsightChat` mount from a new `GET /ai/chat-messages?book_id=…&chapter_index=…` endpoint. Persist via `POST` on each send/response. Full fidelity, multi-device.

**B. Reuse `book_insights` as the source of truth.** Every question the user asks already creates an insight row (if the backend `/ai/insight` route writes there). The chat UI could load its history from `book_insights` directly. Simpler than A — no new table. Trade-off: `book_insights` might be a dedupe-friendly schema whereas a chat thread wants ordered, repeat-preserving rows.

**C. Make ephemerality explicit.** If we decide chat threads are intentionally ephemeral (only "starred" insights persist), show a small tooltip/note in the chat UI explaining it, so users don't lose data silently. Lowest-effort; also lowest-fidelity.

## Priority

**P2** — user-visible data-loss from their perspective (they typed questions, they're gone), but no crash and users can work around it by staying in one browser.

## Path

Probably **Path B** — new persistence strategy touches schema (or a new table), crosses service boundaries (ai router + db services + frontend), and deserves an Architect design doc comparing options A / B / C.

## References

- Frontend: `frontend/src/components/InsightChat.tsx:127`, `:159`, `:183`
- Backend: `backend/migrations/015_reading_companion_fixes.sql` (book_insights table), `016_insight_context.sql`, `022_book_insights_unique.sql`

## Labels

`bug` / `architecture` (if Path B) / `P2`

alfmunny:
## Summary

Implements the backend half of #907 per the design in `docs/design/insightchat-history-persistence.md` (merged #1059). Server-side persistence for InsightChat messages so threads survive browser changes / cache clears / device switches.

## What this ships

- Migration 035 — new `chat_messages` table with declared FKs (CASCADE on users(id) and books(id)) per the #754 policy
- `backend/routers/chat.py` with three endpoints:
  - `GET /api/chat/{book_id}/messages` — newest-first, paginated via `before_id` (1 ≤ limit ≤ 200)
  - `POST /api/chat/{book_id}/messages` — append one message, 8 KB byte cap → 413
  - `DELETE /api/chat/{book_id}/messages` — clear the current (user, book) thread
- DB helpers: `append_chat_message`, `get_chat_messages`, `clear_chat_messages`, `CHAT_MESSAGE_MAX_BYTES`

All endpoints require auth and use the existing `check_book_access` helper. User isolation is structural (keyed on `user_id`).

## Not in this PR

- Frontend wiring — PR 2/2 per the design's rollout
- localStorage fallback migration — handled in PR 2/2
- Removing PR 2/2's localStorage fallback — scheduled follow-up a week after launch

## Test plan

**14 new tests in `test_router_chat.py`** — full suite: **1545 passed**.

Covers:
- DB helpers: insert + round-trip, newest-first + before_id pagination, user/book scoping, clear scoping
- FK cascade: delete user, delete book
- Router: POST+GET roundtrip, 401 for anon on all three verbs, 404 for unknown book, DELETE clears only the current thread, 413 for oversize body, 422 for invalid role, pagination `has_more`, user isolation

Closes #907 (the server contract half). Implementation PR 2/2 will close it once frontend lands.

### DIFF
--- a/backend/main.py
+++ b/backend/main.py
@@ -16,6 +16,7 @@ from routers.insights import router as insights_router
 from routers.uploads import router as uploads_router
 from routers.search import router as search_router
 from routers.decks import router as decks_router
+from routers.chat import router as chat_router
 from services.db import init_db
 
 
@@ -98,6 +99,7 @@ app.include_router(insights_router, prefix="/api")
 app.include_router(uploads_router, prefix="/api")
 app.include_router(search_router, prefix="/api")
 app.include_router(decks_router, prefix="/api")
+app.include_router(chat_router, prefix="/api")
 
 
 @app.get("/api/health")
--- a/backend/migrations/035_chat_messages.sql
+++ b/backend/migrations/035_chat_messages.sql
@@ -1,0 +1,21 @@
+-- Issue #907 / design doc: docs/design/insightchat-history-persistence.md
+-- Persist InsightChat conversation history on the backend so threads
+-- carry across browsers / devices / cache clears. See the design for
+-- the "why a separate table vs reusing book_insights" decision.
+--
+-- Declared FKs from day one per the #754 policy: ON DELETE CASCADE
+-- means delete_user and admin.delete_book need no shadow-cleanup.
+
+CREATE TABLE IF NOT EXISTS chat_messages (
+    id         INTEGER PRIMARY KEY AUTOINCREMENT,
+    user_id    INTEGER NOT NULL REFERENCES users(id) ON DELETE CASCADE,
+    book_id    INTEGER NOT NULL REFERENCES books(id) ON DELETE CASCADE,
+    role       TEXT    NOT NULL CHECK (role IN ('user', 'assistant')),
+    content    TEXT    NOT NULL,
+    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
+);
+
+-- Primary read pattern is "all messages for (user, book) ordered by time",
+-- reverse-paginated via `before_id`. This composite index covers it.
+CREATE INDEX IF NOT EXISTS idx_chat_messages_user_book
+    ON chat_messages (user_id, book_id, created_at);
--- a/backend/routers/chat.py
+++ b/backend/routers/chat.py
@@ -1,0 +1,97 @@
+"""InsightChat message-history endpoints (issue #907).
+
+Persists the running InsightChat conversation server-side so threads
+carry across browsers / devices / cache clears. See the design doc
+`docs/design/insightchat-history-persistence.md`.
+"""
+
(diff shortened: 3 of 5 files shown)


alfmunny:
**PM**: Implementation of the InsightChat history persistence design.

One process note: the parent design doc #1059 merged as a Path B bypass (17-second lifecycle, never carried `needs-user-approval`). The *direction* is user-approved at the issue level (#907), but the specific schema + endpoint shape was never given a user review window.

Not blocking this PR — the design content looks reasonable (chat_messages table with declared FKs matching the #754 policy, newest-first paginated GET, POST+DELETE on /api/chat/{book_id}/messages). But flagging for the user: if you want to skim `docs/design/insightchat-history-persistence.md` before this migration lands, now's the time. Otherwise it'll ship on the next CI pass.

Applied `P2` (label was missing).

PR #1485, opened 2026-04-26
fix: remove unreachable chapter_index < 0 guard in vocabulary save (closes #1484)

### LINKED ISSUE #1484: chore: remove unreachable chapter_index < 0 guard in vocabulary save endpoint
## Problem

In `routers/vocabulary.py` at line 62, the save-word handler has:

```python
if req.chapter_index < 0 or req.chapter_index >= len(_chapters):
```

The `< 0` branch is unreachable because the Pydantic model (`WordSave`) declares `chapter_index: int = Field(..., ge=0)`, which causes FastAPI/Pydantic to reject negative values with a 422 **before** the handler runs. The test `test_save_word_negative_chapter_index_returns_422` confirms the 422 behavior.

This is the same dead-code pattern that was removed from `routers/user.py` in a prior PR (#1477).

## Fix

Remove the `< 0` sub-check, leaving only:

```python
if req.chapter_index >= len(_chapters):
```

alfmunny:
## What changed

Removed the unreachable `req.chapter_index < 0` sub-check in `routers/vocabulary.py` `save` handler (line 62). The condition simplified from:

```python
if req.chapter_index < 0 or req.chapter_index >= len(_chapters):
```
to:
```python
if req.chapter_index >= len(_chapters):
```

## Why

`WordSave.chapter_index` is declared `Field(..., ge=0)`, so FastAPI/Pydantic rejects negative values with a 422 **before** the handler runs. `test_save_word_negative_chapter_index_returns_422` confirms this — the handler never sees a negative value.

This is the same dead-code pattern removed from `routers/user.py` in PR #1477.

## Test plan

- `pytest tests/test_router_vocabulary.py` — 50 passed (including the negative-index 422 test)
- Full backend suite: 1732 passed
- Frontend suite: 1750 passed

Closes #1484

### DIFF
--- a/backend/routers/vocabulary.py
+++ b/backend/routers/vocabulary.py
@@ -59,7 +59,7 @@ async def save(req: WordSave, user: dict = Depends(get_current_user)):
     check_book_access(book, user)
     from services.book_chapters import split_with_html_preference as _split
     _chapters = await _split(req.book_id, book.get("text") or "")
-    if req.chapter_index < 0 or req.chapter_index >= len(_chapters):
+    if req.chapter_index >= len(_chapters):
         raise HTTPException(
             status_code=400,
             detail=f"Chapter index out of range (book has {len(_chapters)} chapter(s)).",


alfmunny:
**Dev**: Superseded by #1487, which included this same fix as part of a comprehensive removal of unreachable chapter_index < 0 guards across all 6 routers. The change to vocabulary.py is already in main.

Issue #364, opened 2026-04-22
ux: mobile reader has no sub-sentence text selection / highlight path


alfmunny:
## Context

Documented in `docs/reader-interaction-design.md` §3 (Mobile section).

`SentenceReader` suppresses the native text-selection loupe via `e.preventDefault()` on touch `pointerdown` (PR #324, UX-003 fix). This correctly prevents the selection loupe from conflicting with the 400 ms long-press annotation gesture — but it also makes it impossible for mobile users to select and highlight a sub-sentence phrase.

## Current mobile annotation path

Long-press (400 ms) → `AnnotationToolbar` — annotates the **full sentence only**, not a sub-sentence selection.

## What's missing

Desktop users can drag-select any phrase, then Highlight → pick color → instant save. Mobile users can only annotate full sentences.

## Possible approach

After the long-press fires and the `AnnotationToolbar` opens, allow a custom handle-drag mode (similar to native iOS text handles) to adjust the selection boundaries before saving. This avoids fighting the browser's native loupe and gives full control over the selection.

Lower priority than core bugs, but affects reading/study use cases on mobile.

alfmunny:
**UI/UX**: Claimed — opening with a design-note PR (per CLAUDE.md UI/UX Dev rules: design note in `docs/design-improvement-plan.md` before coding for larger UX changes). Implementation will follow as a separate PR after the design note merges.

PR #1671, opened 2026-04-27
docs: design note for mobile sub-sentence selection (#364)

alfmunny:
## What

Adds a design note for **#364 — Mobile sub-sentence selection** to `docs/design-improvement-plan.md`, and updates the gap entry in `docs/reader-interaction-design.md` to point at it.

This is the **design phase** for #364 (which is marked `enhancement, ux`, no priority label, on the queue since 2026-04-22). Implementation will follow as a separate PR. Per CLAUDE.md UI/UX Dev rules: *"For larger UX changes: write a short design note in `docs/design-improvement-plan.md` before coding"*.

## Why

`SentenceReader` calls `e.preventDefault()` on touch `pointerdown` to suppress the browser's native selection loupe so the 500ms long-press cleanly opens the word-action drawer (PR #324, UX-003 fix). Side-effect: native text selection is also blocked, so a mobile user can only annotate the **full sentence** via long-press, never a sub-phrase.

The desktop `SelectionToolbar` (Highlight / Note / Chat / Vocab) already exists and works — the blocker is purely that touch input never produces a selection in the first place.

## Recommendation

Three approaches considered: custom iOS-style selection handles, drop touch `preventDefault` with motion-based gesture disambiguation, or an explicit "Select text" mode toggle button.

**Recommended: drop the touch `preventDefault`.** ~5 LOC change in `SentenceReader.tsx`, reuses the existing `SelectionToolbar` end-to-end, leans on the platform's familiar selection UI. The 500ms long-press timer + the existing pointermove `>10px` cancel already disambiguate cleanly:

- Hold still ≥500ms → timer fires → `removeAllRanges()` clears any nascent selection → word-action drawer opens (existing path).
- Drag >10px before 500ms → pointermove cancels timer → browser's native loupe + handles take over → user lifts → existing `SelectionToolbar` mounts above the selection rect.

Brief loupe-flash possible on pure long-press (browser starts the loupe ~300-450ms before our 500ms timer clears the selection); acceptable trade-off.

If post-ship feedback shows the loupe-flash is intrusive, we can iterate toward custom handles or a mode toggle; both are additive on top of this fix.

## Test plan

- [x] Doc-only PR. Full frontend suite still passes: 2544/2544.
- [x] Full backend suite still passes: 1382/1382.
- [ ] Implementation PR (follow-up): manual verification on iOS Safari + Android Chrome, plus a regex regression test asserting the touch `preventDefault` line is gone from `SentenceReader.tsx`.

References #364 (no auto-close — leaving the issue open for the implementation PR).

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -244,3 +244,52 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-25 | Native `confirm()` → optimistic delete + UndoToast in notes/library pages | notes/[bookId]/page.tsx, app/page.tsx | #1084 |
 | 2026-04-25 | Desktop touch-target scoping: `md:min-h-0` on reader header buttons (44px becomes mobile-only) | reader/[bookId]/page.tsx, CLAUDE.md | #1081 |
 | 2026-04-25 | Browse-books CTA on vocabulary + notes empty states | vocabulary/page.tsx, notes/page.tsx | #1093 |
+
+## #364 — Mobile sub-sentence selection (design note, 2026-04-27)
+
+**Status:** Design phase. Implementation deferred to a follow-up PR.
+
+**Problem.** Mobile users cannot highlight a sub-sentence phrase. `SentenceReader` calls `e.preventDefault()` on touch `pointerdown` (introduced in PR #324, UX-003 fix) to suppress the browser's native selection loupe so the 500ms long-press cleanly opens the word-action drawer. Side-effect: native text selection is also blocked, so a mobile user can only annotate the **full sentence** via long-press, never a sub-phrase. Documented as #UX-001 in `docs/reader-interaction-design.md`.
+
+The `SelectionToolbar` infrastructure (selection-change listener, Highlight / Note / Chat / Vocab actions) already exists and works on desktop. The blocker is purely that touch input never produces a selection in the first place.
+
+### Approaches considered
+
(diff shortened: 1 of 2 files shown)


alfmunny:
PM retroactive review (PR auto-merged before I had a chance to see it — flagging the workflow gap separately, not a blocker for this design):

✅ **Approved**. The recommendation is sound:

- The chosen approach (drop touch `preventDefault`, ~5 LOC) is genuinely the smallest reversible change. Reusing the existing `SelectionToolbar` is the right call — building custom handles would be 10× the work for a marginal experience gain.
- The disambiguation (500ms timer + `>10px` pointermove cancel) was already designed for word-action vs. drag-scroll, so leveraging it for word-action vs. text-select is principled, not opportunistic.
- The loupe-flash trade-off is correctly identified as the only real cost, with a sensible iteration path (custom handles / mode toggle) reserved for if it actually bothers users in practice.
- Falling back to platform-native selection UX is the right default on mobile — users already know how it works.

**Implementation follow-up:** Per the test plan, the regression test asserting the `preventDefault` call is gone from `SentenceReader.tsx` is the load-bearing guard. Without it, a future "fix the loupe-flash" PR could silently re-add `preventDefault` and re-create #364.

**Workflow note (separate from design):** Path B in CLAUDE.md says PM reviews the design doc *before* it's merged. This PR auto-merged at 08:35 UTC without a PM review pass. It happens to be doc-only and the design is fine, so no harm here, but in general the design doc should be held until PM approves. Will add this to the next CLAUDE.md cycle.

PR #1705, opened 2026-04-27
fix: save_book UPDATE-or-INSERT to prevent ON DELETE CASCADE wipe (closes #1703)

### LINKED ISSUE #1703: BUG (P0): clicking 'import' on an already-imported book wipes all its translations (FK CASCADE on INSERT OR REPLACE)
## What happened

Pushed 46 chapters of Tale of Two Cities (#98) zh translation to prod via \`seed_translations.py\` over many cycles. User then clicked **import** on the reader page (https://book-reader-ai.vercel.app/reader/98). After that the zh translation was gone — all 46 chapter rows in \`translations\` were wiped.

Confirmed by re-running \`seed_translations.py --file backend/data/translations/98_zh.json\` against prod: imported 46 rows successfully and the reader works again. So the rows were definitely deleted, not corrupted.

## Root cause

\`backend/services/db.py:save_book()\` line 492:
\`\`\`python
await db.execute(\"\"\"INSERT OR REPLACE INTO books ...\"\"\", (...))
\`\`\`

SQLite \`INSERT OR REPLACE\` semantics: when a row with the conflicting PK already exists, it **deletes** the existing row, then inserts the new one. The DELETE step fires \`ON DELETE CASCADE\` on every dependent FK.

\`translations\` schema:
\`\`\`sql
CREATE TABLE translations (
    book_id INTEGER NOT NULL REFERENCES books(id) ON DELETE CASCADE,
    ...
)
\`\`\`

With \`PRAGMA foreign_keys = ON\` (shipped in #700/#748), every save_book() call for an existing book wipes:
- \`translations\` (entire book's translations across all languages)
- \`audio_cache\` (likely)
- \`annotations\` (likely)
- any other table with \`book_id ... ON DELETE CASCADE\`

The trigger path is \`POST /books/{id}/import-stream\` → \`_fetch_and_cache\` → \`save_book\`. The stream's docstring says \"Idempotent: skips already-cached work\", but \`save_book\` itself is unconditionally destructive.

## Why it didn't show before FK enforcement

Pre-#700/#748 the \`PRAGMA foreign_keys\` was OFF in prod (it's still OFF in my dev DB), so the cascade was a no-op. After enabling FK enforcement, the existing \`save_book\` path became data-destructive, and nobody noticed because the testing path doesn't go through 'import an already-imported book'.

## Fix proposal

\`save_book\` already has an early-out for uploaded books (line 489). Add a parallel guard for the gutenberg case, e.g.:

\`\`\`python
async with db.execute(\"SELECT 1 FROM books WHERE id=?\", (book_id,)) as _cur:
    _exists = await _cur.fetchone() is not None
if _exists:
    # UPDATE the columns we know we want to refresh, instead of REPLACE
    await db.execute(
        \"\"\"UPDATE books SET title=?, authors=?, languages=?, subjects=?,
                              download_count=?, cover=?, text=?, images=?
            WHERE id=?\"\"\",
        (...)
    )
else:
    await db.execute(\"INSERT INTO books (...) VALUES (...)\", (...))
\`\`\`

\`UPDATE\` does not trigger ON DELETE CASCADE on dependent FKs, so translations / audio / annotations survive.

Alternatively wrap the path in a \`PRAGMA defer_foreign_keys = ON\` block, but UPDATE-vs-INSERT is the cleaner fix and matches the user's mental model: re-importing should refresh metadata + text, not wipe their translation work.

## Test plan

- Regression test: import a book, write a row to \`translations\`, call \`save_book()\` again, assert the translation row is still there. Should fail today, pass after fix.
- Same for audio_cache, annotations.

## Adjacent damage

The same destructive pattern affects every book in the catalog the moment a user clicks import on it. Backup-restore via \`seed_translations.py\` is the only mitigation right now. Per #1393 the JSON backups under \`backend/data/translations/\` are the durable record (committed in PR #1702); for any book NOT in the marathon, prod is the only copy and re-import is unrecoverable.

## Mitigation already done

Re-seeded 8 books to prod from the JSON backups (Tale of Two Cities #98, Faust #2229, Alice #11, Romeo #1513, Wuthering Heights #768, Blue Castle #67979, Gatsby #64317, A Room with a View #2641). Four books in legacy dict format (#43, #174, #1661, #84) need a separate converter to re-seed; will file as a follow-up if needed.

alfmunny:
## What changed

`services/db.py` — `save_book()` now uses `UPDATE` for existing rows instead of `INSERT OR REPLACE`.

## Why

`INSERT OR REPLACE` silently deletes the existing row before re-inserting, which fires `ON DELETE CASCADE` on every dependent FK. With `PRAGMA foreign_keys = ON` (shipped in #700/#748), every call to `save_book()` on an already-cached book was wiping:
- `translations` (all chapters, all languages)
- `audio_cache`
- `annotations`
- any other table with `book_id ... ON DELETE CASCADE`

The trigger path is the reader "import" button → `POST /books/{id}/import-stream` → `save_book()`. Confirmed in prod: Tale of Two Cities #98 lost all 46 zh translations this way. Closes #1703.

## Fix

- `_existing` guard already existed for `source='upload'` books (#467).
- Extended: if ANY row exists, `UPDATE` the metadata/text columns instead of `REPLACE`. Only `INSERT` for genuinely new books.
- `UPDATE` does not trigger `ON DELETE CASCADE`.

## Test plan

Three regression tests added to `tests/test_db.py`:
- `test_save_book_twice_preserves_translations` — translation row survives re-import
- `test_save_book_twice_preserves_audio_cache` — audio_cache row survives re-import
- `test_save_book_updates_metadata_on_reimport` — title/text are still refreshed on re-import

Full backend suite: 1909 passed.

### DIFF
--- a/backend/services/db.py
+++ b/backend/services/db.py
@@ -512,7 +512,7 @@ async def get_cached_book(book_id: int) -> dict | None:
 
 
 async def save_book(book_id: int, meta: dict, text: str, images: list | None = None) -> None:
-    """Insert or replace a book record (meta + full text + images).
+    """Insert or update a book record (meta + full text + images).
 
     After saving, the translation queue is auto-seeded with this book's
     chapters for every configured target language. The worker (if running)
@@ -521,8 +521,7 @@ async def save_book(book_id: int, meta: dict, text: str, images: list | None = N
     async with aiosqlite.connect(DB_PATH) as db:
         # Guard against clobbering a user's uploaded private book that happens
         # to share the same auto-assigned SQLite ID as a Gutenberg catalog ID.
-        # INSERT OR REPLACE would delete the uploaded row (losing source='upload'
-        # and owner_user_id) and re-insert without those fields. (#467)
+        # (#467)
         async with db.execute(
             "SELECT source FROM books WHERE id=?", (book_id,)
         ) as _cur:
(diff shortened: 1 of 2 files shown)


alfmunny:
PM ack — this is the actual data-loss vector. Important context for downstream:

1. **This is independent of #1561** (EPUB-fetch chapter-cache invalidation). #1561 fixed the *forward* path on background EPUB fetch. **#1703/#1705 fixes a separate, much broader path**: every click of the reader's "import" button on an already-cached book was silently wiping translations + audio_cache + annotations via `INSERT OR REPLACE` → `ON DELETE CASCADE`. The user-reported book 2701 (Moby Dick) mismatch may have been compounded by repeated import clicks, not just the original plain-text→EPUB index drift.

2. **Audit follow-up worth filing:** any other place in `services/db.py` that uses `INSERT OR REPLACE` on a table with FK-CASCADE dependents has the same bug. `save_book` is one; if there are others (settings? user records? deck records with `cards ON DELETE CASCADE`?) they need the same UPDATE-or-INSERT treatment. Worth a 30-min grep + audit issue rather than discovering each case individually via more data-loss reports.

3. The three regression tests in this PR cover the trigger path well; the `_updates_metadata_on_reimport` test is the load-bearing one — it confirms the refresh-content behavior we still want is preserved.

Applied **P0** label (matches issue). Auto-merge is already armed, so this should fire as soon as required CI is green. No blockers from PM.

PR #1746, opened 2026-04-27
fix: use DB chapter text for summaries, not client-supplied text (closes #1738)

### LINKED ISSUE #1738: bug: POST /ai/summary uses client-supplied chapter_text, poisoning shared summary cache
## Problem

`POST /ai/summary` accepts `chapter_text` from the client and uses it verbatim to generate a Gemini summary. The generated summary is stored in the **shared** `chapter_summaries` table and served to **all users** for that (book_id, chapter_index) pair.

Any authenticated user can craft a request with arbitrary `chapter_text` that does not match the actual chapter content:
```json
{
  "book_id": 1,
  "chapter_index": 0,
  "chapter_text": "<attacker-controlled content>",
  "book_title": "Moby Dick",
  "author": "Melville"
}
```

The server:
1. Verifies the book exists + chapter_index is in-bounds ✅
2. Uses the **client-supplied** `chapter_text` for the Gemini call ❌
3. Caches the misleading summary globally (keyed by book_id + chapter_index) ❌

All subsequent users receive the poisoned summary until an admin deletes it.

## Root Cause

`ai.py` line ~282:
```python
content = await gemini.generate_chapter_summary(
    api_key, req.chapter_text, ...   # ← client-controlled input
)
await save_chapter_summary(req.book_id, req.chapter_index, content, ...)
```

By this point, the endpoint has already computed `_chapters = await _split(req.book_id, book.get("text") or "")` for the bounds check. `_chapters[req.chapter_index].text` holds the actual chapter text.

## Fix

Use `_chapters[req.chapter_index].text` (fetched from DB) instead of `req.chapter_text` when calling Gemini. Keep `chapter_text` in the request model to avoid a breaking API change, but ignore it server-side.

## Regression test

POST /ai/summary with a mismatching chapter_text; verify the returned (and cached) summary is based on the real chapter text, not the injected one.

alfmunny:
## What changed

`POST /ai/summary` was forwarding the client-supplied `req.chapter_text` directly to Gemini and caching the result under the book/chapter key. Any authenticated user could submit arbitrary text to permanently poison the shared summary cache for other users.

Fix: the endpoint now ignores `req.chapter_text` and instead uses `_chapters[req.chapter_index].text`, which is the chapter text fetched from the database.

## Why

Cache entries are keyed by `(book_id, chapter_index)` and shared across all users. Letting the client control the text that gets cached is a privilege-escalation / content-poisoning vector.

## Test plan

- Regression test `test_summary_uses_db_chapter_text_not_client_supplied` (new): sends `chapter_text = "INJECTED ATTACKER CONTENT"` in the request, captures what text is passed to `gemini.generate_chapter_summary`, and asserts the attacker string is never forwarded.
- Full backend suite: 1925 passed, 0 failures.

Closes #1738

### DIFF
--- a/backend/routers/ai.py
+++ b/backend/routers/ai.py
@@ -273,7 +273,7 @@ async def summary(req: SummaryRequest, _user: dict = Depends(get_current_user)):
 
         try:
             content = await gemini.generate_chapter_summary(
-                api_key, req.chapter_text, req.book_title, req.author, req.chapter_title
+                api_key, _chapters[req.chapter_index].text, req.book_title, req.author, req.chapter_title
             )
         except Exception as exc:
             logger.exception("POST /ai/summary book=%s ch=%s: %s", req.book_id, req.chapter_index, exc)
--- a/backend/tests/test_router_ai.py
+++ b/backend/tests/test_router_ai.py
@@ -1796,3 +1796,50 @@ async def test_summary_no_queue_api_key_returns_503(client, test_user, tmp_db):
         f"got {resp.status_code}: {resp.text}"
     )
     assert "not available" in resp.json()["detail"].lower() or "configured" in resp.json()["detail"].lower()
+
+
+# ── Issue #1738: summary cache poison via client-supplied chapter_text ─────────
+
+
+async def test_summary_uses_db_chapter_text_not_client_supplied(client, test_user, tmp_db):
+    """Regression #1738: POST /ai/summary must use the chapter text from the DB,
+    not req.chapter_text. Any authenticated user can supply arbitrary chapter_text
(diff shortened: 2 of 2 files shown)


alfmunny:
**PM review.**

Tight, focused fix: 1-line change in \`backend/routers/ai.py\` (uses DB chapter text instead of \`req.chapter_text\`) + 47-line regression test that verifies the attacker string never reaches \`gemini.generate_chapter_summary\`. Closes the cache-poison vector in #1738 cleanly.

**Housekeeping:** PR was filed with only \`bug\` — applied \`P2\` per label policy (matches issue priority).

Vercel preview rate-limit failure is external quota, not a code issue. No blocking concerns. Auto-merge should fire once tests finish.

PR #1755, opened 2026-04-27
ux: fix InsightChat chapter-divider, save-button, and keyboard-hint contrast

alfmunny:
## What changed

`InsightChat.tsx`: fixed three `text-gray-400` usages that fail WCAG AA contrast on white/light backgrounds:

| Location | Before | After |
|---|---|---|
| Chapter-divider label (line 462) | `text-gray-400` | `text-gray-600` |
| "Save to notes" active state (line 542) | `text-gray-400` | `text-gray-600` |
| "Save to notes" saved/disabled state (line 541) | `text-gray-300` | `text-gray-500` |
| Keyboard hint "Enter to send…" (line 614) | `text-gray-400` | `text-gray-600` |

## Why

`text-gray-400` (`#9ca3af`) on white gives ~2.8:1 contrast — below the WCAG AA minimum of 4.5:1 for `text-xs`/`text-[11px]` small text. All three locations show readable text content that users need to see. `text-gray-600` on white gives ~7:1.

## Test plan

- New `InsightChatSecondaryContrast.test.ts`: three assertions verifying none of the failing patterns remain.
- All 387 frontend test suites pass (2568 tests).

Closes #1754

PR #1752, opened 2026-04-27
ux: fix InsightChat load-earlier button contrast (text-gray-400 → text-gray-600)

alfmunny:
## What changed

`InsightChat.tsx`: updated the "Load earlier" button colour classes:
- `text-gray-400` → `text-gray-600` (resting text)
- `hover:text-gray-600` → `hover:text-gray-800` (hover text)
- `border-gray-100` → `border-gray-200` (border, near-invisible on white before)

## Why

`text-gray-400` (`#9ca3af`) on `bg-white` produces ~2.8:1 contrast — below the WCAG AA minimum of 4.5:1 for `text-xs` (small text). `text-gray-600` on white gives ~5.7:1, comfortably passing.

The border fix makes the button boundary visible even at rest (gray-100 on white is essentially invisible).

## Test plan

- New `InsightChatLoadEarlierContrast.test.ts`: asserts that `InsightChat.tsx` no longer contains the `text-gray-400 hover:text-gray-600` pattern.
- All 386 frontend test suites pass (2565 tests).

Closes #1751

PR #1750, opened 2026-04-27
ux: fix VocabWordTooltip saved-state contrast (text-stone-400 → text-stone-600)

alfmunny:
## What changed

`VocabWordTooltip.tsx`: changed the disabled "Saved" button class from `text-stone-400` to `text-stone-600`.

## Why

`text-stone-400` on `bg-stone-100` produces ~2.3:1 contrast ratio — well below the WCAG AA requirement of 4.5:1 for normal text. The "Saved" label is meaningful feedback visible to users who have just saved a vocabulary word; it must be legible.

`text-stone-600` on `bg-stone-100` gives ~6.6:1, comfortably passing.

## Test plan

- New `ContrastWave14.test.ts`: asserts `VocabWordTooltip.tsx` does not use `text-stone-400` in the saved-button class.
- All 386 frontend test suites pass (2565 tests).

Closes #1749

PR #1760, opened 2026-04-27
ux: fix SentenceReader unloaded-disabled segment contrast (text-stone-400/70 → text-stone-500)

### LINKED ISSUE #1758: ux: SentenceReader unloaded-disabled segment uses text-stone-400/70 (contrast failure)
## Problem

In `SentenceReader.tsx:670`, text segments in the disabled+unloaded state render as `text-stone-400/70` (stone-400 at 70% opacity). This gives ~2.7:1 contrast on white — well below WCAG AA 4.5:1. The text is still visible book content that users may try to read.

```tsx
if (disabled) {
  return loaded
    ? "text-stone-500"          // ~4.7:1 — passes
    : "text-stone-400/70";      // ~2.7:1 — fails
}
```

The loaded disabled state already correctly uses `text-stone-500`. The unloaded branch should match.

## Fix

Change `text-stone-400/70` → `text-stone-500` to be consistent with the loaded-disabled state and pass WCAG AA.

## Test

Add a contrast assertion in a new test file verifying `SentenceReader.tsx` does not use `text-stone-400/70`.

alfmunny:
## What changed

`SentenceReader.tsx`: changed the unloaded disabled-segment colour from `text-stone-400/70` to `text-stone-500`.

## Why

`text-stone-400` at 70% opacity on white gives ~2.7:1 contrast — well below the WCAG AA minimum of 4.5:1. The text is visible book content users may try to read during TTS loading. `text-stone-500` (~4.7:1) is consistent with the already-correct loaded-disabled state and passes WCAG AA.

Before:
```tsx
if (disabled) {
  return loaded ? "text-stone-500" : "text-stone-400/70";  // unloaded fails
}
```
After:
```tsx
if (disabled) {
  return loaded ? "text-stone-500" : "text-stone-500";  // consistent
}
```

## Test plan

- New `SentenceReaderDisabledContrast.test.ts`: asserts `SentenceReader.tsx` does not contain `text-stone-400/70`.
- All 387 frontend test suites pass (2566 tests).

Closes #1758

### DIFF
--- a/frontend/src/__tests__/SentenceReaderDisabledContrast.test.ts
+++ b/frontend/src/__tests__/SentenceReaderDisabledContrast.test.ts
@@ -1,0 +1,19 @@
+/**
+ * Contrast regression for issue #1758:
+ * SentenceReader unloaded-disabled segment must not use text-stone-400/70.
+ * text-stone-400 at 70% opacity on white gives ~2.7:1 — fails WCAG AA 4.5:1.
+ * The loaded-disabled state correctly uses text-stone-500 (~4.7:1); unloaded must match.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../components/SentenceReader.tsx"),
+  "utf8",
+);
+
+describe("SentenceReader disabled-segment contrast (closes #1758)", () => {
+  it("does not use text-stone-400/70 for unloaded disabled segments", () => {
+    expect(src).not.toMatch(/text-stone-400\/70/);
+  });
+});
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.**

Fourth contrast fix in the UI/UX audit sweep (after #1750 merged, #1752 / #1755 pending). `text-stone-400/70` (2.7:1 on white) → `text-stone-500` (4.7:1) on the unloaded-disabled segment in SentenceReader. Also normalises loaded/unloaded disabled states to the same colour — small consistency win. 1-line code change + 19-line static-assertion test. Closes #1758.

Applied `P2` per label policy. PM approved.

PR #1741, opened 2026-04-27
design: splitter-version cache-invalidation column + resplit_book.py CLI (#1221)

alfmunny:
## Summary

Path B design doc for issue #1221 — splitter-version cache invalidation + `resplit_book.py` CLI. Required prerequisite before re-splitting any book in the catalog (called out in #964 as a follow-up).

- Adds `books.splitter_version INTEGER NOT NULL DEFAULT 0` (legacy=0, current=1 post-#1055).
- Documents per-table cache-invalidation policy on resplit (translations, audio_cache, chapter_summaries, book_insights chapter-pinned, translation_queue, annotations, word_occurrences → DELETE per book; user_reading_progress → CLAMP; chat_messages / reading_history / book-level insights → KEEP).
- New `backend/scripts/resplit_book.py <id>` with `--dry-run` (default), `--confirm`, `--bump-version-only`. Diff-then-confirm semantics, single-transaction cascade.
- No automatic resplit on read — operator-in-the-loop.
- Implementation lands as a follow-up PR (or via conversion of #1221) once the design is merged.

Closes #1221.

## Path B gate

This PR will land with `needs-user-approval` per CLAUDE.md gate layer 2. PM reviews for readiness and applies `pm-approved`; user is the sole approver and removes `needs-user-approval` + applies `user-approved` to release the merge.

## Test plan

- [ ] PM reviews the design doc against §"Acceptance" and §"Risks / open questions"
- [ ] User reviews the per-table policy table (especially the `annotations` DELETE — user-visible cost)
- [ ] User reviews the `resplit_book.py` UX (dry-run output shape, `--bump-version-only` escape hatch)
- [ ] User confirms versioning scheme (monotonic int, not git SHA) is the right call

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1761, opened 2026-04-27
fix: reconcile subtitle paragraph shift in chapter translation responses (closes #1385)

### LINKED ISSUE #1385: bug: Moby Dick chapter translations misaligned — subtitle paragraph in body causes paragraph-index shift
## Problem

For books like Moby Dick, chapter translations are misaligned with the original text starting from Chapter 32 (Cetology). The translated column has an **extra first paragraph** (the subtitle translation, e.g. "Cetología.") that the original text column does not show, causing every subsequent paragraph to be off by one.

## Root cause

The translations were cached when the book only had **plain text** (Gutenberg text file). The plain-text splitter (`_strip_heading_from_text` in `services/splitter.py`) strips the chapter number line ("CHAPTER XXXII.") but leaves the subtitle ("Cetology.") as the first paragraph of the body text. Gemini then translates it as `paragraphs[0]`.

Later, when an EPUB becomes available, `split_with_html_preference` switches to the EPUB splitter, which strips the subtitle from the body (via `_strip_leading_headings`). The source text now has N paragraphs but the stored translation still has N+1 (with the subtitle as `paragraphs[0]`).

## Evidence

- Chapters 1–31: single-line headings ("CHAPTER I. Loomings.") → stripped correctly by both splitters → aligned
- Chapters 32+: two-line headings ("CHAPTER XXXII." + blank + "Cetology.") → plain-text splitter leaves subtitle in body → misaligned when EPUB takes over
- `title_translation` is inconsistently set (manual admin edit for some chapters), but body mismatch is consistent

## Fix

`_reconcile_subtitle` in `routers/books.py`: when serving a GET or POST (cached) translation, detect `len(translated) == len(source) + 1` with a short single-line `paragraphs[0]` and shift it into `title_translation`. This fixes alignment without requiring a migration or re-translation.

Closes this issue via PR (regression tests included).

alfmunny:
## What changed

Added `_reconcile_subtitle()` helper in `routers/books.py` and applied it at both the GET and POST (cache-hit) translation response paths.

When `len(translated_paragraphs) == len(source_paragraphs) + 1` **and** the first translated paragraph is a short single-line string (≤ 120 chars, no newlines), the extra leading paragraph is shifted into `title_translation` instead of being returned in the `paragraphs` array.

## Why

Books like Moby Dick were translated before their EPUB became available. The plain-text splitter left the chapter subtitle (e.g. "Cetology.") as the first body paragraph. Gemini translated it as `paragraphs[0]`. Later, when the EPUB was indexed, `split_with_html_preference` switched to the EPUB splitter which strips the subtitle from the body — leaving the source at N paragraphs but the translation at N+1. Every subsequent paragraph was off by one from Chapter 32 onwards.

The fix serves correctly aligned translations without requiring a DB migration or re-translation. `title_translation` is already the designated field for translated subtitle/chapter-title text.

## Test plan

- `test_get_translation_reconciles_subtitle_shift` — GET endpoint shifts the extra leading paragraph to `title_translation`
- `test_get_translation_no_reconcile_when_title_already_set` — existing `title_translation` is not overwritten
- `test_get_translation_no_reconcile_when_lengths_match` — no shift when counts already match
- `test_post_translation_cache_hit_reconciles_subtitle_shift` — POST cache-hit path also reconciles
- Full suite: 1928 passed, 0 failures

Closes #1385

### DIFF
--- a/backend/routers/books.py
+++ b/backend/routers/books.py
@@ -182,6 +182,29 @@ async def chapter_queue_status(
     return await queue_status_for_chapter(book_id, chapter_index, target_language)
 
 
+def _reconcile_subtitle(
+    paragraphs: list[str], source_text: str, title_translation: str | None
+) -> tuple[list[str], str | None]:
+    """Shift an extra leading translated paragraph into title_translation.
+
+    When a book was translated from plain text before its EPUB became available,
+    the plain-text splitter may have left the chapter subtitle as the first body
+    paragraph. Later the EPUB splitter strips it, leaving the translation one
+    paragraph longer than the source. Shifts the surplus leading paragraph into
+    title_translation so alignment is restored without re-translating.
+    """
+    if title_translation is not None:
+        return paragraphs, title_translation
+    source_paragraphs = [p.strip() for p in source_text.split("\n\n") if p.strip()]
+    if len(paragraphs) != len(source_paragraphs) + 1:
+        return paragraphs, title_translation
+    first = paragraphs[0]
+    # Accept as a subtitle only if it's a single short line (no embedded newlines).
+    if "\n" in first or len(first) > 120:
+        return paragraphs, title_translation
+    return paragraphs[1:], first
+
+
 @router.get("/{book_id}/chapters/{chapter_index}/translation")
 async def get_chapter_translation(
     book_id: int = Path(..., ge=1),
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.**

Closes the P1 Moby Dick misalignment bug (#1385) without DB migration or re-translation. Heuristic is conservative — fires only when (translation paragraph count == source + 1) AND (extra leading paragraph ≤ 120 chars, no newlines). Disposition shifts the extra to `title_translation`, the designated field. 4 focused router tests cover the happy path, idempotency on already-set title, no-shift on matching lengths, and the POST cache-hit path. Approved on content.

Applied `P1` per label policy (PR was filed with only `bug`; #1385 was P1).

**Tradeoff to flag (non-blocking):** This is a response-layer band-aid, not a root-cause fix. The persisted rows in `translations` remain misaligned by one — every GET request now does reconciliation forever. Failure modes:
1. **Subtitle (dominant case)**: heuristic fires correctly, content rendered in title slot. ✅
2. **Genuine short first paragraph**: heuristic mis-fires, paragraph gets rendered as title rather than first body paragraph. Unusual but possible.
3. **Different misalignment cause**: heuristic dispositions data into title rather than corrupting body. Acceptable failure mode.

A follow-up issue would be the proper fix: either (a) re-translate the affected chapters with the EPUB-aligned source (costs API quota; one-shot), or (b) migrate the persisted rows in-place (DB-side, no API cost). The new `splitter_version` + `resplit_book.py` design from #1741 is precisely the machinery for this — once that lands, the affected books resplit + re-translate the misaligned chapters. Worth filing as a follow-up so we don't carry the reconciliation indefinitely.

I'll file the follow-up after this merges.

Issue #1748, opened 2026-04-27
ux: surface bulk pre-translate entry point now that the import-page panel is gone


alfmunny:
## What

PR #1745 (closes #1739) removed the post-import \"Pre-translate this book?\" panel from \`import/[bookId]/page.tsx\` and moved cost-estimate display per-chapter into the reader sidebar. That's the right call for the read-as-you-go flow — the cost is now shown at the moment of action.

But the import-page panel was also the only entry point for **bulk pre-translate** (\"translate every chapter overnight, then I'll read tomorrow\"). With the panel gone, a user who specifically wants the bulk flow has nowhere to trigger it from the UI.

## Why this matters

The bulk flow is a real use case for users with a Gemini key configured: imagine importing War and Peace and wanting all 365 chapters translated in advance so reading is friction-free. Today, post-#1745, there is no UI affordance for that.

The bulk-translate **endpoints** still exist (queue API). Only the UI entry point is missing.

## Proposed fix

Surface bulk pre-translate from the reader sidebar in one of these forms (UI/UX session to choose):

1. **Overflow menu item** in the reader sidebar / typography panel: "Pre-translate all remaining chapters (~\$X)".
2. **Inline banner** the first time a user opens a book that has a Gemini key configured AND ≥N untranslated chapters: \"You have N chapters untranslated. Pre-translate all? (~\$X)\". Dismissible.
3. **Settings page** entry under per-book actions.

Cost estimate continues to use the same Gemini cost-per-chapter math from #1745, summed over all untranslated chapters.

## Constraint

Whatever form is chosen, the entry point must be **gated on \`hasGeminiKey === true\`** (same as the per-chapter estimate from #1745). Users without a key never see paid pricing.

## Priority

P3 — not a regression of pre-existing functionality (the queue endpoints still work; admins can trigger it). It's a discoverability gap that affects one specific power-user flow.

alfmunny:
**PM**: superseded by PR #1763, which moves the per-chapter cost estimate from #1745 to the "Translate remaining N" queue button and converts it to the **total** cost of all remaining chapters. Reviewing the reader sidebar more carefully: the bulk-translate entry point was never lost — the "Translate remaining N" button has been present all along. My filing of this issue was based on a misread of #1745's diff. With #1763 the bulk flow is both visible and commitment-priced — exactly the spirit of what was asked here. Closing as effectively complete.

PR #1745, opened 2026-04-27
ux: move per-chapter Gemini cost estimate from import page to reader sidebar

alfmunny:
## What changed

- **Import page** (`import/[bookId]/page.tsx`): Removed the post-import "Pre-translate this book?" panel entirely. The import flow is now lean: fetch → split → read. Auto-redirects to the reader as soon as import is done.
- **Reader sidebar** (`reader/[bookId]/page.tsx`): Added a per-chapter Gemini cost estimate (`~$0.01 · ~2.8K tokens`) directly below the "Translate this chapter" button, shown only when the user has a Gemini API key configured (`hasGeminiKey === true`). Users without a key never see paid pricing.

## Why

Cost estimate is most relevant at the moment of triggering a paid translation, not at import time. Moving it eliminates a friction-laden confirmation panel from the import flow and places the cost information where it's actionable (in the reader sidebar, right before the user clicks translate). Closes #1739.

## Test plan

- [x] `ImportPageNoTranslatePrompt.test.tsx` — verifies translate prompt no longer appears after done event (even with source ≠ target language)
- [x] `ReaderGeminiCostEstimate.test.tsx` — verifies cost estimate appears below translate button when `hasGeminiKey=true`, and is absent when `hasGeminiKey=false`
- [x] Updated `ImportPage.branches.test.tsx` — removed tests for removed UI; added new tests for simplified `canStartReading` behavior
- [x] Updated `ContrastWave8.test.ts`, `ImportChaptersRoleAlert.test.ts`, `ImportUploadChaptersTouchTargets.test.ts` — removed assertions for removed UI elements
- [x] All 2564 frontend tests pass; all 1382 backend tests pass

## Docs follow-up

N/A — no user-facing feature page change; the cost estimate is a minor UX improvement.

PR #1763, opened 2026-04-27
feat: move cost estimate from single-chapter button to queue-all button

alfmunny:
## What changed

- Renamed `chapterCostLabel` → `queueTotalCostLabel(text, chapterCount)` — now computes the **total** estimated cost for all remaining chapters, not just one.
- Removed the cost estimate from under the single-chapter "Translate this chapter" button.
- Added the estimate below the "Translate remaining N" queue button (only shown when `hasGeminiKey === true` and `notStarted > 0`).
- Updated `ReaderGeminiCostEstimate.test.tsx`: tests now verify the estimate appears near the queue button with `translationStatus` set, and confirm it no longer appears under "Translate this chapter".

## Why

When a user opens a freshly imported book, the reader shows a "Translate this chapter" button with a cost hint. This was confusing: there is no background enqueue happening during import — the user explicitly queues translation. The per-chapter cost estimate belonged next to the action that actually commits to queuing (the "Translate remaining N" button), where showing the **total** cost for all remaining chapters is actionable and contextually accurate.

## Test plan

- `shows cost estimate near the queue button when hasGeminiKey is true and chapters remain` — estimate appears near "Translate remaining N"
- `does not show cost estimate under the single-chapter translate button` — no estimate when `bookTranslationStatus` is null
- `does not show cost estimate when hasGeminiKey is false` — gated by key presence
- Full frontend suite: 2567 passed, 0 failures

### DIFF
--- a/frontend/src/__tests__/ReaderGeminiCostEstimate.test.tsx
+++ b/frontend/src/__tests__/ReaderGeminiCostEstimate.test.tsx
@@ -1,6 +1,8 @@
 /**
- * Tests for issue #1739: per-chapter Gemini cost estimate in the reader
- * translation sidebar, shown only when hasGeminiKey === true.
+ * Tests for translation cost estimate in the reader sidebar.
+ * Originally issue #1739 (per-chapter estimate under "Translate this chapter").
+ * Updated: cost moved to the "Translate remaining N" queue button area,
+ * showing total queue cost estimate (notStarted × avg chapter cost).
  */
 import React from "react";
 import { render, screen, waitFor, fireEvent } from "@testing-library/react";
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.**

Cleanly relocates the cost estimate from the single-chapter button (where it was added in #1745) to the "Translate remaining N" queue button — and switches it from per-chapter cost to the **total** cost of all remaining chapters. Right call: cost belongs where the commitment happens, and a single-chapter cost is uninteresting in isolation. Tests updated accordingly.

Applied `P2` per label policy.

**Closes part of #1748.** When I filed #1748 after #1745 merged, I was wrong about the bulk-translate entry point being lost — "Translate remaining N" already exists in the reader sidebar. So #1748's discoverability concern was misdiagnosed. With this PR (showing total cost next to the bulk-action button) the bulk flow is now both **visible** and **commitment-priced**. Closing #1748 as effectively addressed.

Issue #1772, opened 2026-04-27
bug: Gemini MODEL constant pinned to 'gemini-3.1-flash-lite-preview' — preview model name suspected stale, all AI calls failing today (insight + summary)


alfmunny:
## Symptom

Admin reports \"AI service request failed\" 500 from both \`POST /ai/insight\` and \`POST /ai/summary\` — every AI call is failing in production today (2026-04-27).

The actual error from Gemini is masked by #<companion-issue> (router exception-swallowing). Once that lands, the error message will surface; until then the diagnosis below is suspected, not confirmed.

## Suspected root cause

\`backend/services/gemini.py:10\` pins the model to:

\`\`\`py
MODEL = \"gemini-3.1-flash-lite-preview\"
\`\`\`

This file hasn't been touched since PR #262 (session 3, weeks ago — the AI chapter summary feature). The \`-preview\` suffix means Google can deprecate the model name without notice; preview models also typically have tighter free-tier quotas than GA models.

If Google has sunset \`gemini-3.1-flash-lite-preview\`, every call returns 404 from Generative Language API, and the route's catch-all surfaces it as \"AI service request failed\".

## How to verify (one-shot)

\`\`\`bash
# Replace $KEY with the queue Gemini key. Compare 200 (working) vs 4xx (broken).
curl -s -o /dev/null -w \"%{http_code}\\\\n\" -X POST \\
  \"https://generativelanguage.googleapis.com/v1beta/models/gemini-3.1-flash-lite-preview:generateContent?key=\$KEY\" \\
  -H \"content-type: application/json\" \\
  -d '{\"contents\":[{\"parts\":[{\"text\":\"hello\"}]}]}'

# Same key against a known-stable Flash model.
curl -s -o /dev/null -w \"%{http_code}\\\\n\" -X POST \\
  \"https://generativelanguage.googleapis.com/v1beta/models/gemini-2.5-flash:generateContent?key=\$KEY\" \\
  -H \"content-type: application/json\" \\
  -d '{\"contents\":[{\"parts\":[{\"text\":\"hello\"}]}]}'
\`\`\`

If row 1 is 4xx and row 2 is 200, the model is the cause.

## Fix

Bump \`MODEL\` to a current GA Flash model:

\`\`\`py
MODEL = \"gemini-2.5-flash\"  # or whatever current GA Flash is at ship time
\`\`\`

[...]

PR #1776, opened 2026-04-27
ux: show error message when translation API fails unexpectedly (closes #1767)

### LINKED ISSUE #1767: ux: translation error silently renders nothing when API fails unexpectedly
## Problem

When `requestChapterTranslation` throws a non-401, non-403 error, the reader sets:
```ts
setTranslationUsedProvider("error · check admin queue");
setTranslationLoading(false);
```

But the render chain in `reader/[bookId]/page.tsx` (lines 2037–2047) only handles specific values:
- `"cache"` / `"cache · ..."`
- `"translated"` / `"translated · ..."`
- values starting with `"queue failed"`
- falls through to `null` for anything else

So `"error · check admin queue"` renders **nothing** — the user sees a blank translation panel with no feedback that anything went wrong.

## Steps to reproduce

1. Open reader, navigate to a chapter, enable translation
2. Simulate a network failure or 500 error from `/api/translations/request`
3. Translation loading spinner disappears — no error message shown

## Expected

A user-visible error message such as "Translation failed — try again" with a retry action.

## Fix

Add a case for values starting with `"error"` in the render chain (reader/[bookId]/page.tsx line 2045+) that renders a red error span. Also consider changing the message text from admin-specific guidance to a user-friendly fallback.

alfmunny:
## What changed

When `requestChapterTranslation` throws a non-401/403 error, the reader sets `translationUsedProvider` to `"error · check admin queue"`. The render chain only handled specific known values (`"cache"`, `"translated"`, `"queue failed"`) and fell through to `null` for everything else — so users saw a blank translation panel with zero feedback that anything went wrong.

Added an explicit case for values starting with `"error"` that renders a user-visible red error message: _"Translation failed — please try again"_.

## Why

Silent failure is a critical UX gap. Users who hit unexpected translation errors (network failures, 5xx from the AI provider) had no way to know the request failed or to retry.

## Test plan

- Added `frontend/src/__tests__/TranslationErrorFeedback.test.ts` — a static AST assertion that the render chain includes a `startsWith("error")` case.
- All 390 frontend suites and 1382 backend tests pass.
- Reproducing scenario: trigger a 500 from the translation API → reader now shows "Translation failed — please try again" in red instead of blank.

Closes #1767

### DIFF
--- a/frontend/src/__tests__/TranslationErrorFeedback.test.ts
+++ b/frontend/src/__tests__/TranslationErrorFeedback.test.ts
@@ -1,0 +1,20 @@
+/**
+ * Regression for issue #1767:
+ * When requestChapterTranslation throws a non-401/403 error the reader sets
+ * translationUsedProvider to a string starting with "error". The render chain
+ * in reader/[bookId]/page.tsx must have an explicit case for this prefix;
+ * without it the user sees a blank panel with no feedback.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../app/reader/[bookId]/page.tsx"),
+  "utf8",
+);
+
+describe("translation error feedback (closes #1767)", () => {
+  it("render chain handles translationUsedProvider values starting with 'error'", () => {
+    expect(src).toMatch(/translationUsedProvider\.startsWith\(["']error["']\)/);
+  });
+});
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.**

Direct fix for the silent-failure UX gap I diagnosed earlier this session — when a translation request errors out, the reader was rendering a blank panel instead of feedback. This adds an explicit `startsWith("error")` case in the render chain that shows "Translation failed — please try again" in red.

22-line code change + 20-line static AST assertion test. Closes #1767.

Applied `P2`.

**Connection point:** with this + #1778 + the upcoming #1772 fix, today's outage class becomes properly observable — user gets an error message, operator gets a server log line, follow-on diagnosis is fast.

Issue #1078, opened 2026-04-24
fix: POST /ai/translate and POST /ai/tts missing exception logging on 500 responses


alfmunny:
## Bug

`POST /ai/translate` and `POST /ai/tts` have outer `except Exception: raise HTTPException(status_code=500)` handlers that do not log the exception. When an unexpected error occurs (network failure, DB error, unexpected service error) the engineer has no log entry to diagnose the failure.

PR #1070 (closes #1051) added `logger.exception(...)` to `/ai/insight`, `/ai/qa`, `/ai/references`, and `/ai/summary`, but missed `/ai/translate` (outer handler, ~line 466) and `/ai/tts` (~line 478).

## Fix

Add `logger.exception()` calls before the `raise HTTPException` in both handlers, matching the pattern already used for other AI endpoints.

## Regression test

Test that a 500 response from the translate/tts endpoint triggers a log warning (use `caplog`).

alfmunny:
**Dev**: Claimed — starting fix now.

alfmunny:
**Dev**: PR #1105 merged. Removing in-progress label.

PR #1778, opened 2026-04-27
test: regression for AI router exception logging on /ai/insight, /ai/qa, /ai/references, /ai/summary (closes #1771)

### LINKED ISSUE #1771: bug: AI router swallows underlying provider exceptions — admin can't diagnose 'AI service request failed' without server-log access
## Symptom

Admin reports \"AI service request failed\" 500 from \`/ai/insight\` and \`/ai/summary\`. The error message is identical regardless of cause (rate limit, deprecated model, bad key, network outage, malformed Gemini response). PM diagnosed today by reading the source — the user couldn't.

## Root cause

Every Gemini call site in \`backend/routers/ai.py\` is wrapped in a catch-all that throws away the underlying exception:

\`\`\`py
try:
    result = await gemini.generate_insight(key, ...)
except Exception:
    raise HTTPException(status_code=500, detail=\"AI service request failed\")
\`\`\`

Locations:
- \`routers/ai.py:111\` (insight)
- \`routers/ai.py:123\` (qa)
- \`routers/ai.py:148\` (references)
- \`routers/ai.py:204\` (summary)

\`services/gemini.py\` has **zero logging** — the actual exception (e.g. \"model gemini-3.1-flash-lite-preview not found\", \"429 quota exceeded\", \"401 invalid api key\") is never recorded anywhere. Operators have to attach a debugger or comment out the catch to see what broke.

## Why \"keep the user-facing message generic\" is correct

Leaking provider details to the response (e.g. \"Gemini returned 429: …\") is the actual class of bug we shipped #759 / #763 / #764 / #768 / #771 / #775 to fix. So the wire response should stay vague.

The fix is to **log the exception server-side** while keeping the user-facing message identical. Operator gets the diagnostic; user gets nothing they can use to enumerate provider state.

## Fix

In each catch block:

\`\`\`py
import logging
logger = logging.getLogger(__name__)

try:
    result = await gemini.generate_insight(key, ...)
except Exception as e:
    logger.exception(\"AI service request failed in insight: %s\", e)
    raise HTTPException(status_code=500, detail=\"AI service request failed\")
\`\`\`

Same pattern at all four sites. Optional: also \`logger.exception\` inside \`services/gemini.py\` at each provider call so the failing provider/model is captured even when the route catch happens to be high-level.

## Test

Add a router test that mocks \`gemini.generate_insight\` to raise a labelled exception, asserts the 500 response is \"AI service request failed\" (unchanged), and uses \`caplog\` to verify the underlying exception text DOES appear in server logs.

## Why now

The current opaque-error pattern made today's outage (gemini-3.1-flash-lite-preview model name suspected) require 30 minutes of repo-grep to diagnose. Once a second provider is wired (via #1759), the same opacity will block fallback debugging too.

## Labels

\`bug\`, \`P1\` (operability — not a user-visible regression but blocks every future production AI failure investigation).

## Path

Path A — Dev claims, single-router change with logging additions and a router-level test using \`caplog\`.

alfmunny:
## What changed

Adds caplog regression tests for the four AI router endpoints that were missing them: `/ai/insight`, `/ai/qa`, `/ai/references`, `/ai/summary`. Each test:

1. Mocks the underlying Gemini call to raise a `RuntimeError` with a uniquely identifiable diagnostic token (`GEMINI_DIAG_TOKEN_xyzqq_42 — model not found`).
2. Asserts the response is a generic 500 with **no provider detail leaked to the wire** (the diag token must not appear in `resp.json()["detail"]`).
3. Asserts the **diag token DOES appear in server-side log records** — either in the formatted message or in `exc_info` traceback.

## Why

`logger.exception` was added to all six AI catch sites in #1051 (#1070) and #1078 (#1105), but only `/ai/translate` and `/ai/tts` got caplog tests. Without regression coverage for the other four, a future refactor that drops `logger.exception` from one of them won't fail any test — and operators lose the ability to diagnose production AI failures from server logs.

Today's "all AI calls failing" outage (filed under #1772) is exactly this class of failure: when something breaks at the provider layer, the user sees a generic 500 and the operator's only diagnostic path is server logs. These tests lock that path in.

No production code is changed — the existing logging is correct. This PR is pure regression coverage.

## Test plan

- 4 new tests pass: `test_{insight,qa,references,summary}_gemini_error_logs_underlying_exception`
- Each test uses `_has_diag_in_logs(caplog)` helper that walks `caplog.records` and checks both `getMessage()` and `exc_info` for the diagnostic token.
- Each test also asserts the wire detail does not contain the token (no leak).
- Full backend suite: 1929 passed, 0 failures.

Closes #1771

### DIFF
--- a/backend/tests/test_router_ai.py
+++ b/backend/tests/test_router_ai.py
@@ -1843,3 +1843,111 @@ async def test_summary_uses_db_chapter_text_not_client_supplied(client, test_use
     assert "word" in captured_text[0], (
         f"Expected DB chapter text to be passed to Gemini, got: {repr(captured_text[0][:120])}"
     )
+
+
+# ── Issue #1771: Gemini exception text must reach server logs (operability) ───
+#
+# All four AI endpoints catch Gemini exceptions and surface a generic 500 to
+# the user. The user-facing message stays vague (no provider detail leak), but
+# the underlying exception MUST appear in server logs so operators can diagnose
+# the actual cause (deprecated model, quota, bad key, etc.) without attaching
+# a debugger or commenting out the catch block.
+
+
+_DIAG_MSG = "GEMINI_DIAG_TOKEN_xyzqq_42 — model not found"
+
+
+def _has_diag_in_logs(caplog) -> bool:
+    """The underlying exception message must appear somewhere in the log
+    record's text (message or formatted output, including traceback)."""
+    for r in caplog.records:
+        # logger.exception("...: %s", exc) embeds the exception via %s, AND
+        # also attaches the traceback as exc_info — both can carry the diag
+        # token. getMessage() formats the args; format() includes traceback.
+        formatted = r.getMessage()
+        if _DIAG_MSG in formatted:
+            return True
+        if r.exc_info and any(_DIAG_MSG in str(part) for part in r.exc_info if part):
+            return True
+    return False
+
+
+async def test_insight_gemini_error_logs_underlying_exception(client, test_user, caplog):
+    """Regression #1771: POST /ai/insight must log the underlying Gemini
+    exception so operators can diagnose without server-log surgery."""
+    import logging
+    await _set_key(test_user)
+    with patch("routers.ai.gemini") as mock_gemini:
+        mock_gemini.generate_insight = AsyncMock(side_effect=RuntimeError(_DIAG_MSG))
+        with caplog.at_level(logging.ERROR, logger="routers.ai"):
+            resp = await client.post("/api/ai/insight", json={
+                "chapter_text": "text", "book_title": "Book", "author": "Author",
+            })
+    assert resp.status_code == 500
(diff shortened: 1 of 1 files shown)


alfmunny:
**PM review.**

Closes the operability bug (#1771) PM filed earlier this session. 4 caplog regression tests for the 4 missing AI router endpoints (`/insight`, `/qa`, `/references`, `/summary`). Each test:
- Mocks Gemini to raise with a unique diagnostic token
- Asserts the wire response is generic (no token leak — no recurrence of #759/#763/#764/#768/#771/#775)
- Asserts the token IS in caplog records (operator can diagnose)

Pure regression coverage; no production code changes. Test-only, 108 lines.

Applied `P1` (matches #1771's priority).

Once this lands, #1772 (Gemini MODEL = gemini-3.1-flash-lite-preview suspected sunset) becomes diagnosable — the next AI failure will surface the actual exception in logs.

PR #1782, opened 2026-04-27
fix: accept uppercase/mixed-case file extensions in upload (closes #1781)

### LINKED ISSUE #1781: bug: file upload rejects valid files with uppercase extensions (BOOK.TXT, Book.Epub)
## Symptom

Users uploading files with uppercase or mixed-case extensions get a 400 "Only .txt and .epub files are supported" error, even though the file is a valid book.

Affected filenames:
- `BOOK.TXT`
- `Book.Epub`
- `book.EPUB`
- `Novel.TxT`

## Root cause

`backend/routers/uploads.py` lines 109–116:

```py
if filename.endswith(".txt"):
    fmt = "txt"
    max_size = MAX_TXT_BYTES
elif filename.endswith(".epub"):
    fmt = "epub"
    max_size = MAX_EPUB_BYTES
else:
    raise HTTPException(status_code=400, detail="Only .txt and .epub files are supported.")
```

`str.endswith` is case-sensitive. macOS and Windows filesystems are typically case-insensitive, so a user dragging in `Book.EPUB` may not realize the case matters. The upload silently fails with a confusing error.

## Fix

Lowercase the extension before the check:

```py
filename_lower = filename.lower()
if filename_lower.endswith(".txt"):
    fmt = "txt"
    max_size = MAX_TXT_BYTES
elif filename_lower.endswith(".epub"):
    ...
```

Note: the original `filename` is still stored as-is in the DB (preserving case for display). Only the format detection is normalized.

## Test

Regression test: upload a file named `BOOK.TXT` and assert it succeeds (status 200, returns book_id). Same for `Book.Epub`.

## Labels

`bug`, `P2` (UX paper-cut, not a security or data issue, but a confusing failure mode for non-technical users).

alfmunny:
## What changed

`backend/routers/uploads.py`: lowercase the filename before extension matching. The `str.endswith` check was case-sensitive, so `BOOK.TXT`, `Book.Epub`, `novel.EPUB` all failed with the misleading "Only .txt and .epub files are supported" error.

The original `filename` is still stored as-is in the DB (preserves user's casing for display); only the format-detection branch is normalized.

## Why

macOS and Windows filesystems are typically case-insensitive, so users dragging in files with uppercase extensions don't realize the case matters. The upload silently fails at the API boundary with a confusing 400. Real bug found during a routine bug-hunt scan.

## Test plan

Two parameterized regression tests, six total cases:
- `test_upload_accepts_uppercase_txt_extension` — `BOOK.TXT`, `Book.Txt`, `novel.TXT` all return 200 with `format=txt`
- `test_upload_accepts_uppercase_epub_extension` — `BOOK.EPUB`, `Book.Epub`, `novel.EPub` all return 200 with `format=epub`
- Full backend suite: 1939 passed, 0 failures

Closes #1781

### DIFF
--- a/backend/routers/uploads.py
+++ b/backend/routers/uploads.py
@@ -106,10 +106,11 @@ async def upload_book(
     filename = file.filename or ""
     if len(filename) > 255:
         raise HTTPException(status_code=422, detail="Filename too long (max 255 characters).")
-    if filename.endswith(".txt"):
+    filename_lower = filename.lower()
+    if filename_lower.endswith(".txt"):
         fmt = "txt"
         max_size = MAX_TXT_BYTES
-    elif filename.endswith(".epub"):
+    elif filename_lower.endswith(".epub"):
         fmt = "epub"
         max_size = MAX_EPUB_BYTES
     else:
--- a/backend/tests/test_router_uploads.py
+++ b/backend/tests/test_router_uploads.py
@@ -1149,3 +1149,29 @@ async def test_confirm_chapters_unmatched_original_index_saves_empty_text(client
     ch = chapters_resp.json()["chapters"][0]
     assert ch["title"] == "Phantom Chapter"
     assert ch["text"] == ""  # text="" because orig_idx not in orig_by_idx (line 270)
+
+
+# ── Issue #1781: extension check must be case-insensitive ─────────────────────
+
+
+@pytest.mark.parametrize("filename", ["BOOK.TXT", "Book.Txt", "novel.TXT"])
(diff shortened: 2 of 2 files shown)


alfmunny:
**PM review.**

Clean bug-hunt find. Lowercase normalization on filename before extension matching, original casing preserved in DB. 3-line code change + 26-line parameterised test (6 cases across .TXT/.Txt/.EPUB/.EPub). Closes #1781.

Applied `P2` per label policy (real upload-blocker for users with uppercase extensions, common on macOS/Windows; workaround is rename, so not P1).

PR #1397, opened 2026-04-25
fix: fix invalid Gemini model name (500 on Insight) + Insight panel page-scroll bug

alfmunny:
## What changed

Two user-reported production bugs in the Insight chat feature:

**1. Invalid Gemini model name (#1388 — P0)**
`backend/services/gemini.py` had `MODEL = "gemini-3.1-flash-lite-preview"` which is not a valid Gemini API model. Every call to `POST /api/ai/insight` and `POST /api/ai/qa` threw an exception, caught and re-raised as HTTP 500. Updated to `gemini-2.0-flash-lite` (stable, equivalent flash-lite tier model).

**2. Opening Insight panel scrolls page to top (#1389 — P1)**
`InsightChat.tsx` auto-scroll effect called `bottomRef.current?.scrollIntoView({ behavior: "smooth" })` without `block: "nearest"`. `scrollIntoView` scrolls ALL ancestor containers including the document, so every time the Insight panel opened and loaded its first message, the reader page jumped to the top. Fixed by scrolling only the messages container directly (`messagesBoxRef.current.scrollTop = messagesBoxRef.current.scrollHeight`).

## Test plan

- Backend: added `test_model_name_is_stable_and_valid` to `test_gemini.py` — asserts MODEL is not a preview alias and follows a valid Gemini identifier pattern.
- Frontend: added regression test in `InsightChat.coverage2.test.tsx` — mocks `scrollIntoView` and asserts it is NOT called when the panel opens and loads.
- Full backend suite: 1665 passed ✓
- Full frontend suite: 2483 passed ✓

Closes #1388, #1389

PR #262, opened 2026-04-22
feat: AI chapter summary with shared cache

alfmunny:
## Summary

- Add `POST /ai/summary` endpoint that generates a structured chapter summary using the queue Gemini key (no personal key required)
- Summaries are cached in a new `chapter_summaries` table and shared across all users — first reader pays the Gemini cost, everyone else gets instant cache hits
- Add **Summary** tab to the reader sidebar with auto-load on first open, loading skeleton, cached badge, and admin refresh button

## What changed

**Backend**
- `backend/migrations/018_chapter_summaries.sql` — new table `chapter_summaries(book_id, chapter_index, model, content, created_at)`
- `backend/services/db.py` — `get_chapter_summary()` and `save_chapter_summary()` with upsert-on-conflict
- `backend/services/gemini.py` — `generate_chapter_summary()` + `SYSTEM_SUMMARY` prompt (structured: Overview / Key Events / Characters / Themes)
- `backend/routers/ai.py` — `POST /ai/summary` (cache-first, uses queue key) + `DELETE /ai/summary` (admin-only refresh)

**Frontend**
- `frontend/src/components/ChapterSummary.tsx` — new component: auto-loads, skeleton loader, markdown render, refresh button, chapter-change reset
- `frontend/src/app/reader/[bookId]/page.tsx` — adds "📋 Summary" button to toolbar and Summary panel to sidebar
- `frontend/src/lib/api.ts` — `generateChapterSummary()` API client function

## Test plan

- [x] Backend: 9 new tests in `test_router_summary.py` — cache hit, cache miss, no queue key → 503, book not found → 404, Gemini failure → 500, admin delete, non-admin delete → 403
- [x] Backend: 4 new DB tests in `test_db.py` — get/save chapter summary, overwrite, null model
- [x] Frontend: 9 new component tests in `ChapterSummary.test.tsx` — auto-load, no-load when hidden, cached badge, error state, 503 message, refresh, chapter reset, empty state
- [x] Full backend suite: 874 passed
- [x] Full frontend suite: 1525 passed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1786, opened 2026-04-27
chore: remove stale empty-string entry from MODEL_LIMITS (closes #1783)

### LINKED ISSUE #1783: chore: remove stale empty-string entry from MODEL_LIMITS (misleading comment, unused values)
## Symptom

`backend/services/model_limits.py` lines 44-50 carry a stale comment + stale values for the `""` entry in `MODEL_LIMITS`:

```py
MODEL_LIMITS: dict[str, _Limits] = {
    # "" = server-side default (currently gemini-3.1-flash-lite-preview).
    # Same quota + pricing as gemini-3.1-flash-lite on Tier 1.
    "": {
        "rpm": 4000, "rpd": 150000, "max_output_tokens": 7500,
        "input_usd_per_m": 0.10, "output_usd_per_m": 0.40,
    },
```

The comment says the server default is `gemini-3.1-flash-lite-preview`, but that was changed to `gemini-2.0-flash-lite` in PR #1397 (2026-04-26). The values still match the OLD default, not the current one.

## Why it matters

This stale comment was the basis for issue #1772's premise ("MODEL is pinned to a stale preview alias"), which it turned out was already fixed. The comment is a forensic landmine — anyone tracing what "server default" means will be misled.

The `""` entry itself is dead code: every caller of `limits_for(model)` passes a real model name from `get_model_chain()` / `DEFAULT_CHAIN`. No code path ever queries `""` in production, so the wrong values cause no actual incorrect behavior today. But the documentation is wrong and confusing.

## Fix

Remove the `""` entry entirely (since it's never queried), or update its comment + values to reflect the actual current default (`gemini-2.0-flash-lite`: rpm 4000, rpd UNLIMITED_RPD, input $0.075/M, output $0.30/M).

Removing is cleaner — the `MODEL_LIMITS.get(model, CUSTOM_LIMITS)` fallback already handles unknown models conservatively, which is the correct behavior for any unexpected `""` lookup.

## Test

The existing tests don't reference `""` (`MODEL_LIMITS[""]` is not used in any test). After removal, run the full suite to confirm nothing breaks. Add a tiny regression test asserting the dict has no empty-string key (locks in the cleanup).

## Labels

`bug`, `P3` (cosmetic / forensic — dead code with misleading docs, no current functional impact, but its presence enabled the false-alarm bug report #1772).

alfmunny:
## What changed

Removed the `""` (empty-string) entry from `MODEL_LIMITS` in `backend/services/model_limits.py`. The entry was an alias for an old server default that was bumped in PR #1397 — its comment + values were never updated, leaving them as forensic landmines.

Added `backend/tests/test_model_limits.py` with three small tests:
1. `test_model_limits_has_no_empty_string_key` — locks in the absence of the stale alias.
2. `test_default_chain_models_all_have_limits` — every model in `DEFAULT_CHAIN` must have a real entry.
3. `test_limits_for_unknown_model_falls_back_to_custom` — `""` and unknown models correctly fall through to `CUSTOM_LIMITS`.

## Why

In practice nothing queries `limits_for("")` — every callsite passes a real model name from the configured chain. The `MODEL_LIMITS.get(model, CUSTOM_LIMITS)` fallback already handles unknown lookups conservatively, which is the correct default for any unexpected `""` query.

But the stale comment ("currently gemini-3.1-flash-lite-preview") was the basis for the false-alarm bug report #1772, where an Architect mistakenly believed the server default still pinned to a stale preview model. Removing the entry eliminates the misleading documentation source.

## Test plan

- 3 new model_limits tests pass
- Translation queue suite (118 tests) unaffected
- Full backend suite: 1942 passed, 0 failures (+3 from the new tests)

Closes #1783

### DIFF
--- a/backend/services/model_limits.py
+++ b/backend/services/model_limits.py
@@ -42,12 +42,6 @@ DEFAULT_CHAIN: list[str] = [
 UNLIMITED_RPD = 1_000_000  # treat as effectively unlimited
 
 MODEL_LIMITS: dict[str, _Limits] = {
-    # "" = server-side default (currently gemini-3.1-flash-lite-preview).
-    # Same quota + pricing as gemini-3.1-flash-lite on Tier 1.
-    "": {
-        "rpm": 4000, "rpd": 150000, "max_output_tokens": 7500,
-        "input_usd_per_m": 0.10, "output_usd_per_m": 0.40,
-    },
     # Frontier 3.1 family
     "gemini-3.1-pro-preview": {
         "rpm": 25, "rpd": 250, "max_output_tokens": 60000,
--- a/backend/tests/test_model_limits.py
+++ b/backend/tests/test_model_limits.py
@@ -1,0 +1,35 @@
+"""Tests for services.model_limits."""
+
+from services.model_limits import (
+    MODEL_LIMITS,
+    DEFAULT_CHAIN,
+    CUSTOM_LIMITS,
+    limits_for,
+)
+
+
+def test_model_limits_has_no_empty_string_key():
+    """Regression #1783: the empty-string key was a stale alias for the old
+    server default (gemini-3.1-flash-lite-preview). Removed in cleanup —
+    callers always pass real model names from the chain. Lock in absence."""
+    assert "" not in MODEL_LIMITS, (
+        "MODEL_LIMITS must not contain an empty-string key — that entry was "
+        "a stale alias for an old default model and is no longer needed."
+    )
+
+
+def test_default_chain_models_all_have_limits():
+    """Every model in DEFAULT_CHAIN must have a corresponding MODEL_LIMITS
+    entry — otherwise the queue worker would fall back to CUSTOM_LIMITS for
+    well-known curated-default models."""
+    for model in DEFAULT_CHAIN:
+        assert model in MODEL_LIMITS, (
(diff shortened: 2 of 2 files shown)


alfmunny:
**PM review.**

Clean chore: removes a stale `""`-key entry from `MODEL_LIMITS` whose comment was outdated ("currently gemini-3.1-flash-lite-preview"). The fallback path `MODEL_LIMITS.get(model, CUSTOM_LIMITS)` already handles unknown lookups, so dropping the alias is safe. 3 new tests lock in absence + fallback behaviour. Closes #1783.

Applied `P3` per label policy.

**Note on the "#1772 false alarm" framing:** the PR body characterises #1772 as a misdiagnosis caused by this stale comment. That conflates two separate things:
- This stale comment in `model_limits.py` (informational — never affected runtime)
- The actual `MODEL = "gemini-3.1-flash-lite-preview"` constant at `services/gemini.py:10` (which IS the runtime model and is what #1772 cited)

#1772 is **not** automatically a false alarm just because this comment is gone. The runtime constant in `gemini.py` is still pinned to the preview model name and hasn't been touched since PR #262. Whether Google has actually sunset that preview model is the open question — needs the curl-verification step from #1772 to settle. The verification can now run cleanly because #1771 caplog regressions (#1778 merged) will surface the actual exception text on the next failure.

Approving this chore. #1772 stays open until verified.

Issue #1340, opened 2026-04-25
ux: TypographyPanel floating settings panel missing role and aria-controls linkage (WCAG 4.1.2)


alfmunny:
## Problem

The `TypographyPanel` component renders as a floating interactive widget (font size, spacing, width, font family, paragraph focus toggle) but has no ARIA role or accessible name on its root element. Screen readers cannot identify it as a discrete interactive region.

Additionally, the two trigger buttons in `reader/[bookId]/page.tsx` use `aria-expanded` but are missing `aria-controls`, so assistive technology cannot programmatically navigate from the button to the panel.

## WCAG violations

- **4.1.2 Name, Role, Value (Level A):** UI components must have accessible name and role. The panel div has neither.
- **4.1.2** additionally: trigger buttons with `aria-expanded` should reference the controlled element via `aria-controls`.

## Reproduction

Open the reader, click the `Aa` typography button, inspect with a screen reader (NVDA/VoiceOver). The panel opens but is not announced as a dialog or labelled region.

## Fix

1. Add `id="typography-panel"`, `role="dialog"`, and `aria-label="Typography settings"` to the root `<div>` in `TypographyPanel.tsx`.
2. Add `aria-controls="typography-panel"` to both Aa trigger buttons in `reader/[bookId]/page.tsx` (lines ~880 and ~1006).

## Files

- `frontend/src/components/TypographyPanel.tsx` (panel root)
- `frontend/src/app/reader/[bookId]/page.tsx` (two trigger buttons)

alfmunny:
**UI/UX**: Claimed by UI/UX Dev — starting work now.

PR #1795, opened 2026-04-27
ux: remove role=dialog from TypographyPanel — non-modal popover (closes #1789)

### LINKED ISSUE #1789: ux: TypographyPanel uses role=dialog without aria-modal — screen readers roam page behind panel
## Problem

`TypographyPanel.tsx` has `role="dialog"` but **no `aria-modal="true"`**.

Per WAI-ARIA spec, `role="dialog"` without `aria-modal="true"` means assistive technologies (VoiceOver, NVDA) continue to expose background content to users while the panel is visible. This breaks WCAG 4.1.2 Name/Role/Value (Level A).

The panel also has no focus trap — keyboard users can Tab past it into background content, which is visually confusing since the panel blocks the view.

## Reproduction

1. Open reader → click Typography (Aa) button  
2. With VoiceOver: navigate with VO+right — AT wanders into content behind the panel

## Fix

Two options:

**Option A (preferred for a settings popover):** Remove `role="dialog"` entirely. The panel is non-modal by design; it closes on click-outside or Escape. Use `role="region"` or no role, keep `aria-label="Typography settings"`. Focus stays wherever the user left it.

**Option B (full modal):** Keep `role="dialog"`, add `aria-modal="true"`, implement a basic focus trap (move focus to first focusable child on open; Tab wraps within panel; restore focus on close).

Option A is ~2 lines; Option B is ~20 lines. Either eliminates the AT-roaming issue.

## Scope

Single component: `src/components/TypographyPanel.tsx` line 132.

alfmunny:
## What changed

Removed `role="dialog"` from `TypographyPanel.tsx` (line 132). The panel keeps `aria-label="Typography settings"` and `id="typography-panel"`.

Also updated `TypographyPanelRole.test.ts` to reflect the corrected semantics from #1789 — the original #1340 fix needed `aria-label` + `id`, not `role="dialog"`.

## Why

Closes #1789. `role="dialog"` without `aria-modal="true"` causes screen readers (VoiceOver, NVDA) to continue exposing background page content while the panel is visible — a WCAG 4.1.2 Level A violation.

The typography settings panel is a non-modal popover (closes on click-outside or Escape; does not trap focus). Using `role="dialog"` incorrectly implied modal semantics. Removing the role eliminates the AT-roaming issue while keeping the accessible name (`aria-label`) so the panel is still announced correctly.

## Test plan

- `TypographyPanelDialogRole.test.ts` — new test that asserts: if `role="dialog"` is present it must be paired with `aria-modal="true"`; passes because role is now absent
- `TypographyPanelRole.test.ts` — updated to reflect corrected semantics (removed assertion for `role="dialog"`, retained `aria-label` + `id` + `aria-controls` checks)
- All 2576 frontend tests pass

## Docs follow-up

### DIFF
--- a/frontend/src/__tests__/TypographyPanelDialogRole.test.ts
+++ b/frontend/src/__tests__/TypographyPanelDialogRole.test.ts
@@ -1,0 +1,27 @@
+/**
+ * Accessibility regression for issue #1789:
+ * TypographyPanel used role="dialog" without aria-modal="true", causing
+ * screen readers to roam page content behind the panel (WCAG 4.1.2 Level A).
+ * Fix: remove role="dialog"; the panel is a non-modal settings popover —
+ * no role annotation is needed (aria-label is sufficient).
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../components/TypographyPanel.tsx"),
+  "utf8",
+);
+
+describe("TypographyPanel dialog role (closes #1789)", () => {
+  it("does not use role=dialog without aria-modal", () => {
+    const hasDialogRole = /role="dialog"/.test(src);
+    const hasAriaModal = /aria-modal="true"/.test(src);
+    // Either no role=dialog, or role=dialog accompanied by aria-modal=true
+    if (hasDialogRole) {
+      expect(hasAriaModal).toBe(true);
+    } else {
+      expect(hasDialogRole).toBe(false);
+    }
+  });
+});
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM review.**

A11y correction: removes `role="dialog"` from TypographyPanel — the panel is a non-modal popover (no focus trap, closes on outside click), so `role="dialog"` without `aria-modal="true"` mis-announced semantics and let AT roam to background content. WCAG 4.1.2 fix. `aria-label` + `id` + `aria-controls` retained.

1-line component change + 27-line new test asserting absence-of-bad-pattern + 11-line update to existing test. Closes #1789.

The new test "if role=dialog is present then aria-modal must be true" is a useful pattern — passing because role is absent. Good defensive shape.

Applied `P2`.

PR #1821, opened 2026-04-28
ux: DeckCard navigation — deck list no longer a dead end (closes #1820)

### LINKED ISSUE #1820: ux: DeckCard on /decks page has no navigation to deck detail — dead-end UX
## Problem

The `/decks` page renders deck cards but clicking on a card does nothing. There is no way to navigate from the deck list to an individual deck's detail/edit page (`/decks/{id}`).

The `DeckCard` component has a delete button, but no link or button to open the deck. Users who want to manage deck membership (add/remove words) have no path to the deck detail page from the list.

## Reproduction

1. Navigate to `/decks`
2. Click anywhere on a deck card (not the trash icon) — nothing happens
3. There is no button/link to view/edit a deck

## Expected

Clicking the deck card content area (or an explicit "View" / "Open" button) navigates to `/decks/{id}`

## Fix

Add an `onClick` prop to `DeckCard`. Make the main content area a clickable button that calls `router.push('/decks/' + deck.id)` from the `/decks` page. Keep delete button separate.

alfmunny:
## What changed

- Added an `onClick?: () => void` prop to `DeckCard`
- When `onClick` is provided, the card content area renders as a `<button>` (`data-testid="deck-card-link-{id}"`, `aria-label="Open deck {name}"`) that calls `onClick` on click
- The delete button remains a separate element so clicking delete does not trigger navigation
- In `decks/page.tsx`, each `DeckCard` now receives `onClick={() => router.push('/decks/' + d.id)}`

## Why

The `/decks` page listed all decks with no way to navigate into a deck's detail page. Clicking a card did nothing — a complete dead-end for users who want to view, study, or manage the words in a deck. Closes #1820.

## Test plan

- [x] Added 4 new unit tests to `DeckCard.test.tsx`:
  - Clicking the content button calls `onClick`
  - Content button carries an accessible label including the deck name
  - Content button is absent when `onClick` is not supplied
  - Clicking content button does not trigger `onDelete`
- [x] All 2604 frontend tests pass
- [x] All 1942 backend tests pass
- Manual: navigate to `/decks`, click any deck card body → lands on `/decks/{id}` detail page; delete button still works independently

## Docs follow-up

Design change log updated in `docs/design-improvement-plan.md` (Wave 12).

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -283,6 +283,18 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 |------|--------|---------|----|
 | 2026-04-28 | Focus-mode HUD truncated chapter title: added `title` tooltip for hover-reveal | reader/[bookId]/page.tsx | #1770 |
 
+## Wave 12 — Navigation + Contrast (2026-04-28)
+
+### WCAG 1.4.3 Placeholder contrast
+| Date | Change | File(s) | PR |
+|------|--------|---------|----|
+| 2026-04-28 | `placeholder:text-stone-400` (2.4:1) → `placeholder:text-stone-600` (7.2:1) on 4 inputs | decks/[deckId]/page.tsx, AnnotationToolbar.tsx, SearchBar.tsx, InsightChat.tsx | #1819 |
+
+### UX dead-end fixes
+| Date | Change | File(s) | PR |
+|------|--------|---------|----|
+| 2026-04-28 | DeckCard: added `onClick` prop + content-area button so `/decks` list navigates to `/decks/{id}` | DeckCard.tsx, decks/page.tsx | pending #1820 |
+
 ---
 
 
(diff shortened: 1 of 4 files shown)


alfmunny:
**PM review.**

Fixes a real dead-end: `/decks` page had no way to click into a deck. Adds `onClick` prop to DeckCard, body renders as `<button>` with aria-label, delete button kept separate to avoid event-bubble navigation. 4 new tests cover the surfaces.

Approved on content. Applied `P2`.

🟢 **Non-blocking follow-up:** `<button onClick={() => router.push(...)}>` works but is semantically a link, not an action. Users lose right-click "Open in new tab", cmd-click, hover URL preview, and crawler indexing. Worth a P3 follow-up to convert to `<Link href={`/decks/${id}`}>` from `next/link` — same affordance, real anchor semantics. Not blocking this PR; the dead-end is the actual user-blocking issue and it's fixed here.

Issue #1790, opened 2026-04-27
feat: implement pricing plans Free/Pro/Premium per design doc (impl tracker for #1774)


alfmunny:
## Context

Design doc **`docs/design/pricing-plans.md`** merged via PR #1779 with \`user-approved\`. This is the implementation tracker.

## Open questions for the user (gate impl start)

The design explicitly flagged three settle-before-impl decisions:

1. **Pro vs Premium differentiator** — proposed: Pro = capped (50/mo), Premium = uncapped + priority queue + extras. Alternatives in §"Open questions" #1 of the design.
2. **Pricing values ($/month)** — both tiers. Code reads from env vars; user decides values.
3. **Reading-quota reset cadence** — calendar month (recommended) vs rolling 30-day.

**Implementation should not begin until these three are settled.** Apply \`user-approved\` to this issue (or a comment with the answers) when ready.

## Scope (from design)

- **Migration 038** — \`users.tier\` + Stripe columns; default \`'free'\`, no data movement.
- **Migration 039** — \`billing_events\` audit log with \`stripe_event_id UNIQUE\` for idempotency.
- **Authorization layer** — \`require_tier(min_tier)\`, \`require_book_quota\`, \`require_translation_quota\` dependencies.
- **5 gated translation routes** — \`POST /books/{id}/chapters/{n}/translation\`, \`/translation/retry\`, \`/translations/enqueue-all\`, \`/ai/translate\`, plus the GET \`/books/{id}/chapters\` book-quota gate.
- **5 new billing routes** — \`POST /billing/checkout\`, \`POST /billing/portal\`, \`POST /billing/webhook\`, \`GET /billing/me\`, \`PUT /user/byok\`.
- **Daily cron** \`scripts/expire_subscriptions.py\` for defensive downgrade on missed webhooks.
- **Frontend** — \`/pricing\` page, \`/account/billing\` page, \`<PaywallModal>\` with three variants, plan badge in header.
- **Tests** — 18 backend + 4 frontend + 2 E2E (Stripe test-mode flow).

## Acceptance (full list in design §"Acceptance")

- Migrations 038 + 039 ship with backfill tests.
- All three new \`require_*\` dependencies tested independently.
- All 5 gated routes return 402 with structured detail when blocked.
- Webhook handler is idempotent (UNIQUE constraint on \`stripe_event_id\`).
- E2E test against Stripe test mode covers Pro checkout → tier flip → portal cancel → tier flip back.
[...]

PR #1825, opened 2026-04-28
feat: pricing-plans PR D — daily expire-subscriptions cron (#1790)

alfmunny:
## Summary

**PR D** of the pricing-plans series. Defensive belt-and-suspenders against missed Stripe webhooks: any user whose `tier_period_end` has passed AND has no active subscription gets downgraded to `'free'`.

## Why this script exists

The Stripe webhook (PR C2) handles `customer.subscription.deleted` by clearing `stripe_subscription_id` but **NOT immediately downgrading** the tier — user paid through `tier_period_end`. This script enforces the actual downgrade after the period expires.

It also catches the failure modes where:
1. Stripe webhook drop — the `.deleted` event never arrived; we still downgrade based on the period end alone.
2. Active subscription with expired period — Stripe should have sent `subscription.updated` to extend the period; if it didn't, this script catches the gap.

## Selection guard

Only users with `stripe_subscription_id IS NULL` get downgraded. **Active subscriptions stay paid** even if `tier_period_end` is stale (Stripe will send `subscription.updated` to extend it; we don't race ahead of the webhook).

## Usage

```bash
python -m backend.scripts.expire_subscriptions          # apply
python -m backend.scripts.expire_subscriptions --dry-run  # report only
```

## Tests

8 new in `test_expire_subscriptions.py`:
- pro → free downgrade
- premium → free downgrade  
- active subscription stays pro (guard: `stripe_subscription_id IS NOT NULL`)
- future tier_period_end stays pro
- free user no-op
- `--dry-run` preserves DB
- summary contains user_id / email / from_tier / period_end
- empty-candidates path returns 0 / 0

## User-only follow-up

Wire the script into the host scheduler (Railway / systemd timer / cron) at a daily off-minute (e.g. 03:17 UTC). Will file as part of the consolidated user-only Stripe-config issue.

## Series progress

- **A** ✓ (migrations + require_tier + 4 translation-route gates)
- **B** ✓ (require_book_quota — 3-books/month free)
- **C1** ✓ (GET /billing/me)
- **C2** ✓ (Stripe webhook handler with idempotency + tier transitions)
- **C3** ✓ (POST /billing/checkout + /billing/portal)
- **D** (this PR — daily expire cron)
- **E** — frontend pricing + billing UI
- **F** — BYOK toggle

Part of #1790.

### DIFF
--- a/backend/scripts/expire_subscriptions.py
+++ b/backend/scripts/expire_subscriptions.py
@@ -1,0 +1,132 @@
+"""Daily cron: defensively downgrade users whose paid period has ended.
+
+PR D of the pricing-plans series (#1790 / docs/design/pricing-plans.md).
+
+The Stripe webhook (PR C2) handles `customer.subscription.deleted` by
+clearing `stripe_subscription_id` and leaving the tier alone — the user
+paid through `tier_period_end`, so they keep paid features until then.
+This script enforces the actual downgrade: any user whose
+`tier_period_end < now()` AND has no active subscription gets bumped
+back to `'free'`.
+
+Belt-and-suspenders for two failure modes:
+  1. Stripe webhook drop — the .deleted event never arrived; we still
+     downgrade based on the period end alone.
+  2. Active subscription with expired period — Stripe should have sent
+     subscription.updated to extend the period; if it didn't, this
+     script catches the gap.
+
+Usage
+-----
+    python -m backend.scripts.expire_subscriptions          # default
+    python -m backend.scripts.expire_subscriptions --dry-run  # report only
+
+Schedule: daily at an off-minute (e.g. 03:17 UTC). Cheap query, indexed
+on tier_period_end via the existing users PRIMARY KEY scan.
+"""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import logging
+import sys
+from datetime import datetime, timezone
+
+import aiosqlite
+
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.**

PR D of pricing-plans — the defensive cron for missed webhooks. 132-line script + 146-line tests.

**Strong calls:**
- **Selection guard `stripe_subscription_id IS NULL` only** — active subscriptions stay paid even when `tier_period_end` is stale, avoiding the race against pending `subscription.updated` webhooks. Right call to not let a defensive cron clobber a healthy subscription.
- **Two failure modes covered**: webhook drop (`.deleted` never arrived) AND active-with-expired-period (Stripe should send `updated` but didn't). The cron catches both.
- **`--dry-run` flag** — operator can verify before applying. Standard.
- **Off-minute schedule recommendation (03:17 UTC)** — not :00, won't dogpile with the rest of the planet's cron schedules. Good ops hygiene.

8 tests cover: each tier's downgrade path, active-stays guard, future-period stays, free no-op, --dry-run preserves DB, summary content, empty-candidates returns 0/0.

Applied `P1` per label policy.

Series progress: A ✓ B ✓ C1 ✓ C2 ✓ C3 ✓ D (this PR). Remaining: E (frontend), F (BYOK toggle), user-only Stripe Dashboard config + cron-scheduler wiring.

PR #1838, opened 2026-04-28
ux: guard BUTTON/A in reader keyboard shortcut handler (WCAG 2.1.1)

### LINKED ISSUE #1837: ux: Space key TTS shortcut intercepts button activation in reader page (WCAG 2.1.1)
## Problem

The reader page registers a global `keydown` listener that intercepts the Space key for TTS play/pause. The guard only skips `INPUT`, `TEXTAREA`, and `SELECT`:

```js
const tag = (e.target as HTMLElement)?.tagName;
if (tag === "INPUT" || tag === "TEXTAREA" || tag === "SELECT") return;

// ...
} else if (e.key === " ") {
  e.preventDefault();  // ← fires even when a <button> is focused!
  ttsControlsRef.current?.play();
}
```

When a keyboard user focuses any `<button>` on the reader page (e.g., "Next chapter", sidebar tabs, "Confirm & translate") and presses Space:

1. The guard passes (tagName is "BUTTON", not INPUT/TEXTAREA/SELECT)
2. `e.preventDefault()` suppresses the native button click
3. TTS is toggled instead

**This breaks Space-to-activate for every button on the page when TTS is running.** Screen readers also rely on Space to activate controls.

**File:** `frontend/src/app/reader/[bookId]/page.tsx`, lines 694–729

## Proposed fix

Extend the guard to also return early when the focused element is a `<button>`, `<a>`, or has `role="button"`:

```js
const tag = (e.target as HTMLElement)?.tagName;
if (tag === "INPUT" || tag === "TEXTAREA" || tag === "SELECT" || tag === "BUTTON" || tag === "A") return;
// or: if ((e.target as HTMLElement)?.closest('button, a, [role="button"]')) return;
```

The Space shortcut for TTS only makes sense when focus is on the reading content itself (the prose), not on interactive controls.

## Labels
- ux, P1 (breaks core keyboard navigation in the reader)

alfmunny:
## What changed

Expanded the keyboard shortcut guard in `reader/[bookId]/page.tsx` to include `BUTTON` and `A` tags.

**Before:** only `INPUT`, `TEXTAREA`, `SELECT` were excluded from the global `keydown` handler.

**After:** `BUTTON` and `A` are also excluded — so Space on a focused button activates the button natively instead of being swallowed by the TTS play/pause shortcut.

## Why

WCAG 2.1.1 (Keyboard) requires that all functionality operable by pointer is also operable by keyboard. The Space key on a button must activate it. The prior guard called `e.preventDefault()` on Space for any focused element that wasn't an input field, which silently broke Space-to-activate for every `<button>` in the reader (annotation toggles, panel close buttons, etc.).

## Test plan

- `ReaderSpaceShortcutGuard.test.ts` — 2 static-source assertions:
  - guard includes `tag === "BUTTON"`
  - guard includes `tag === "A"`
- Full frontend suite: 2627/2627 ✓

Closes #1837

### DIFF
--- a/frontend/src/__tests__/ReaderSpaceShortcutGuard.test.ts
+++ b/frontend/src/__tests__/ReaderSpaceShortcutGuard.test.ts
@@ -1,0 +1,19 @@
+import { readFileSync } from "fs";
+import { join } from "path";
+
+const src = readFileSync(
+  join(__dirname, "../app/reader/[bookId]/page.tsx"),
+  "utf-8"
+);
+
+describe("Reader keyboard shortcut guard (WCAG 2.1.1)", () => {
+  it("keyboard shortcut guard must skip BUTTON tag to preserve Space-to-activate", () => {
+    // The guard must include BUTTON so Space key on a focused button
+    // activates the button rather than triggering TTS play/pause.
+    expect(src).toMatch(/tag === "BUTTON"/);
+  });
+
+  it("keyboard shortcut guard must skip A tag to preserve Space/Enter on links", () => {
+    expect(src).toMatch(/tag === "A"/);
+  });
+});
--- a/frontend/src/app/reader/[bookId]/page.tsx
+++ b/frontend/src/app/reader/[bookId]/page.tsx
@@ -692,9 +692,10 @@ export default function ReaderPage() {
   // Keyboard shortcuts
   useEffect(() => {
     function handleKeyDown(e: KeyboardEvent) {
-      // Skip if user is typing in an input/select/textarea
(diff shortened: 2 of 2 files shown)


alfmunny:
**PM review.**

Real keyboard-functionality fix: the reader's Space shortcut for TTS play/pause was preventDefault-ing on every focused element except inputs — so Space on a focused `<button>` got swallowed instead of activating the button. WCAG 2.1.1. Fix expands the guard to skip BUTTON and A elements.

3-line component change + 19-line static-assertion test. Closes #1837. Applied `P2`.

PR #1849, opened 2026-04-28
ux: add aria-live region to search results for screen reader announcement (closes #1848)

### LINKED ISSUE #1848: ux: search results page missing aria-live announcement when results load (WCAG 4.1.3)
## Problem

When a screen reader user submits a search query on `/search`, the loading state is announced (`role="status" aria-label="Searching"`) but when results arrive, there is no announcement. The user has no way to know:
- How many results were found
- That the search completed (from the "no results" empty state either)

The vocabulary page (`vocabulary/page.tsx:470`) correctly uses `role="status" aria-live="polite" aria-atomic="true"` for filtering — the same pattern should be applied here.

## Reproduction

1. Use a screen reader (VoiceOver / NVDA)
2. Navigate to `/search`
3. Submit a search query
4. Loading spinner is announced: "Searching" ✓
5. Results appear silently — no announcement ✗

## Fix

Add a visually-hidden `aria-live="polite"` status element in `SearchResultsInner` that announces the result count after loading completes:

```tsx
  {!loading && data && data.total > 0 && `Found ${data.total} results for ${q}`}
  {!loading && data && data.total === 0 && `No matches for ${q}`}
</div>
```

## File

`frontend/src/app/search/page.tsx` — `SearchResultsInner` component

## WCAG

4.1.3 Status Messages (Level AA)

alfmunny:
## What changed

`SearchResultsInner` (`frontend/src/app/search/page.tsx`) now includes a persistent `role="status" aria-live="polite" aria-atomic="true"` element that announces the search outcome to screen readers:

- **Results found**: "Found N results for {query}."
- **No results**: "Search complete. No results found for {query}."
- **Error**: "Search failed."

Previously, only the loading state was announced (`role="status" aria-label="Searching"`). Result arrival was silent — screen reader users had no way to know the search completed.

## Why

WCAG 4.1.3 (Status Messages, Level AA) requires dynamically injected completion messages to be reachable by assistive technologies without requiring focus. The vocabulary filter page already uses this pattern correctly (`aria-live="polite"` on line 470) — the search page now matches it.

## Test plan

- `SearchResultsLiveRegion.test.ts` (new) — verifies `aria-live="polite"`, `aria-atomic="true"`, and `role="status"` are present in the search page source
- All 2645 existing Jest tests pass
- Existing `SearchPageUx.test.ts` assertions (role=alert on error, Browse books CTA, focus-visible rings) still pass

Closes #1848

### DIFF
--- a/frontend/src/__tests__/SearchResultsLiveRegion.test.ts
+++ b/frontend/src/__tests__/SearchResultsLiveRegion.test.ts
@@ -1,0 +1,29 @@
+/**
+ * Regression for #1848 — search results page must announce completion via
+ * an aria-live region so screen readers know results have loaded (WCAG 4.1.3).
+ */
+import { readFileSync } from "fs";
+import { join } from "path";
+
+const src = readFileSync(
+  join(__dirname, "../app/search/page.tsx"),
+  "utf-8"
+);
+
+describe("Search results live region (closes #1848)", () => {
+  it("SearchResultsInner has an aria-live status region for result announcements", () => {
+    expect(src).toContain('aria-live="polite"');
+  });
+
+  it("status region is aria-atomic so the full count is read at once", () => {
+    expect(src).toContain('aria-atomic="true"');
+  });
+
+  it("status region uses role=status", () => {
+    // Find the aria-live region and confirm it also carries role="status"
+    const idx = src.indexOf('aria-live="polite"');
+    expect(idx).toBeGreaterThan(-1);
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.**

WCAG 4.1.3 fix on the search results page. Persistent live region (`role="status" aria-live="polite" aria-atomic="true"`) announces search outcomes — "Found N results for X", "No results for X", or "Search failed". Matches the pattern already used by vocab page. 9-line net component change + 29-line test. Closes #1848.

Applied `P2`.

PR #1851, opened 2026-04-28
ux: add aria-live filter count announcement to vocabulary page (closes #1850)

### LINKED ISSUE #1850: ux: vocabulary filter missing aria-live result count — screen readers don't announce filtered results (WCAG 4.1.3)
## Problem

When a screen reader user types in the vocabulary search box or selects a tag filter, the results update silently. The empty-state message (`No words match "{search}"`) and tag empty state (`No words tagged with...`) are rendered without any live region, so screen readers cannot announce the change.

The export status uses `role="status" aria-live="polite" aria-atomic="true"` correctly (line 470). The filter results need the same treatment.

## Fix

Reuse or extend the existing `aria-live` region to also announce filter counts:
- After filtering: `"{N} word{plural} found"` (or `"No words match '{search}'"`)
- After tag selection: `"N words tagged with '{tag}'"` (or `"No words tagged with..."`)

## File

`frontend/src/app/vocabulary/page.tsx`

## WCAG

4.1.3 Status Messages (Level AA)

alfmunny:
## What changed

Added a visually-hidden `role="status" aria-live="polite" aria-atomic="true"` element to `vocabulary/page.tsx` that announces the filter result count to screen readers whenever the user types a search term or selects a tag.

- Active filter (search or tag selected) → announces `"N word(s) found."`
- No filter active → element is not rendered (avoids false announcements on page load)

The existing `aria-live` region (export status, line 470) is preserved unchanged.

## Why

When a screen reader user types in the vocabulary search box or selects a tag, the word list silently shrinks or shows an empty state. Without a live region, AT users have no way to know how many results remain after filtering. WCAG 4.1.3 (Status Messages, Level AA) requires dynamically injected filter/result state to be reachable by screen readers.

The `sr-only` class hides the announcement visually to avoid showing a duplicate word count alongside the list.

## Test plan

- `VocabFilterLiveRegion.test.ts` (new) — asserts `sr-only` element with `aria-live="polite"` exists and references `filtered.length`
- All 2647 existing Jest tests pass, including `VocabularyPage.tagfilter` and `VocabularyPage.branches` tests that check the visible empty-state messages

Closes #1850

### DIFF
--- a/frontend/src/__tests__/VocabFilterLiveRegion.test.ts
+++ b/frontend/src/__tests__/VocabFilterLiveRegion.test.ts
@@ -1,0 +1,34 @@
+/**
+ * Regression for #1850 — vocabulary filter must announce result count via
+ * an sr-only aria-live region so screen readers know when filtering changes
+ * the word list (WCAG 4.1.3).
+ */
+import { readFileSync } from "fs";
+import { join } from "path";
+
+const src = readFileSync(
+  join(__dirname, "../app/vocabulary/page.tsx"),
+  "utf-8"
+);
+
+describe("Vocabulary filter live region (closes #1850)", () => {
+  it("has an sr-only aria-live region for filter result announcements", () => {
+    // Must have a visually-hidden live region (not just the export status div)
+    // that announces filter results
+    expect(src).toContain("sr-only");
+    // That sr-only element must be inside or accompanied by aria-live
+    const srIdx = src.indexOf("sr-only");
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.**

WCAG 4.1.3 fix on the vocabulary page filter — visually-hidden `role="status" aria-live="polite"` element announces "N word(s) found" when search/tag filter is active. Conditional render (only when filter active) avoids false announcements on page load. 7-line component change + 34-line test. Closes #1850.

Applied `P2`.

PR #1855, opened 2026-04-28
ux: fix AddWordPicker dialog hidden from screen readers by aria-hidden wrapper (closes #1854)

### LINKED ISSUE #1854: ux: AddWordPicker dialog hidden from screen readers by aria-hidden wrapper
## Problem

In `frontend/src/app/decks/[deckId]/page.tsx`, the `AddWordPicker` component wraps the entire dialog — including the `role="dialog"` element — inside an overlay div with `aria-hidden="true"`:

```tsx
<div
  className="fixed inset-0 z-50 ..."
  onClick={onClose}
  aria-hidden="true"   // ← hides everything inside, including the dialog!
>
  <div
    role="dialog"
    aria-modal="true"
    aria-labelledby="add-word-picker-title"
    ...
  >
```

`aria-hidden="true"` on the outer wrapper makes the inner `role="dialog"` and all its content invisible to screen readers. Screen reader users cannot access the Add Word picker at all.

**WCAG violation:** 4.1.2 Name, Role, Value (Level A) — dialog cannot be found or operated by screen readers.

## Fix

Remove `aria-hidden="true"` from the backdrop overlay div. The backdrop is a click target, not a decorative element — making the whole overlay `aria-hidden` was incorrect. The inner `aria-modal="true"` on the dialog element is the correct way to indicate modal scope to screen readers.

alfmunny:
## What changed

Removed `aria-hidden="true"` from the backdrop overlay div in `AddWordPicker` (deck detail page). The attribute was applied to the outermost container which also wraps the `role="dialog"` element — making the entire modal invisible to screen readers.

## Why

`aria-hidden="true"` on an ancestor hides all descendant content from the accessibility tree. Screen reader users could not find or interact with the Add Word picker at all. The correct scope indicator for modal dialogs is `aria-modal="true"` on the dialog element itself, which was already present.

**WCAG 4.1.2 Name, Role, Value (Level A)**

## Test plan

- [x] New static-source regression test: `AddWordPickerAriaHidden.test.ts` — verifies the backdrop overlay does not carry `aria-hidden="true"` and the dialog has `role="dialog"`, `aria-modal="true"`, and `aria-labelledby`
- [x] Full frontend suite: 2652 tests pass

### DIFF
--- a/frontend/src/__tests__/AddWordPickerAriaHidden.test.ts
+++ b/frontend/src/__tests__/AddWordPickerAriaHidden.test.ts
@@ -1,0 +1,52 @@
+/**
+ * Regression for #1854 — AddWordPicker dialog must not be wrapped in
+ * aria-hidden="true"; doing so hides the dialog from all screen readers
+ * (WCAG 4.1.2).
+ */
+import { readFileSync } from "fs";
+import { join } from "path";
+
+const src = readFileSync(
+  join(__dirname, "../app/decks/[deckId]/page.tsx"),
+  "utf-8"
+);
+
+describe("AddWordPicker dialog accessibility (closes #1854)", () => {
+  it("dialog is not wrapped inside an aria-hidden element", () => {
+    // Find the dialog element
+    const dialogIdx = src.indexOf('role="dialog"');
+    expect(dialogIdx).toBeGreaterThan(-1);
+
+    // Look at the surrounding 500 chars before the dialog for aria-hidden="true"
+    const beforeDialog = src.slice(Math.max(0, dialogIdx - 500), dialogIdx);
+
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.**

Real a11y bug: `aria-hidden="true"` on the backdrop wrapping the AddWordPicker dialog would hide the entire modal from AT — screen reader users couldn't interact with it at all. Fix removes the attribute. The dialog already has `aria-modal="true"` + `role="dialog"` + `aria-labelledby` which are the correct scope indicators. WCAG 4.1.2.

1-line component fix + 52-line regression test that pins down both the absence of `aria-hidden` AND the presence of the correct dialog markers. Closes #1854.

Applied `P2`.

PR #1857, opened 2026-04-28
ux: announce flashcard card position to screen readers after grading (closes #1856)

### LINKED ISSUE #1856: ux: flashcard page card counter and progress not announced to screen readers after grading
## Problem

In `frontend/src/app/vocabulary/flashcards/page.tsx`, the card counter and card content update silently after a card is graded:

```tsx
{/* Card counter — no aria-live, silent after grade */}
<p className="text-center text-xs text-stone-500">
  {currentIndex + 1} of {cards.length} remaining
</p>
```

When a user grades a card (Again/Hard/Good/Easy), the new card appears and the counter updates, but nothing is announced to screen readers. Screen reader users cannot tell which card they're on without manually navigating to the counter.

**WCAG 4.1.3 Status Messages (Level AA)** — dynamically injected status updates that don't receive focus must use `aria-live` or `role="status"`.

## Fix

Wrap the card counter in a `role="status" aria-live="polite"` region so screen readers announce the updated count after each grade. The announcement text should include the current word so users know what's next: e.g. `"Card 3 of 10: Ephemeral"`.

alfmunny:
## What changed

Wrapped the flashcard card counter (`N of M remaining`) in `role="status" aria-live="polite" aria-atomic="true"` and appended the current word to the announcement text.

**Before:** `<p>2 of 10 remaining</p>` — completely silent to screen readers after grading.

**After:** `<p role="status" aria-live="polite" aria-atomic="true">2 of 10 remaining: Ephemeral</p>` — announces position and next word on every card transition.

## Why

WCAG 4.1.3 (Status Messages, Level AA) requires dynamically injected status updates that don't receive focus to use `aria-live` or `role="status"`. Screen reader users grading flashcards received no feedback about where they were in the deck.

## Test plan

- [x] New static-source regression test: `FlashcardCounterLiveRegion.test.ts` — verifies card counter has `role="status"` and `aria-live="polite"` in close proximity
- [x] Full frontend suite: 2652 tests pass

### DIFF
--- a/frontend/src/__tests__/FlashcardCounterLiveRegion.test.ts
+++ b/frontend/src/__tests__/FlashcardCounterLiveRegion.test.ts
@@ -1,0 +1,35 @@
+/**
+ * Regression for #1856 — flashcard page must announce the current card
+ * position and word to screen readers after grading via an aria-live region
+ * (WCAG 4.1.3 Status Messages).
+ */
+import { readFileSync } from "fs";
+import { join } from "path";
+
+const src = readFileSync(
+  join(__dirname, "../app/vocabulary/flashcards/page.tsx"),
+  "utf-8"
+);
+
+describe("Flashcard page card counter live region (closes #1856)", () => {
+  it("has an aria-live region that announces card progress to screen readers", () => {
+    expect(src).toContain('aria-live="polite"');
+  });
+
+  it("card counter or announcement references currentIndex", () => {
+    const liveIdx = src.indexOf('aria-live="polite"');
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.**

WCAG 4.1.3 fix: flashcard card counter `N of M remaining` is now wrapped in `role="status" aria-live="polite" aria-atomic="true"` AND appends the current word to the announcement ("2 of 10 remaining: Ephemeral"). Screen reader users now get position feedback on every card transition. 3-line component change + 35-line test. Closes #1856.

Applied `P2`.

PR #1859, opened 2026-04-28
ux: announce admin books search result count to screen readers (closes #1858)

### LINKED ISSUE #1858: ux: admin books search result count not announced to screen readers
## Problem

In `frontend/src/app/admin/books/page.tsx`, the filter result count (`N / M`) displayed when searching books has no `aria-live` attribute:

```tsx
{searchQuery && (
  <span className="text-xs text-stone-500">
    {filteredCount} / {books.length}
  </span>
)}
```

When an admin types in the search box, the book list and count update silently. Screen readers do not announce how many results match the query.

**WCAG 4.1.3 Status Messages (Level AA)** — dynamically injected status messages must use `aria-live` or `role="status"`.

## Fix

Add `role="status" aria-live="polite" aria-atomic="true"` to the result count span. Also extract the filtered count into a variable to avoid the double filter call.

alfmunny:
## What changed

Added `role="status" aria-live="polite" aria-atomic="true"` to the search result count span in the admin books page.

**Before:** `<span className="text-xs text-stone-500">5 / 12</span>` — silent to screen readers.

**After:** `<span role="status" aria-live="polite" aria-atomic="true" ...>5 / 12</span>` — announces the filtered count as the admin types.

## Why

WCAG 4.1.3 (Status Messages, Level AA): dynamically injected status updates that don't receive focus must use `aria-live` or `role="status"`. Screen reader users typing in the admin books search filter received no feedback about how many results matched.

## Test plan

- [x] New regression test: `AdminBooksSearchLiveRegion.test.ts` — verifies the count span has `aria-live` or `role="status"` in its context, and the input has `aria-label="Filter books"`
- [x] Full frontend suite: 2654 tests pass

### DIFF
--- a/frontend/src/__tests__/AdminBooksSearchLiveRegion.test.ts
+++ b/frontend/src/__tests__/AdminBooksSearchLiveRegion.test.ts
@@ -1,0 +1,25 @@
+/**
+ * Regression for #1858 — admin books search result count must be announced
+ * to screen readers via aria-live or role="status" (WCAG 4.1.3).
+ */
+import { readFileSync } from "fs";
+import { join } from "path";
+
+const src = readFileSync(
+  join(__dirname, "../app/admin/books/page.tsx"),
+  "utf-8"
+);
+
+describe("Admin books search live region (closes #1858)", () => {
+  it("result count span has role=status or aria-live", () => {
+    // The visible count (e.g. "5 / 10") when searching must be in a live region
+    const countIdx = src.indexOf("books.length");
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.**

Same WCAG 4.1.3 pattern. `role="status" aria-live="polite"` on the admin books filter count span. 1-line component change + 25-line test. Closes #1858.

Applied `P3` (admin-only surface — narrower impact than user-facing pages got).

PR #1861, opened 2026-04-28
ux: announce home page book discovery search results to screen readers (closes #1860)

### LINKED ISSUE #1860: ux: home page book discovery search results not announced to screen readers
## Problem

In `frontend/src/app/page.tsx`, the "Discover Books" Gutenberg search shows results in a grid after the user clicks Search, but there is no `aria-live` announcement when results appear. Screen readers don't know how many books were found.

Current state: only the loading skeleton has `role="status" aria-label="Loading search results"`. When results replace the skeleton, screen readers receive no announcement.

Compare to `/search/page.tsx` which has a proper `role="status" aria-live="polite"` live region (added in #1848).

**WCAG 4.1.3 Status Messages (Level AA)**

## Fix

Add an `sr-only` `aria-live="polite"` region that announces the result count once the search completes:
- Results found: `"Found N books for [query]."`  
- No results: the existing visible message is in a non-live div, so a sr-only companion is needed  
- Error: already covered by `role="alert"`

alfmunny:
## What changed

Added a visually-hidden `role="status" aria-live="polite" aria-atomic="true"` region to the "Discover Books" search section on the home page. It appears after the search completes (not during loading) and announces:

- **Results found:** `"Found N book(s) for <query>."`  
- **No results:** `"Search complete. No results for <query>."`  
- **Error:** already covered by the existing `role="alert"` div

## Why

WCAG 4.1.3 (Status Messages, Level AA): when search results load dynamically without focus moving, screen readers must be notified via `aria-live`. The Discover Books search had a loading skeleton with `role="status"` but no announcement when the actual results appeared. Compare to `/search/page.tsx` which gained this fix in #1848.

## Test plan

- [x] New regression test: `HomeSearchLiveRegion.test.ts` — verifies `sr-only` element with `aria-live` and `searchResults.length` reference exist in the page source
- [x] Existing `HomePage.test.tsx` still passes (no-results text duplication avoided by using distinct phrasing in the live region)
- [x] Full frontend suite: 2656 tests pass

### DIFF
--- a/frontend/src/__tests__/HomeSearchLiveRegion.test.ts
+++ b/frontend/src/__tests__/HomeSearchLiveRegion.test.ts
@@ -1,0 +1,25 @@
+/**
+ * Regression for #1860 — home page book discovery search results must
+ * announce the result count to screen readers via an aria-live region
+ * (WCAG 4.1.3 Status Messages).
+ */
+import { readFileSync } from "fs";
+import { join } from "path";
+
+const src = readFileSync(join(__dirname, "../app/page.tsx"), "utf-8");
+
+describe("Home page discovery search live region (closes #1860)", () => {
+  it("has an sr-only aria-live region for search result announcements", () => {
+    expect(src).toContain("sr-only");
+    const srIdx = src.indexOf("sr-only");
+    const window = src.slice(Math.max(0, srIdx - 300), srIdx + 300);
+    expect(window).toContain("aria-live");
+  });
+
+  it("live region references search result count", () => {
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.**

Same WCAG 4.1.3 pattern. `sr-only` live region for the home-page Discover Books search. Announces "Found N book(s) for X" / "Search complete. No results for X". Distinct phrasing avoids duplication with the visible no-results text. 9-line component change + 25-line test. Closes #1860.

Applied `P2` (user-facing home page).

PR #1868, opened 2026-04-28
ux: use alt='' on BookCard cover image to prevent double title announcement (closes #1867)

### LINKED ISSUE #1867: ux: BookCard cover image alt text duplicates button label (WCAG 1.1.1)
## Problem

In `frontend/src/components/BookCard.tsx`, the cover image has `alt={book.title}`:

```tsx
<img
  src={book.cover}
  alt={book.title}     // ← announces title to screen reader
  className="w-full h-40 object-cover rounded-lg mb-2"
/>
```

The image sits inside a `<button>` whose accessible name is already derived from the two `<p>` children below it:

```tsx
<p className="font-serif font-semibold text-sm text-ink line-clamp-2 flex-1">
  {book.title}         // ← announces title again
</p>
<p className="text-xs text-amber-700 mt-1 line-clamp-1">
  {book.authors.join(", ")}
</p>
```

Result: screen readers announce the book title **twice** per card — once for the image alt text and once for the visible heading. With 12+ cards on screen this creates significant noise.

## Fix

Change `alt={book.title}` → `alt=""` on the cover image. The image is purely decorative inside the button; the adjacent text already provides the accessible name.

## WCAG

1.1.1 (Non-text Content) — decorative images inside interactive elements should have empty alt so they're ignored by AT, not duplicating the element's accessible name.

alfmunny:
## What changed

Changed `alt={book.title}` → `alt=""` on the cover `` in `BookCard.tsx`. Updated the existing cover-image test to query by DOM element rather than `getByRole("img", { name })` (which no longer matches a presentational image).

## Why

The cover image sits inside a `<button>` whose accessible name is already computed from the two visible `<p>` elements below it (title + authors). With `alt={book.title}`, screen readers announced the title **twice** per card — once from the image alt text, once from the paragraph. On a library grid with 12+ cards this creates significant redundancy and AT noise.

Setting `alt=""` marks the image as presentational (WCAG 1.1.1 — decorative images inside interactive elements must use empty alt so AT ignores them rather than duplicating the element's accessible name).

Closes #1867

## Test plan

- `BookCardCoverAlt.test.ts` (new): verifies cover img has `alt=""` and does NOT have `alt={book.title}`
- `BookCard.test.tsx` (updated): cover-image assertion updated to use `container.querySelector("img")` and checks `alt=""` directly
- Full frontend suite: 2663 tests pass

### DIFF
--- a/frontend/src/__tests__/BookCard.test.tsx
+++ b/frontend/src/__tests__/BookCard.test.tsx
@@ -29,8 +29,13 @@ test("renders author name", () => {
 
 test("renders cover image when cover URL is provided", () => {
   render(<BookCard book={BOOK} onClick={jest.fn()} />);
-  const img = screen.getByRole("img", { name: "Pride and Prejudice" });
+  // Cover image uses alt="" (decorative — button label comes from the <p> text)
+  // eslint-disable-next-line testing-library/no-container
+  const { container } = render(<BookCard book={BOOK} onClick={jest.fn()} />);
+  const img = container.querySelector("img");
+  expect(img).not.toBeNull();
   expect(img).toHaveAttribute("src", BOOK.cover);
+  expect(img).toHaveAttribute("alt", "");
 });
 
 test("renders SVG placeholder when no cover URL", () => {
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM review.**

WCAG 1.1.1 — cover image inside an already-named `<button>` should be presentational (`alt=""`). Was double-announcing book title for every card on a library grid. Updated the existing test to query by DOM since the image is no longer role=img. Closes #1867. 1-line component change + 27-line new test + 6-line update to existing test.

Applied `P2`.

PR #1870, opened 2026-04-28
ux: add aria-live region for Save preferences confirmation (WCAG 4.1.3, closes #1869)

### LINKED ISSUE #1869: ux: profile page Save preferences success has no screen-reader announcement (WCAG 4.1.3)
## Problem

In `frontend/src/app/profile/page.tsx`, the "Save preferences" button changes its label to "Saved!" on success:

```tsx
<button onClick={savePreferences}>
  {prefsSaved ? "Saved!" : "Save preferences"}
</button>
```

Dynamic changes to a button's accessible name (its text content) are not reliably announced by screen readers — they fire no ARIA live region event. VoiceOver, Narrator, and NVDA all handle button-label mutations inconsistently. Users who can't see the green background colour change have no reliable way to confirm the save succeeded.

By contrast, the same page already uses `role="status"` live regions for API key and Obsidian save feedback — the preferences save should follow the same pattern.

## Fix

Add a visually-hidden `role="status"` live region adjacent to the button that announces "Preferences saved." when `prefsSaved` is true:

```tsx
  {prefsSaved ? "Preferences saved." : ""}
</div>
```

## WCAG

4.1.3 (Status Messages) — success confirmations injected into the DOM must be announced without receiving focus.

alfmunny:
## What changed
Added a `role="status" aria-live="polite" aria-atomic="true"` live region (visually hidden via `sr-only`) directly above the Save preferences button on the profile page. When `prefsSaved` flips to `true`, the region announces "Preferences saved." to screen readers.

## Why
WCAG 4.1.3 (Status Messages) requires that dynamically injected success/status messages be communicated to AT without requiring focus. The prior implementation only changed the button label to "Saved!" — screen readers don't reliably announce button label mutations, so users relying on AT received no confirmation that their preferences were saved.

## Test plan
- New regression test `ProfilePrefsSavedAnnouncement.test.ts` (static source analysis):
  - Asserts a `role="status"` or `aria-live` region exists within 600 chars of the last `prefsSaved` JSX reference
  - Asserts the live region content is conditional on `prefsSaved`
- Full frontend suite: 420 suites, 2665 tests passing
- Full backend suite: 2016 tests passing

Closes #1869

### DIFF
--- a/frontend/src/__tests__/ProfilePrefsSavedAnnouncement.test.ts
+++ b/frontend/src/__tests__/ProfilePrefsSavedAnnouncement.test.ts
@@ -1,0 +1,34 @@
+/**
+ * Regression test for #1869 — profile page "Save preferences" button changed
+ * its label to "Saved!" but had no aria-live region. Screen readers don't
+ * reliably announce dynamic button label changes, so a separate role="status"
+ * live region is needed next to the button.
+ */
+import { readFileSync } from "fs";
+import { join } from "path";
+
+const src = readFileSync(
+  join(process.cwd(), "src/app/profile/page.tsx"),
+  "utf-8"
+);
+
+describe("Profile page preferences save feedback (WCAG 4.1.3)", () => {
+  it("has an aria-live region near the Save preferences button", () => {
+    // Find the save button area (uses prefsSaved in its className)
+    const idx = src.lastIndexOf("prefsSaved");
+    expect(idx).toBeGreaterThan(-1);
+    // Within a reasonable window around the last prefsSaved reference (near button), there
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.**

WCAG 4.1.3 fix on profile Save preferences. The button-label-mutation-only approach ("Saved!") wasn't reliable for AT — screen readers don't consistently announce label changes. Now an `sr-only` `role="status" aria-live="polite"` region announces "Preferences saved" when `prefsSaved` flips. 3-line component change + 34-line test. Closes #1869.

Applied `P2`.

PR #1874, opened 2026-04-28
ux: add sr-only text to component skeleton live regions (WCAG 4.1.3, closes #1873)

alfmunny:
## What changed

Follow-up to #1872 — extends sr-only accessible text to 7 component files that were missed in the first pass:

- `ReadingStats.tsx` — loading stats skeleton
- `AnnotationsSidebar.tsx` — loading annotations + refreshing annotations spinners
- `ChapterSummary.tsx` — loading summary skeleton
- `InsightChat.tsx` — loading messages skeleton + AI is typing indicator
- `QueueTab.tsx` — loading queue skeleton
- `SentenceReader.tsx` — two loading translation skeletons
- `TranslationView.tsx` — two loading translation skeletons

Each `role="status" aria-label="..."` container now has `<span className="sr-only">...</span>` as its first child so screen readers announce the loading state on mount (WCAG 4.1.3).

## Why

ARIA live regions (`role="status"`) announce **text content changes**, not `aria-label`. Containers containing only animated skeleton `<div>`s are completely silent to assistive technology.

## Test plan

- Extended `SkeletonSrOnlyText.test.ts` to cover all 7 component files (24 total)
- Updated `QueueTab.coverage.test.tsx` to use `getAllByText` since the QueueTab skeleton already had visible "Loading queue…" text — the sr-only span now causes two matches
- All 2689 frontend tests pass
- All 2016 backend tests pass

Closes #1873

PR #1872, opened 2026-04-28
ux: add sr-only text to skeleton live regions so AT announces loading state (WCAG 4.1.3, closes #1871)

alfmunny:
## What changed

Added `<span className="sr-only">Loading [context]...</span>` as the first child of every `role="status" aria-label="..."` skeleton/spinner container across 15 files (18 instances total).

## Why

ARIA live regions (`role="status"`) announce **text content** changes to assistive technology — they do NOT announce `aria-label`. Every skeleton loader in the app used `role="status" aria-label="Loading X"` with only animated `<div>` children inside, meaning screen readers received no announcement when loading began. Users relying on AT had no way to know the page was fetching data.

The fix is to add a visually-hidden `sr-only` span whose text matches the `aria-label`, giving the live region actual text content to announce when the element mounts.

**Files fixed:**
- `app/loading.tsx`, `app/page.tsx`, `app/search/page.tsx`
- `app/admin/layout.tsx`, `app/admin/audio/page.tsx`, `app/admin/books/page.tsx`, `app/admin/users/page.tsx`, `app/admin/uploads/page.tsx`
- `app/reader/[bookId]/page.tsx` (chapter + annotations skeletons)
- `app/decks/page.tsx`, `app/decks/[deckId]/page.tsx`
- `app/notes/page.tsx`, `app/notes/[bookId]/page.tsx`
- `app/vocabulary/flashcards/page.tsx`, `app/vocabulary/page.tsx`
- `app/upload/page.tsx`, `app/upload/[bookId]/chapters/page.tsx`

## Test plan

New `SkeletonSrOnlyText.test.ts` (17 tests, static source analysis): for each of the 17 affected files, asserts that every `role="status" aria-label="..."` container contains either `sr-only` text or visible paragraph/span content. Regression guard prevents future skeleton containers from missing accessible text.

- Frontend: 421 suites, 2682 tests passing
- Backend: 2016 tests passing

Closes #1871

PR #1876, opened 2026-04-28
ux: make API-key and Obsidian save confirmations always-present live regions (WCAG 4.1.3, closes #1875)

### LINKED ISSUE #1875: ux: profile page API-key and Obsidian save confirmations not announced by screen readers (WCAG 4.1.3)
## Problem

The Gemini API-key save confirmation (line 340) and the Obsidian settings save confirmation (line 445) in `src/app/profile/page.tsx` use a conditionally-mounted `role="status"` element:

```tsx
{keyMessage && (
  <p role="status" className="...">{keyMessage.text}</p>
)}
// and
{obsidianMsg && (
  <p role="status" className="...">{obsidianMsg.text}</p>
)}
```

Because the element is not in the DOM before the message appears, screen readers never observe a live-region text change — they just see a new element mount. Most AT will not announce it. Only sighted users see the feedback.

This is the same anti-pattern fixed for the Save-preferences button in #1869 (PR #1870).

## Fix

Keep the `role="status"` container always present in the DOM (empty string when no message), so the live region is registered before content changes:

```tsx
<p role="status" aria-live="polite" aria-atomic="true" className="mt-3 text-sm ...">
  {keyMessage?.text ?? ''}
</p>
```

Apply the same pattern to the Obsidian message.

## Impact

Screen reader users receive no feedback when saving API key or Obsidian settings — they cannot tell if the action succeeded or failed without inspecting the page manually.

## References
- Closes as part of the WCAG 4.1.3 sweep: #1869, #1871, #1873
- WCAG 4.1.3 Status Messages (Level AA)

alfmunny:
## What changed

Fixed two dynamically-mounted `role="status"` feedback paragraphs in the profile page that screen readers never announced:

- **Gemini API key save confirmation** (was `{keyMessage && <p role="status">...`)
- **Obsidian settings save confirmation** (was `{obsidianMsg && <p role="status">...`)

Both are now always present in the DOM with empty content when idle, using `msg?.text ?? ""`. Added `aria-live="polite" aria-atomic="true"` explicitly for clarity.

## Why

ARIA live regions announce **text-content changes**, not new element mounts. A conditionally-rendered `role="status"` element is invisible to AT until it exists — by which time it's too late for the announcement. Same root cause as #1869.

## Test plan

- Updated `ProfileObsidianFeedbackRole.test.ts`:
  - Asserts `keyMessage &&` and `obsidianMsg &&` patterns are GONE (prevents regression to conditional mount)
  - Asserts always-present `role="status"` + `aria-live` exists near `keyMessage?.text` and `obsidianMsg?.text`
- All 2689 frontend tests pass

Closes #1875

### DIFF
--- a/frontend/src/__tests__/ProfileObsidianFeedbackRole.test.ts
+++ b/frontend/src/__tests__/ProfileObsidianFeedbackRole.test.ts
@@ -1,6 +1,9 @@
 /**
- * Regression tests for #1255: profile page feedback messages and reader
- * obsidian toast must have live-region roles so screen readers announce them.
+ * Regression tests for #1255 and #1875:
+ * - Profile page feedback messages must have live-region roles so screen readers announce them.
+ * - #1875: keyMessage and obsidianMsg containers must be always-present in the DOM (not
+ *   conditionally mounted) so AT observes a text-content change, not a new element mount.
+ *   Always-present pattern: role="status" aria-live="polite" with text={msg?.text ?? ""}.
  */
 import * as fs from "fs";
 import * as path from "path";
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.**

Same root cause as #1869, #1872, #1874: live regions need to be **mounted** to announce text changes. Conditionally-rendered `role="status"` elements were silent until DOM-mounted, by which time it was too late. Fix makes both Gemini API key + Obsidian save confirmations always-present (`msg?.text ?? ""`). Updated test now asserts the conditional-mount pattern is gone (prevents regression). 6-line component change + 19-line test rewrite. Closes #1875.

Applied `P2`.

PR #1879, opened 2026-04-28
ux: add focus management to reader mobile chat sheet (WCAG 2.4.3, closes #1877)

alfmunny:
## What changed

Fixed missing focus management on the reader page's mobile chat sheet (`role="dialog"`).

Changes in `src/app/reader/[bookId]/page.tsx`:
- Added `chatSheetRef = useRef<HTMLDivElement>(null)` for the dialog element
- Added `ref={chatSheetRef} tabIndex={-1} focus:outline-none` to the dialog div
- Added `useEffect` that moves focus into the sheet when it opens and restores focus to the previous element on close
- Added `useEffect` that registers an `Escape` keydown listener while the sheet is open (dismisses both `sidebarOpen` and `chatSheetText`)

## Why

The mobile chat sheet had `role="dialog" aria-modal="true"` but no focus management — all four WAI-ARIA dialog pattern requirements were absent:

1. No `ref` on the dialog element
2. No focus shift into the dialog on open (screen readers don't know context changed)
3. No `Escape` key handler
4. No focus restoration on close

VoiceOver/TalkBack users opening the chat sheet kept reading content behind the modal backdrop, and had no keyboard path to dismiss it. Same fix pattern as `AnnotationsSidebar.tsx` which is fully compliant.

## Test plan

- Extended `ReaderAriaExpandedDialog.test.ts` with 4 new assertions:
  - `chatSheetRef` in the dialog's `ref` attribute
  - `tabIndex={-1}` on the dialog div
  - `chatSheetRef.current?.focus()` call exists
  - `Escape` handler calls both `setSidebarOpen` and `setChatSheetText`
- All 2693 frontend tests pass

Closes #1877

PR #1882, opened 2026-04-28
ux: restore focus to trigger on AnnotationToolbar close (WCAG 2.4.3)

### LINKED ISSUE #1878: ux: AnnotationToolbar dialog does not restore focus on close (WCAG 2.4.3)
## Problem

`src/components/AnnotationToolbar.tsx` opens as a `role="dialog"` panel and correctly moves focus to the textarea on mount, but never restores focus to the trigger element when the dialog closes (via Save, Cancel, or Escape).

After dismissal, keyboard focus is stranded — users must Tab from the beginning of the page.

```tsx
// Current (no restoration):
useEffect(() => {
  textareaRef.current?.focus();
}, []);
```

## Fix

Capture the active element before moving focus, and restore it in cleanup:

```tsx
useEffect(() => {
  textareaRef.current?.focus();
}, []);
```

## References
- WCAG 2.4.3 Focus Order (Level AA)
- WAI-ARIA Authoring Practices Guide — Dialog Pattern: "When the dialog closes, focus should return to the element that invoked the dialog."
- Reference implementation: `AnnotationsSidebar.tsx` (already correct)

alfmunny:
## What changed
Fixes `AnnotationToolbar` to restore focus to the trigger element on unmount, satisfying WCAG 2.4.3 (Focus Order).

The `useEffect` that moved focus into the textarea on open now also captures `document.activeElement` before focusing and restores it in the cleanup function.

## Why
When a modal dialog closes, keyboard and AT users must return to the element that opened it. Without restoration, focus drops to `document.body` — the user loses their place in the reading flow.

## Test plan
- Extended `frontend/src/__tests__/AnnotationToolbarDialogRole.test.tsx` with an RTL test that:
  1. Focuses a synthetic trigger button
  2. Mounts `AnnotationToolbar` (focus moves to textarea)
  3. Unmounts
  4. Asserts `document.activeElement === trigger`
- All 2692 frontend tests pass

Closes #1878

### DIFF
--- a/frontend/src/__tests__/AnnotationToolbarDialogRole.test.tsx
+++ b/frontend/src/__tests__/AnnotationToolbarDialogRole.test.tsx
@@ -1,6 +1,7 @@
 /**
  * Regression test for #1222: AnnotationToolbar panel must have role="dialog",
  * aria-modal="true", and aria-labelledby so screen readers announce it correctly.
+ * #1878: dialog must restore focus to the trigger element on unmount (WCAG 2.4.3).
  */
 import React from "react";
 import { render, screen } from "@testing-library/react";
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.**

WCAG 2.4.3 fix on AnnotationToolbar — captures `document.activeElement` on open, restores on unmount. Without this, closing the toolbar drops focus to body and the user loses their place. 2-line component change + 20-line RTL test that exercises the full mount/unmount cycle. Closes #1878.

Same pattern as #1879's reader chat sheet. Applied `P2`.

PR #1885, opened 2026-04-28
ux: restore focus to trigger on VocabWordTooltip close (WCAG 2.4.3)

### LINKED ISSUE #1881: ux: VocabWordTooltip dialog does not restore focus on close (WCAG 2.4.3)
## Problem

`src/components/VocabWordTooltip.tsx` opens as `role="dialog" aria-modal="true"` and correctly moves focus to itself on mount, but the `useEffect` does not capture or restore the previously-focused element:

```tsx
useEffect(() => {
  ref.current?.focus();
}, []);
```

After the tooltip is dismissed, focus is stranded (not returned to the word that was clicked/selected).

## Fix

```tsx
useEffect(() => {
  ref.current?.focus();
}, []);
```

## References
- WCAG 2.4.3 Focus Order (Level AA)
- Related: #1878 (AnnotationToolbar), #1880 (AddWordPicker), #1877 (reader chat sheet) — same pattern

alfmunny:
## What changed
Fixes `VocabWordTooltip` to restore focus to the trigger element on unmount, satisfying WCAG 2.4.3 (Focus Order).

The `useEffect` that moved focus into the dialog on open now also captures `document.activeElement` before focusing and restores it in the cleanup function.

## Why
When a tooltip/popover dialog closes, keyboard and AT users must return to the element that opened it. Without restoration, focus drops to `document.body` — users lose their position in the reading text.

## Test plan
- Extended `frontend/src/__tests__/VocabWordTooltipDialog.test.ts` with an assertion verifying the `prev?.focus?.()` cleanup pattern is present
- All 2695 frontend tests pass

Closes #1881

### DIFF
--- a/frontend/src/__tests__/VocabWordTooltipDialog.test.ts
+++ b/frontend/src/__tests__/VocabWordTooltipDialog.test.ts
@@ -1,6 +1,7 @@
 /**
  * Regression test for #1315: VocabWordTooltip must be announced to screen
  * readers as a dialog with an accessible name and receive focus on mount.
+ * #1881: dialog must restore focus to the trigger on unmount (WCAG 2.4.3).
  */
 import * as fs from "fs";
 import * as path from "path";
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.**

Same WCAG 2.4.3 pattern as #1882/#1883, this time on VocabWordTooltip. `prev` capture + cleanup-time `prev?.focus?.()`. 3-line component change + 12-line test addition. Closes #1881.

Applied `P2`.

PR #1887, opened 2026-04-28
ux: add aria-label to Gutenberg icon-link in reader header (WCAG 4.1.2)

### LINKED ISSUE #1884: ux: reader Gutenberg link has no reliable accessible name (WCAG 4.1.2)
## Problem

In `src/app/reader/[bookId]/page.tsx`, the "View on Project Gutenberg" link near the book title uses only a `title` attribute for its accessible name:

```tsx
<a href={...} title="View on Project Gutenberg">
  <ArrowUpRightIcon className="w-3 h-3" aria-hidden="true" />
</a>
```

The icon has `aria-hidden="true"`, so AT sees no text content. While `title` can serve as an accessible name, it is unreliable across AT/browser combinations and is not announced by some screen readers (especially in mobile mode or with certain NVDA/VoiceOver profiles).

**WCAG 4.1.2 (Name, Role, Value):** Interactive elements must have a reliable accessible name. `aria-label` is the correct mechanism for icon-only links.

## Fix

Add `aria-label="View on Project Gutenberg"` to the anchor element, keeping the `title` as a visible tooltip.

## Reproduction
- Open any book in the reader
- Tab to the Gutenberg link icon in the header
- Screen reader announces only the title tooltip (unreliable) rather than a defined accessible name

alfmunny:
## What changed
Added `aria-label="View on Project Gutenberg"` to the icon-only anchor link in the reader header that links to Project Gutenberg.

## Why
The link rendered only an SVG icon (`ArrowUpRightIcon`) with `aria-hidden="true"`, so screen readers had no accessible name for the link. This violated WCAG 4.1.2 (Name, Role, Value). The `title` attribute alone is unreliable across AT/browser combinations and does not provide a programmatic accessible name.

## Test plan
- Added `frontend/src/__tests__/ReaderGutenbergLinkAriaLabel.test.ts` with 2 static assertions:
  - `aria-label` containing "View on Project Gutenberg" is present in the source file
  - `aria-label` containing "Gutenberg" appears near the link
- All 2700 frontend tests pass

Closes #1884

## Docs follow-up
- Design change logged in `docs/design-improvement-plan.md` Change Log.

### DIFF
--- a/frontend/src/__tests__/ReaderGutenbergLinkAriaLabel.test.ts
+++ b/frontend/src/__tests__/ReaderGutenbergLinkAriaLabel.test.ts
@@ -1,0 +1,30 @@
+/**
+ * Regression test for #1884: icon-only "View on Project Gutenberg" link in the
+ * reader header must have aria-label (WCAG 4.1.2 — Name, Role, Value).
+ *
+ * The link icon has aria-hidden="true", so title= alone is unreliable across
+ * AT/browser combinations. aria-label is the correct mechanism.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../app/reader/[bookId]/page.tsx"),
+  "utf8"
+);
+
+describe("Reader Gutenberg link accessible name (closes #1884)", () => {
+  it("Gutenberg external link has aria-label (not title only)", () => {
+    const idx = src.indexOf("View on Project Gutenberg");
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.**

WCAG 4.1.2 fix on the icon-only Gutenberg link in the reader header. `aria-label="View on Project Gutenberg"` so AT has an accessible name for what was just an SVG with `aria-hidden`. 1-line component change + 30-line test. Closes #1884.

Applied `P2`.

PR #1889, opened 2026-04-28
ux: make notes export confirmation an always-present live region (WCAG 4.1.3)

### LINKED ISSUE #1886: ux: notes export confirmation not announced to screen readers (WCAG 4.1.3)
## Problem

In `src/app/notes/[bookId]/page.tsx`, the Obsidian export confirmation message is conditionally mounted with no live region:

```tsx
{exportMsg && (
  <div className="...">
    <p className="text-xs text-amber-800 ...">{exportMsg}</p>
  </div>
)}
```

When a user clicks "Export" and the message appears, AT users will not hear the confirmation because:
- No `role="status"` or `role="alert"` on the container
- No `aria-live` attribute — dynamic insertion is silent to AT

**WCAG 4.1.3 (Status Messages):** Export success/failure feedback must be conveyed through an accessible live region.

## Fix

Replace with an always-present live region:
```tsx
<p role="status" aria-live="polite" aria-atomic="true" className={`text-xs truncate px-3 py-1.5 ... ${exportMsg ? "bg-amber-50 border border-amber-200 text-amber-800 rounded" : ""}`}>
  {exportMsg ?? ""}
</p>
```

This matches the pattern already used in `profile/page.tsx` for API key save confirmations (PR #1876).

## Reproduction
- Open a book notes page with annotations
- Have an Obsidian export token configured
- Click "Export"
- Screen reader does not announce the success/failure message

alfmunny:
## What changed
Changed the notes page export confirmation message from a conditionally-mounted element to an always-present live region using the `exportMsg ?? ""` pattern.

**Before:**
```tsx
{exportMsg && (
  <div className="…">{exportMsg}</div>
)}
```

**After:**
```tsx
<p
  role="status"
  aria-live="polite"
  aria-atomic="true"
  className={`… ${exportMsg ? "bg-amber-50 border border-amber-200 text-amber-800" : ""}`}
>
  {exportMsg ?? ""}
</p>
```

## Why
When `exportMsg` was conditionally mounted, the `role="status"` element didn't exist before the message appeared. Screen readers register live regions at parse time — a newly inserted `role="status"` node is unreliable for AT announcement. The always-present pattern (empty string when no message) ensures AT has the live region registered before any content is placed in it. WCAG 4.1.3 (Status Messages).

## Test plan
- Added `frontend/src/__tests__/NotesExportLiveRegion.test.ts` with 3 static assertions:
  - Always-present pattern (`exportMsg ?? ""`) is used instead of conditional mount
  - `role="status"` or `role="alert"` is present near the expression
  - `aria-live` is present near the expression
- All 2702 frontend tests pass

Closes #1886

### DIFF
--- a/frontend/src/__tests__/NotesExportLiveRegion.test.ts
+++ b/frontend/src/__tests__/NotesExportLiveRegion.test.ts
@@ -1,0 +1,39 @@
+/**
+ * Regression test for #1886: notes export confirmation must be announced via
+ * an accessible live region (WCAG 4.1.3 — Status Messages).
+ *
+ * The conditional {exportMsg && <p>...} pattern is silent to AT because
+ * dynamic DOM insertion without aria-live is not announced. Fix: always-present
+ * container using the role="status" + aria-live="polite" pattern.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../app/notes/[bookId]/page.tsx"),
+  "utf8"
+);
+
+describe("Notes export confirmation live region (closes #1886)", () => {
+  it("export message uses always-present live region (not conditional mount)", () => {
+    // The broken pattern: {exportMsg && <element>}
+    // After the fix, exportMsg should be used as content, not as a render gate
+    // Look for the always-present pattern: exportMsg ?? "" or exportMsg?.
+    expect(src).toMatch(/exportMsg\s*\?\?|\bexportMsg\?\.text/);
+  });
+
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.**

Same WCAG 4.1.3 pattern as #1876 (API-key/Obsidian) and #1869 — conditional mount on `role="status"` is unreliable; always-present element with `exportMsg ?? ""` ensures AT registers the live region before content arrives. 10-line component change + 39-line test. Closes #1886.

Applied `P2`.

PR #1896, opened 2026-04-28
ux: replace blocking alert()/confirm() with inline accessible feedback in admin/users (closes #1895)

### LINKED ISSUE #1895: ux: admin pages use blocking alert()/confirm() dialogs — replace with accessible inline feedback (WCAG 3.3.1)
## Problem

The admin section uses native \`alert()\` and \`confirm()\` for feedback and confirmation prompts. These:

- Block the browser's JavaScript thread (breaks screen reader announcement queue)
- Cannot be styled or made accessible (no ARIA roles, no focus trap from the app's perspective)
- Dismiss all open dialogs and modals in assistive technology
- Some browsers and extensions suppress them entirely (silent failures for users)

**Files affected:**
- \`admin/users/page.tsx\` — 1× \`alert()\` (error feedback), 1× \`confirm()\` (delete)
- \`admin/audio/page.tsx\` — 1× \`alert()\`
- \`admin/books/page.tsx\` — 10+ \`alert()\`/\`confirm()\`
- \`components/QueueTab.tsx\` — 8+ \`alert()\`/\`confirm()\`

## Fix

Replace:
- \`alert(msg)\` → inline \`role="alert"\` banner / toast state
- \`confirm(msg)\` → two-step button (click once shows inline "Are you sure? / Yes / No" replacing the action button)

## This issue

Scoped to \`admin/users/page.tsx\` first (smallest scope, 2 calls). Subsequent issues will cover audio, books, and QueueTab.

## Acceptance

- \`alert()\` calls → set an error state rendered as \`<div role="alert">\`
- \`confirm()\` calls → inline two-step confirmation button (no modal needed for admin)
- Tests updated to assert on DOM elements instead of \`window.alert\` mock

alfmunny:
## What changed

Replaced blocking `alert()` and `confirm()` dialogs in `admin/users/page.tsx` with fully accessible inline patterns:

- **Error feedback**: `act()` now catches errors and stores them in `actError` state, rendered as a dismissible `<div role="alert">` banner instead of `alert()`
- **Delete confirmation**: Two-step inline confirmation — clicking Delete swaps the button for "Delete? / Yes / No" inline controls instead of `confirm()`
- **SpinnerRow**: Added `<span class="sr-only">Loading users...</span>` and `aria-hidden` on the spinner div for screen reader correctness

## Why

`alert()` and `confirm()` break the screen-reader announcement queue, are unstyled (inconsistent UX), and can be suppressed by browser/OS settings. WCAG 3.3.1 requires that errors are identified to users in text — blocking dialogs do not satisfy this reliably. The inline two-step pattern is accessible, keyboard-navigable, and dismissible.

## Test plan

- Added `AdminUsersNoBlockingDialogs.test.ts` — 4 static assertions: no `alert(`, no `confirm(`, `role="alert"` present, `pendingDelete` state present
- Updated `AdminUsersPage.test.tsx` — replaced 3 tests that mocked `window.confirm`/`window.alert` with tests using the two-step button interaction (find "Confirm delete", find "Cancel delete", find `role="alert"` with "Failed")
- Full frontend suite: 2737 tests passed
- Full backend suite: 1942 tests passed

Closes #1895

## Docs follow-up

- [ ] Docs updated (if applicable) — design-improvement-plan.md Change Log updated

### DIFF
--- a/frontend/src/__tests__/AdminUsersNoBlockingDialogs.test.ts
+++ b/frontend/src/__tests__/AdminUsersNoBlockingDialogs.test.ts
@@ -1,0 +1,29 @@
+/**
+ * Regression test for issue #1895 — admin/users must not use blocking
+ * alert() or confirm() dialogs (WCAG 3.3.1).
+ */
+import { readFileSync } from "fs";
+import { join } from "path";
+
+const src = readFileSync(
+  join(__dirname, "../app/admin/users/page.tsx"),
+  "utf8",
+);
+
+describe("admin/users — no blocking dialogs (#1895)", () => {
+  it("does not call alert()", () => {
+    expect(src).not.toMatch(/\balert\s*\(/);
+  });
+
+  it("does not call confirm()", () => {
+    expect(src).not.toMatch(/\bconfirm\s*\(/);
+  });
+
+  it("shows action errors via a role=alert element", () => {
+    expect(src).toMatch(/role="alert"/);
+  });
+
+  it("uses two-step delete confirmation (pendingDelete state)", () => {
+    expect(src).toMatch(/pendingDelete/);
+  });
+});
--- a/frontend/src/__tests__/AdminUsersPage.test.tsx
+++ b/frontend/src/__tests__/AdminUsersPage.test.tsx
@@ -145,8 +145,7 @@ describe("AdminUsersPage", () => {
     });
   });
 
-  it("calls delete endpoint when Del is confirmed", async () => {
-    jest.spyOn(window, "confirm").mockReturnValue(true);
+  it("calls delete endpoint after two-step confirmation", async () => {
     mockAdminFetch
       .mockResolvedValueOnce(SAMPLE_USERS)
       .mockResolvedValueOnce({})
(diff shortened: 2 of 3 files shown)


alfmunny:
**PM review.**

Substantive UX + a11y fix. Replaces blocking `alert()`/`confirm()` in admin/users with inline accessible patterns:
- **Error**: dismissible `<div role="alert">` banner instead of `alert()`
- **Delete confirm**: two-step inline "Delete? / Yes / No" instead of `confirm()`
- **Spinner**: sr-only loading text + aria-hidden on visual spinner

Native dialogs break the screen-reader queue, can be suppressed, and aren't styled consistently. The replacement is keyboard-navigable, AT-friendly, and visually consistent with the rest of the admin UI. 47-line net component change + new `AdminUsersNoBlockingDialogs.test.ts` regression test (asserts no `alert(` / `confirm(` reappear). Existing AdminUsersPage tests rewritten for the two-step pattern.

Applied `P2`. Worth noting: this is a **pattern fix** — once landed, would be a good idea for UI/UX to grep for other `alert(`/`confirm(` callsites and migrate them too. Non-blocking follow-up territory.

PR #1913, opened 2026-04-28
ux: add aria-label to Obsidian export button in reader header (closes #1912)

alfmunny:
## What changed

Added `aria-label="Export vocabulary to Obsidian"` to the vocabulary export button in the reader header (`frontend/src/app/reader/[bookId]/page.tsx`).

## Why

The button had only `title="Export vocabulary to Obsidian"` and visible text `"Obsidian"`. Since the `ExportIcon` is `aria-hidden="true"` (baked into Icons.tsx), the button's computed accessible name was just **"Obsidian"** — not descriptive for screen reader users. `title=` is not reliably announced by NVDA/JAWS when visible text is present.

Adding `aria-label` makes the full action name — "Export vocabulary to Obsidian" — the accessible name, satisfying WCAG 2.1 §4.1.2 Name, Role, Value (Level A).

## Test plan

- New static source assertion test: `ReaderPageObsidianAriaLabel.test.ts`
  - Asserts `aria-label="Export vocabulary to Obsidian"` is present on the button
  - Asserts `aria-label=` appears near the `onClick={handleObsidianExport}` handler
- Full frontend suite: 2777 tests pass

Closes #1912

PR #1911, opened 2026-04-28
feat: mobile toolbar UX improvements + vocabulary language filter + fix long-press annotation conflict

alfmunny:
### Mobile bottom toolbar
- Remove redundant `‹`/`›` arrow buttons (swipe gestures and tap-zones already handle chapter navigation)
- Add 📝 Notes button: tapping opens a scrollable annotations panel above the toolbar; each entry jumps to the annotated sentence
- Fix 🌐 Translation button: `onDoubleClick` caused an iOS tap-cancel bug where `onClick` and `onDoubleClick` fired together, resetting state immediately — replaced with a clean single-tap toggle; also adds `setTranslationRequested(true)` so the translation API call actually fires (it was blocked by a manual-trigger guard that only the desktop sidebar had been setting)
- Standardise all toolbar elements to `h-10` for visual consistency with the chapter dropdown

### Vocabulary language filter (closes #1864-style request)
- Backend `get_vocabulary()` now joins the `books` table and returns `book_language` (first entry from the `languages` JSON array) on each occurrence — no schema migration needed
- `VocabularyOccurrence` TypeScript interface gains `book_language?: string | null`
- Vocabulary page shows language filter pill-tabs (All / Deutsch / English / …) when words span more than one language; single-language users see no change

### Bug fix: long-press annotation overlaps SelectionToolbar (closes #1865)
- **Root cause:** iOS fires `selectionchange` before the 400 ms annotation timer, causing `SelectionToolbar` to appear. When the timer fires, `AnnotationToolbar` also appears — both toolbars visible simultaneously.
- **Fix:** call `window.getSelection().removeAllRanges()` inside `handlePointerDown`'s timer callback (mirrors the existing `handleSegLongPress` path), clearing the native selection so `SelectionToolbar` dismisses before `AnnotationToolbar` opens. Also adds `navigator.vibrate(10)` haptic feedback.

## Test plan
- [ ] Mobile Safari/Chrome: bottom toolbar shows `🌐 ▶ 📝 [dropdown] 💬`, no arrows; swipe left/right navigates chapters
- [ ] Tap 📝 → notes panel slides up; tap an annotation → jumps to sentence
- [ ] Tap 🌐 → translation loads; tap again → translation off (no double-tap needed)
- [ ] Vocabulary page: save words from books in different languages → language tabs appear; selecting a tab filters correctly
- [ ] Mobile long-press on sentence → only `AnnotationToolbar` appears, no `SelectionToolbar` overlap
[...]

PR #1914, opened 2026-04-28
ux: add aria-label to Obsidian export button in reader header (closes #1912)

### LINKED ISSUE #1912: ux: Obsidian export button accessible name is 'Obsidian' — needs aria-label for screen readers
## Problem

The vocabulary-export button in the reader header has only `title="Export vocabulary to Obsidian"` and visible text `Obsidian`, but no `aria-label`:

```tsx
// frontend/src/app/reader/[bookId]/page.tsx ~line 1208
<button
  onClick={handleObsidianExport}
  title="Export vocabulary to Obsidian"
  className="hidden lg:flex …"
>
  <ExportIcon className="w-3.5 h-3.5 shrink-0" />
  Obsidian
</button>
```

The icon already has `aria-hidden="true"` (baked into Icons.tsx), so the button's computed accessible name is just **"Obsidian"** — which tells screen reader users nothing about what the button does.

`title=` is not reliably announced by NVDA/JAWS when a button already has a text label, so the descriptive tooltip is invisible to AT.

## Expected behaviour

Accessible name should be **"Export vocabulary to Obsidian"** so screen readers announce a complete action.

## Fix

Add `aria-label="Export vocabulary to Obsidian"` to the button.

## WCAG criterion

WCAG 2.1 §4.1.2 Name, Role, Value (Level A)

alfmunny:
## What changed

Added `aria-label="Export vocabulary to Obsidian"` to the vocabulary export button in the reader header (`frontend/src/app/reader/[bookId]/page.tsx`).

## Why

The button had `title="Export vocabulary to Obsidian"` and visible text `"Obsidian"`. Since `ExportIcon` is `aria-hidden="true"` (baked into Icons.tsx), the button's computed accessible name was just **"Obsidian"** — unhelpful for screen reader users. `title=` is not reliably announced by NVDA/JAWS when visible text is present.

Adding `aria-label` makes "Export vocabulary to Obsidian" the accessible name, satisfying WCAG 2.1 §4.1.2 Name, Role, Value (Level A).

## Test plan

- New static source assertion test: `ReaderPageObsidianAriaLabel.test.ts`
  - Asserts `aria-label="Export vocabulary to Obsidian"` is present
  - Asserts it appears near `onClick={handleObsidianExport}`
- Full frontend suite: 2777 tests pass

Closes #1912

### DIFF
--- a/frontend/src/__tests__/ReaderPageObsidianAriaLabel.test.ts
+++ b/frontend/src/__tests__/ReaderPageObsidianAriaLabel.test.ts
@@ -1,0 +1,25 @@
+/**
+ * Regression test for #1912:
+ * The Obsidian vocabulary-export button must have aria-label="Export vocabulary to Obsidian"
+ * so screen readers announce a complete action instead of just "Obsidian".
+ */
+import { readFileSync } from "fs";
+import { join } from "path";
+
+const src = readFileSync(
+  join(__dirname, "../app/reader/[bookId]/page.tsx"),
+  "utf8",
+);
+
+describe("Reader page Obsidian export button accessibility (closes #1912)", () => {
+  it("Obsidian export button has aria-label describing the export action", () => {
+    expect(src).toMatch(/aria-label="Export vocabulary to Obsidian"/);
+  });
+
+  it("Obsidian export button does not rely solely on title= for its accessible name", () => {
+    // Verify aria-label appears near the onClick={handleObsidianExport} button attribute
+    const exportBtnSection =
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.**

Clean refile of #1913 (closed) without the stacked-on-#1911 carryover I flagged. Now exactly 2 files: 1-line component change + 25-line test. WCAG 4.1.2 fix on Obsidian export button. Closes #1912.

Applied `P3`. Good responsiveness to the diff-noise feedback.

PR #1920, opened 2026-04-28
ux: add lang attribute to translated text in SentenceReader (closes #1919)

### LINKED ISSUE #1919: ux: translated text in reader missing lang attribute — screen readers use wrong pronunciation (WCAG 3.1.2)
## Problem

When a user reads a book with translation enabled (e.g. German → Chinese), the translated paragraphs rendered by `SentenceReader` have no `lang` attribute:

```html
<!-- Current — missing lang -->
<p class="font-serif text-base text-amber-800 italic">{translationText}</p>
```

Screen readers default to the page language (`<html lang="en">`) for pronunciation, so Chinese or German text is read with English phonemes. This violates WCAG 3.1.2 (Language of Parts).

## Root cause

`SentenceReader` has no `translationLang` prop. The reader page knows `translationLang` (state variable) but doesn't pass it to the component.

Two locations need the fix:
- `frontend/src/components/SentenceReader.tsx` ~line 874 (parallel mode)
- `frontend/src/components/SentenceReader.tsx` ~line 908 (inline mode)

## Fix

1. Add `translationLang?: string` to `SentenceReader` props
2. Add `lang={translationLang}` to both translated `<p>` elements
3. Pass `translationLang={translationLang}` in `page.tsx` where `<SentenceReader` is used

## Test plan
Static-source test asserting the translated `<p>` element has `lang=`.

alfmunny:
## What changed
- Added `translationLang?: string` prop to `SentenceReader` component
- Both parallel-mode and inline-mode translated `<p>` elements now carry `lang={translationLang}`
- Reader page passes `translationLang={translationEnabled ? translationLang : undefined}` to `<SentenceReader>` so the attribute is present only when translation is active

## Why
WCAG Success Criterion 3.1.2 (Language of Parts) requires that text in a language different from the page's primary language carries a `lang` attribute. Without it, screen readers use the page's default language engine to pronounce translated content, producing unintelligible output for non-Latin scripts.

## Test plan
- New test file `SentenceReaderTranslationLang.test.ts` (4 tests):
  - `SentenceReader` accepts a `translationLang` prop
  - Parallel-mode translated paragraph has `lang=` attribute
  - Inline-mode translated paragraph has `lang=` attribute
  - Reader page passes `translationLang=` to `<SentenceReader>`
- All 2789 frontend tests pass
- Backend suite passing (running in background)

Closes #1919

### DIFF
--- a/frontend/src/__tests__/SentenceReaderTranslationLang.test.ts
+++ b/frontend/src/__tests__/SentenceReaderTranslationLang.test.ts
@@ -1,0 +1,44 @@
+/**
+ * Regression test for #1919:
+ * Translated text rendered by SentenceReader must carry a lang attribute so
+ * screen readers use the correct pronunciation engine (WCAG 3.1.2).
+ */
+import { readFileSync } from "fs";
+import { join } from "path";
+
+const componentSrc = readFileSync(
+  join(__dirname, "../components/SentenceReader.tsx"),
+  "utf8",
+);
+
+const pageSrc = readFileSync(
+  join(__dirname, "../app/reader/[bookId]/page.tsx"),
+  "utf8",
+);
+
+describe("SentenceReader translated text lang attribute (closes #1919)", () => {
+  it("SentenceReader accepts a translationLang prop", () => {
+    expect(componentSrc).toMatch(/translationLang/);
+  });
+
+  it("parallel mode translated paragraph has lang attribute", () => {
+    // Locate the parallel-mode translation block (inside data-translation div)
+    const parallelBlock =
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM review.**

WCAG 3.1.2 (Language of Parts). Without `lang` on translated content, screen readers pronounce non-English translations (e.g. Chinese) using the page's English engine — produces unintelligible output. Conditional pass-through (`translationEnabled ? translationLang : undefined`) means the attribute appears only when translation is active. 6-line component change + 1-line page wiring + 44-line test. Closes #1919.

Solid catch — easy to miss but high-impact for AT users with non-Latin translations.

Applied `P2`.

Issue #1925, opened 2026-04-28
ux: notes page search-filtered empty state lacks serif headline and clear-search CTA


alfmunny:
## Problem

When the user searches in the Notes page and no books match, the UI shows only:

```tsx
<p className="text-sm">No books match your search.</p>
```

This violates the design system empty-state rule (CLAUDE.md Graphic design rules → Empty states): every empty state must have **a subtle illustration or icon, a headline (`font-serif`), a sub-text explanation, and a primary CTA button**.

The icon and headline already exist for the truly-empty case (`books.length === 0`), but the search-filtered sub-case skips them and gives the user no way to clear the search without scrolling back to the top.

## Reproduction

1. Open the Notes page with at least one book that has notes.
2. Type a query that matches no book titles in the search field.
3. See bare `<p className="text-sm">No books match your search.</p>` — no headline, no clear-search button.

## Proposed fix

Replace the bare `<p>` with a full empty state block that includes:
- `font-serif` headline: `No books match &ldquo;{search}&rdquo;`
- Sub-text: `Try a different search term.`
- CTA button: `Clear search` that calls `setSearch("")`

`frontend/src/app/notes/page.tsx` line ~177.

alfmunny:
**UI/UX**: Claimed by UI/UX Dev — starting work now.

PR #1934, opened 2026-04-28
ux: add regression test for EmptyNotesIcon in notes search-filtered empty state (closes #1927)

### LINKED ISSUE #1927: ux: notes search-filtered empty state still missing icon (follow-up to #1926)
## Problem

PR #1926 upgraded the notes overview page's search-filtered empty state with a font-serif headline, sub-text, and clear-search CTA — closing the bulk of #1925.

However, the design-system empty-state rule (CLAUDE.md → Graphic design rules → Empty states) requires **a subtle illustration or icon (SVG, not emoji)** in addition to the headline + sub-text + CTA. The truly-empty branch (`books.length === 0`) already shows an icon; the new filtered branch added by #1926 has the other three but no icon.

## Reference precedent

`#1924` (vocabulary filtered empty state, merged just before #1926) included `EmptyVocabIcon` at the top of the block. The same pattern should be applied here for visual consistency across filtered empty states.

## Proposed fix

In `frontend/src/app/notes/page.tsx` at the search-filtered branch (line ~177 after #1926), prepend the existing notes-page icon (or a suitable SVG from `@/components/Icons.tsx`) above the headline. Match the vocabulary page's styling.

## Out of scope

- The truly-empty case is already correct.
- This is purely a polish follow-up, not a regression.

alfmunny:
## What changed

`NotesSearchFilteredEmptyState.test.ts`: extended the existing regression test (#1925) to also assert that `EmptyNotesIcon` is present in the filtered empty state.

The icon was already in place — `EmptyNotesIcon` sits on the outer `filtered.length === 0` div, before the `books.length === 0` sub-conditional, so it renders for both the truly-empty and search-filtered branches. However the existing test's regex started from `books.length === 0` (the inner branch) and therefore never captured the icon.

Added a second regex anchored at `filtered.length === 0` through `Clear search` that covers the full outer block and asserts `EmptyNotesIcon` is present.

## Why

Closes the gap left by #1927: prevents future refactors from silently moving the icon inside only the truly-empty branch (which would break the filtered empty state visually with no test failure).

## Test plan

- New assertion in existing test file: `filtered.length === 0` block contains `EmptyNotesIcon`
- Full frontend suite: 2807 tests passed
- Backend suite running in CI

Closes #1927

## Docs follow-up

### DIFF
--- a/frontend/src/__tests__/NotesSearchFilteredEmptyState.test.ts
+++ b/frontend/src/__tests__/NotesSearchFilteredEmptyState.test.ts
@@ -1,7 +1,7 @@
 /**
- * Regression test for #1925:
- * Notes overview page search-filtered empty state must include a font-serif
- * headline and a clear-search CTA button — not just a bare <p>.
+ * Regression tests for #1925 and #1927:
+ * Notes overview page search-filtered empty state must include an SVG icon,
+ * a font-serif headline, and a clear-search CTA button.
  */
 import { readFileSync } from "fs";
 import { join } from "path";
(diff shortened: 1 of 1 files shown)


alfmunny:
**PM review.** Good catch — my filing of #1927 misdiagnosed the icon as missing when it's actually rendered by the outer `filtered.length === 0` wrapper that covers both sub-branches. The narrow regex in the original #1925 test just didn't capture it. Turning that into an explicit assertion at the wrapper level is the right move: prevents a future refactor from silently moving the icon inside only the truly-empty branch. Test-only PR, +17/-6, no source change. Closes #1927. Applied P3.

Issue #1939, opened 2026-04-28
ux: vocabulary and notes pages show 'please refresh' on load failure with no retry button


alfmunny:
## Problem

The vocabulary page, notes index page, and notes book page all show an error state when the API request fails:

```
Failed to load vocabulary. / Failed to load notes.
Please refresh the page to try again.
```

But there's no "Try again" button — users must manually reload the entire page. This is especially frustrating on mobile where a full reload is expensive.

**Affected pages:**
- `/vocabulary` — `fetchError` path (line 572)
- `/notes` — `fetchError` path (line 156)
- `/notes/[bookId]` — `fetchError` path (line 710)

## Fix

Extract the fetch call into a named `loadData` function and add a `<button>Try again</button>` CTA to each error state. Error state should also have an icon per design rules.

## Acceptance criteria

- Error state shows an SVG icon (e.g., `AlertCircleIcon`) + font-serif headline + retry CTA button
- Clicking retry re-runs the fetch without full page reload

alfmunny:
**UI/UX**: Claimed — starting work now.

Issue #1942, opened 2026-04-28
ux: extend error-state retry pattern to upload-chapters page (follow-up to #1939/#1940)


alfmunny:
## Context

PR #1940 (closes #1939) added an `AlertCircleIcon` + serif headline + "Try again" retry button to the vocabulary, notes index, and notes book error states. The pattern works well; the same shape can be applied to other user-facing error states for consistency.

## Surface to address

`frontend/src/app/upload/[bookId]/chapters/page.tsx` (lines 85-99) — the load-failure branch:

```tsx
<div role="alert" className="text-center max-w-sm">
  <p className="font-serif text-lg text-ink mb-2">Could not load chapters</p>
  <p className="text-sm text-red-600 mb-6">{error}</p>
  <button onClick={() => router.push("/upload")}>Try another file</button>
</div>
```

Has: serif headline ✓, sub-text ✓, CTA ✓
Missing: icon ✗, **retry CTA** (the existing button only navigates away — no in-place retry).

## Proposed fix

- Add `AlertCircleIcon` (already imported elsewhere) before the headline, `aria-hidden=\"true\"`
- Extract the chapter fetch into a named callback and add a primary "Retry" button that re-runs it without a navigation. Keep "Try another file" as a secondary CTA. Match the dual-button layout already used in the reader chapter error state (`reader/[bookId]/page.tsx` line ~1403, which has Retry + Back to library).

## Out of scope

- Reader chapter error state already has the full pattern (icon + serif headline + sub-text + Retry + secondary CTA) — no change needed.
- Admin pages (`admin/books`, `admin/audio`, etc.) and `QueueTab.tsx` use inline `<div>{error}</div>` patterns rather than full error-state blocks; those are different shape and lower priority.
- Flashcards / decks pages have no `fetchError` paths in the grep — confirm before broadening.

## Reference

- #1939 / PR #1940 for the established pattern
- `frontend/src/app/reader/[bookId]/page.tsx` line 1403 for the dual-button layout exemplar

alfmunny:
**UI/UX**: Claimed by UI/UX Dev — starting work now.

PR #1940, opened 2026-04-28
ux: add retry button to vocabulary and notes error states (closes #1939)

alfmunny:
## What changed

Vocabulary, notes index, and notes book pages all showed a bare "Please refresh the page" message when the API fetch failed — no retry action available.

This PR:
- Extracts the fetch call into a named `loadVocab` / `loadNotes` / `loadData` callback in each page
- Replaces the plain error text with a proper error state: `AlertCircleIcon` + font-serif headline + **Try again** button that calls the callback without a full page reload
- Updates the sub-copy from "Please refresh" → "Check your connection and try again."

## Why

Users on flaky connections had no recovery path except a manual browser refresh. A retry button is the minimum expected UX for an error state.

## Test plan

- [x] 9 new static-source assertions in `ErrorStateRetryButton.test.ts` — verify icon, font-serif headline, and "Try again" text in each page's error block
- [x] Full frontend suite: 2576 tests pass
- [x] Backend suite: passing

Closes #1939

[ ] Docs updated (if applicable)

PR #1943, opened 2026-04-28
ux: show proper error state on decks page fetch failure (closes #1941)

### LINKED ISSUE #1941: ux: decks page shows 'No study decks yet' on fetch failure instead of error state
## Problem

When the `listDecks()` API call fails (e.g. network error, server down), `decks/page.tsx` renders the empty state — "No study decks yet." with a "New deck" CTA — instead of an error message. This is misleading: a user with many decks sees what looks like an empty library.

Root cause: `const isEmpty = !loading && (error || decks.length === 0)` conflates the error case with genuine emptiness, so both render the same UI.

## Expected behaviour

Fetch failure → error state with AlertCircleIcon, font-serif headline ("Failed to load decks."), sub-text ("Check your connection and try again."), and a "Try again" retry button that re-calls `listDecks()`.

Genuine empty list → current "No study decks yet." empty state (correct).

## Reproduction

1. Open `/decks` while the backend is unreachable.
2. Observe: "No study decks yet." with "New deck" button.
3. Expected: error state with retry button.

## Fix

- Extract `listDecks` call into a `loadDecks` useCallback.
- In JSX: render error state (`fetchError`) before the empty state (`decks.length === 0`).
- Follow the same pattern as #1939 (vocabulary / notes pages).

alfmunny:
## What changed

`decks/page.tsx` previously combined `error || decks.length === 0` into a single `isEmpty` flag, causing both a fetch failure and a genuinely empty list to render the same "No study decks yet." UI — misleading for users who have decks but hit a temporary network error.

## Why

A user with decks who encounters a flaky connection saw an empty-library state with a "New deck" CTA, giving no indication of the real problem and no way to retry without a full page reload.

## Fix

- Extracted `listDecks` call into a `loadDecks` useCallback.
- Added a dedicated `fetchError` state; JSX now renders three distinct branches: loading → error (with retry) → empty (no decks) → list.
- Error state follows the same pattern as #1939: `AlertCircleIcon` + font-serif headline + "Try again" button that calls `loadDecks` directly.
- Removed the stale test assertion that validated the wrong behaviour and added a new `DecksPage.fetcherror.test.tsx` with 3 tests covering the error state, retry flow, and genuine-empty distinction.

## Test plan

- [ ] `npx jest "DecksPage" --no-coverage` — 8 tests pass
- [ ] Full `npm test -- --no-coverage --ci` — 2831 tests pass (450 suites)
- [ ] Backend pytest — 1942 passed

Closes #1941

### DIFF
--- a/frontend/src/__tests__/DecksPage.fetcherror.test.tsx
+++ b/frontend/src/__tests__/DecksPage.fetcherror.test.tsx
@@ -1,0 +1,88 @@
+/**
+ * Regression test for #1941: decks page must show a proper error state
+ * (with a retry button) when listDecks() fails — not the empty-decks state.
+ */
+import React from "react";
+import { render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+jest.mock("next-auth/react", () => ({
+  useSession: () => ({ status: "authenticated", data: { backendToken: "token" } }),
+}));
+
+const mockPush = jest.fn();
+jest.mock("next/navigation", () => ({
+  useRouter: () => ({ push: mockPush }),
+}));
+
+jest.mock("@/lib/api", () => ({
+  listDecks: jest.fn(),
+  deleteDeck: jest.fn(),
+}));
+
+import * as api from "@/lib/api";
+import DecksPage from "@/app/decks/page";
+
+const mockListDecks = api.listDecks as jest.MockedFunction<typeof api.listDecks>;
+
+beforeEach(() => {
+  jest.clearAllMocks();
+});
+
+const flushPromises = () => new Promise((r) => setTimeout(r, 0));
+
+test("shows error state with retry button on fetch failure — not empty-decks state", async () => {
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM review.** Genuine bug fix — the `isEmpty = !loading && (error || decks.length === 0)` conflation was actively misleading users with decks who hit a network error (showing "No study decks yet" + "New deck" CTA instead of a retry path). The three-way branch split (loading → error → empty → list) is the right correctness fix on top of the UX polish.

Cleanly mirrors #1940's extract-into-callback pattern. New test file with retry-flow and error-vs-empty distinction is exactly what's needed to lock the fix in.

Side note: my grep for the follow-up #1942 didn't catch this because the page used `error` directly inside the `isEmpty` derivation rather than a dedicated `fetchError` boolean — good independent catch by UI/UX.

Closes #1941. Applied P2 (correctness defect on a core flow); auto-merge will land it on green CI.

PR #1945, opened 2026-04-28
ux: show error state on popular-books fetch failure in home page (closes #1944)

### LINKED ISSUE #1944: ux: home page popular-books section shows misleading empty state on fetch failure
## Problem

In `app/page.tsx`, the Discover tab fetches popular books with:
```ts
.catch(() => { setPopularBooks([]); setPopularTotal(0); })
```

On API failure, `popularBooks` is set to `[]` and the empty-state message — "No popular books available yet." — is displayed. Users have no way to distinguish a network error from a genuinely empty catalogue, and no retry button.

## Expected behaviour

Fetch failure → error state with AlertCircleIcon, sub-text ("Failed to load books. Check your connection."), and a "Try again" retry button that re-fetches popular books for the current language/page.

Genuine empty response → current "No popular books available yet." message (correct).

## Fix

- Track a separate `popularError` boolean.
- On catch, set `popularError = true` instead of silently setting an empty list.
- In JSX: render an error state when `popularError` is true; render empty-state only when the fetch succeeded but returned zero books.
- Add a retry button that calls the fetch again (reset `popularError` first).

Follows the same pattern as #1939 (vocabulary/notes) and #1941 (decks).

alfmunny:
## What changed

`app/page.tsx` silently coerced `getPopularBooks()` errors into an empty list:
```ts
.catch(() => { setPopularBooks([]); setPopularTotal(0); })
```
This caused network failures on the Discover tab to render "No popular books available yet." — indistinguishable from a genuinely empty catalogue.

## Why

Users on flaky connections had no indication the catalogue failed to load and no way to retry without a full page reload.

## Fix

- Added a `popularError` boolean state.
- On catch, set `popularError = true` (don't clobber the existing list).
- Added a `loadPopularBooks` function to enable retry without re-triggering the effect.
- JSX now renders three distinct states: loading → error (AlertCircleIcon + "Try again" button) → empty (no books returned) → list.
- Updated the stale `HomePage.branches.test.tsx` assertion that validated the old wrong behaviour.
- Added `HomePage.popularerror.test.tsx` with 3 tests: error state shown, retry restores list, genuine-empty shows correct message.

Follows the same pattern as #1939 (vocabulary/notes), #1941 (decks), and prior retry-button fixes.

## Test plan

- [ ] `npx jest "HomePage.popular" --no-coverage` — 3 tests pass
- [ ] Full `npm test -- --no-coverage --ci` — 2834 tests pass (451 suites)

Closes #1944

### DIFF
--- a/frontend/src/__tests__/HomePage.branches.test.tsx
+++ b/frontend/src/__tests__/HomePage.branches.test.tsx
@@ -320,15 +320,17 @@ describe("HomePage — getReadingProgress merges backend data (lines 73-83)", ()
 // ── Line 116: getPopularBooks catch branch ────────────────────────────────────
 
 describe("HomePage — getPopularBooks error (line 116)", () => {
-  it("sets empty popular books when API fails", async () => {
+  it("shows error state (not empty-state) when API fails", async () => {
     mockGetPopularBooks.mockRejectedValue(new Error("Network error"));
 
     await renderHome();
 
-    // Should show "No popular books available yet."
+    // Should show error state with retry button, not the empty-library message
     await waitFor(() =>
-      expect(screen.getByText("No popular books available yet.")).toBeInTheDocument(),
+      expect(screen.getByRole("alert")).toBeInTheDocument(),
     );
+    expect(screen.queryByText("No popular books available yet.")).not.toBeInTheDocument();
+    expect(screen.getByRole("button", { name: /try again/i })).toBeInTheDocument();
   });
 });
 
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM review.** Same shape of correctness fix as #1943 — silent error → empty-list coercion was misleading users (network failures rendering as a genuinely empty catalogue with no retry path). Three-branch render is the right pattern, and not clobbering the existing `popularBooks` on catch preserves the prior page's results across a transient error.

"On retry, reset `popularError` first" is the correct ordering.

Bumping the issue's P3 to **P2** on the PR — same reasoning as #1943: this is a correctness defect on the home Discover surface, not just polish, and Discover is core to the new-user flow. UI/UX may keep filing these as P3 by default; PM is sweeping to P2 when the underlying bug is misleading-state-on-failure. Closes #1944. Auto-merge will land it on green CI.

PR #1950, opened 2026-04-28
ux: add error state + retry to flashcards page when fetch fails (closes #1949)

### LINKED ISSUE #1949: ux: flashcards page shows blank card area on API error — missing error state + retry
## Problem

\`frontend/src/app/vocabulary/flashcards/page.tsx\` — \`loadData()\` swallows errors silently:

\`\`\`ts
const loadData = useCallback(async () => {
  setLoading(true);
  try {
    const [due, statsData] = await Promise.all([getDueFlashcards(...), getFlashcardStats(...)]);
    setCards(due);
    setStats(statsData);
    setDone(due.length === 0);
  } finally {
    setLoading(false);
  }
}, [selectedDeckId]);
\`\`\`

When the API fails the \`catch\` path is missing. Result: \`loading = false\`, \`cards = []\`, \`done = false\` → the user sees the header and deck selector with a completely blank card area — no error message, no retry button.

The "All done for today!" state only renders when \`done = true\` (set only on success), so the error path produces an unreachable-looking blank region.

## Proposed fix

- Add \`fetchError\` boolean state to \`loadData\`
- On catch: set \`fetchError = true\`
- Below the deck selector, render a \`role="alert"\` error state (matching the established pattern) with \`AlertCircleIcon\`, a serif headline, sub-text, and a **Retry** button calling \`loadData()\`
- Clear \`fetchError\` at the top of \`loadData\` on each invocation

## Out of scope

- Grade-submission errors (\`handleGrade\`) already surface via a silently-dropped \`finally\` — those are lower priority since the grade not saving doesn't leave the user blocked (they can re-flip and grade again). File as a separate issue if needed.

## Reference

- PR #1946 (upload-chapters), PR #1948 (deck-detail) for the \`AlertCircleIcon + Retry\` pattern

alfmunny:
## What changed

`frontend/src/app/vocabulary/flashcards/page.tsx` — `loadData()` had no `catch` block: API failures were silently dropped, leaving `loading = false`, `cards = []`, `done = false` → the user saw a blank card area with no indication of what went wrong and no way to retry.

Changes:
- Added `fetchError` boolean state
- Added `catch` to `loadData` → sets `fetchError = true`; clears it on each retry
- Added an error state JSX block (below the deck selector, above the done/card area):
  - `role="alert"` wrapper
  - `AlertCircleIcon` (`aria-hidden="true"`)
  - Serif headline "Failed to load flashcards."
  - Sub-text "Check your connection and try again."
  - Primary **Retry** button calling `loadData()` directly
- Guarded the card content branch (`!fetchError && currentCard`) so error and card content never render together

## Why

Closes #1949. When `getDueFlashcards` or `getFlashcardStats` fail on a slow or interrupted connection, users had no feedback and no recovery path — just a blank screen where the cards should appear. The existing "All done for today!" state correctly fires when the queue is genuinely empty (success path), but the error path was completely unhandled.

## Test plan

- `frontend/src/__tests__/FlashcardsPage.errorstate.test.tsx` (new): 3 regression tests
  1. Error state renders `role="alert"`, headline, Retry button when fetch fails (not blank)
  2. Retry button re-fetches and shows cards on success
  3. Genuine empty queue still shows "All done for today!" — not the error state
- Full frontend suite: 2840/2840 passed

## Docs follow-up

No change needed — refinement of existing page, not a new UX pattern.

### DIFF
--- a/frontend/src/__tests__/FlashcardsPage.errorstate.test.tsx
+++ b/frontend/src/__tests__/FlashcardsPage.errorstate.test.tsx
@@ -1,0 +1,91 @@
+/**
+ * Regression test for #1949: flashcards page must show an error state + retry
+ * button when getDueFlashcards fails — not a blank card area.
+ */
+import React from "react";
+import { render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+jest.mock("next-auth/react", () => ({
+  useSession: () => ({ status: "authenticated", data: { backendToken: "tok" } }),
+}));
+
+jest.mock("next/navigation", () => ({
+  useRouter: () => ({ push: jest.fn() }),
+}));
+
+jest.mock("@/lib/api", () => ({
+  getDueFlashcards: jest.fn(),
+  getFlashcardStats: jest.fn(),
+  reviewFlashcard: jest.fn(),
+  listDecks: jest.fn(),
+}));
+
+import * as api from "@/lib/api";
+import FlashcardsPage from "@/app/vocabulary/flashcards/page";
+
+const mockGetDueFlashcards = api.getDueFlashcards as jest.MockedFunction<typeof api.getDueFlashcards>;
+const mockGetFlashcardStats = api.getFlashcardStats as jest.MockedFunction<typeof api.getFlashcardStats>;
+const mockListDecks = api.listDecks as jest.MockedFunction<typeof api.listDecks>;
+
+const STATS = { reviewed_today: 0, due_today: 1, total_words: 5 };
+const CARD = {
+  vocabulary_id: 1,
+  word: "Schadenfreude",
+  definition: "pleasure from others' misfortune",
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.** Third correctness fix in this thread (after #1943 / #1945) — `try/finally` with no `catch` was swallowing errors entirely, leaving the page in an indistinguishable-from-bug state (blank card region, no "done", no error). The `!fetchError && currentCard` guard is the right way to keep error and card states from coexisting; the third regression test (genuine-empty still shows "All done for today!") locks in the success-vs-error distinction.

Pragmatic on the grade-submission `finally` — leaving that for a follow-up is the right call since the user isn't blocked there. If a separate issue is filed, link it on this PR for traceability.

Closes #1949. Applied P2 (silent error → blank UI is correctness, not polish — matches the #1943/#1945 family); auto-merge will land it on green CI.

Issue #1953, opened 2026-04-28
ux: inline action-error banners on upload, search, and home page missing AlertCircleIcon


alfmunny:
## Problem

Three pages show bare red text banners without an icon when an inline user action fails. These use `role="alert"` correctly but have no visual icon to help users quickly identify the error:

- `/upload` (`upload/page.tsx:189`) — file upload failure
- `/search` (`search/page.tsx:169`) — in-app search failure  
- `/` (home page `page.tsx:605`) — Gutenberg book search failure

Unlike the initial-load error series (issues #1939–#1951), these are action-triggered errors so a standalone Retry button is not needed — the user retries by repeating the action. But the `AlertCircleIcon` is missing for visual consistency with the rest of the app.

## Expected behaviour

Each banner should include `AlertCircleIcon` at the start, matching the established pattern in `actError` banners on admin pages.

## Fix

Add `<AlertCircleIcon className="w-4 h-4 shrink-0" aria-hidden="true" />` to each bare error banner and make the banner a flex row.

alfmunny:
**UI/UX**: Claimed — implementing now.

alfmunny:
**UI/UX**: PR #1954 merged — closing.

PR #1954, opened 2026-04-28
ux: add AlertCircleIcon to inline action-error banners (closes #1953)

alfmunny:
## What changed

Three pages displayed bare red text `role="alert"` banners with no icon when an inline user action failed:

- **`/upload`** — file upload failure
- **`/search`** (in-app content search) — search API failure
- **`/`** (home page) — Gutenberg book search failure

Added `AlertCircleIcon` to each banner as a flex row, matching the visual pattern established across admin action-error banners.

## Why

Unlike the initial-load error series fixed in #1939–#1952, these are action-triggered errors where the user's natural retry is to repeat the action. A dedicated Retry button is not needed, but the icon was missing for visual consistency.

## Test plan

- [x] New `SearchPage.errorstate.test.tsx`: search error alert contains SVG icon; alert clears after successful re-fetch
- [x] New `UploadPage.erroricon.test.tsx`: upload error alert contains SVG icon
- [x] Full frontend suite: 2854/2854 passing (includes newly merged `AdminPages.errorstate.test.tsx` from #1952)

Closes #1953

PR #1956, opened 2026-04-28
ux: add AlertCircleIcon to import and chapter-editor error banners (closes #1955)

### LINKED ISSUE #1955: ux: inline error banners on import and chapter-editor pages missing AlertCircleIcon
## Problem

Two pages have bare `role="alert"` banners with no icon when an inline action fails:

- `/import/[bookId]` (`import/[bookId]/page.tsx:299`) — import job failure  
- `/upload/[bookId]/chapters` (`upload/[bookId]/chapters/page.tsx:153`) — confirm-chapters action failure

Both pages already import `AlertCircleIcon` but don't use it in the error banner.

## Fix

Add `AlertCircleIcon` + flex row to each bare error banner. Both are action-triggered inline errors (no Retry button needed — user retries the action).

alfmunny:
## What changed

Two pages (`/import/[bookId]` and `/upload/[bookId]/chapters`) had bare `role="alert"` banners with no icon when an inline action failed. Both pages already imported `AlertCircleIcon` but weren't using it in the error banner.

- **`/import/[bookId]`**: error shown when import stream throws (network error or API error)
- **`/upload/[bookId]/chapters`**: error shown when `confirmChapters` fails after chapters have loaded

Added `AlertCircleIcon` + flex row to each, matching the pattern from #1953.

## Why

Completes the error-banner icon audit. All `role="alert"` banners in the app now have an icon.

## Test plan

- [x] `ChapterEditor.erroricon.test.tsx`: confirm-error alert has SVG after `confirmChapters` rejects
- [x] `ImportPage.erroricon.test.tsx`: import-error alert has SVG after stream throws
- [x] Full frontend suite: 2853/2853 passing

Closes #1955

### DIFF
--- a/frontend/src/__tests__/ChapterEditor.erroricon.test.tsx
+++ b/frontend/src/__tests__/ChapterEditor.erroricon.test.tsx
@@ -1,0 +1,55 @@
+/**
+ * Regression test for #1955: chapter-editor inline confirm-error banner
+ * must include AlertCircleIcon (SVG) when confirmChapters fails.
+ */
+import React from "react";
+import { render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+jest.mock("next/navigation", () => ({
+  useParams: () => ({ bookId: "42" }),
+  useRouter: () => ({ push: jest.fn() }),
+}));
+
+const mockGetDraftChapters = jest.fn();
+const mockConfirmChapters = jest.fn();
+
+jest.mock("@/lib/api", () => ({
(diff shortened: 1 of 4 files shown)


alfmunny:
**PM review.** Closes the icon audit — body explicitly notes "All `role=\"alert\"` banners in the app now have an icon" which is the right scope marker. Same one-line-per-page shape as #1954, both tests cover the SVG-after-rejection assertion. Both pages already imported `AlertCircleIcon`; just hadn't used it in the banner.

Closes #1955. Applied P3; auto-merge will land it on green CI.

PR #1958, opened 2026-04-28
ux: add title attributes to truncated text in admin tables (closes #1957)

### LINKED ISSUE #1957: ux: truncated text in admin tables lacks title attributes — hover reveals nothing
## Problem

Admin table cells that truncate long text with CSS `truncate` do not have a `title` attribute, so hovering over a truncated cell reveals nothing. Admins cannot see full book titles, filenames, user names, or email addresses without clicking through to a detail view.

## Affected elements

| File | Element | Data |
|---|---|---|
| `admin/uploads/page.tsx:157` | `<td class="...truncate">` | `u.title` (book title) |
| `admin/uploads/page.tsx:158` | `<td class="...truncate">` | `u.filename` |
| `admin/books/page.tsx:374` | `<span class="...truncate">` | `b.title` |
| `admin/users/page.tsx:106` | `<span class="...truncate">` | `u.name` |
| `admin/users/page.tsx:114` | `<p class="...truncate">` | `u.email` |

Note: `admin/uploads/page.tsx:165` (`uploader_email`) already has `title={u.uploader_email}` — that pattern should be applied to all the above.

## Fix

Add `title={value}` to each truncated element so sighted users can hover to reveal the full text.

## Reproduction

Go to `/admin` → any of the three tabs. Find a book/user/upload with a long title. The truncated text has no tooltip.

alfmunny:
## What changed

Added `title` attributes to truncated text elements in all three admin panels so hovering over a truncated cell reveals the full content:

- `admin/uploads`: book title (`u.title`) and filename (`u.filename`) table cells
- `admin/users`: user name (`u.name`) span and email (`u.email`) paragraph
- `admin/books`: book title (`b.title`) span

The `uploader_email` column in uploads already had this pattern — this change applies it consistently.

## Why

Admins managing books, users, and uploads had no way to see the full content of long titles, filenames, or email addresses — the truncated text was unrecoverable without navigating away.

## Test plan

- `AdminTruncateTitle.test.tsx` — 3 new regression tests (one per admin page) verify that each truncated element has `title={value}` matching the displayed text
- Full Jest suite: 2859 tests pass
- Full pytest suite: passing

Closes #1957

### DIFF
--- a/frontend/src/__tests__/AdminTruncateTitle.test.tsx
+++ b/frontend/src/__tests__/AdminTruncateTitle.test.tsx
@@ -1,0 +1,113 @@
+/**
+ * Regression test for #1957: truncated text in admin tables must have
+ * title attributes so users can hover to reveal full content.
+ */
+import React from "react";
+import { render, screen } from "@testing-library/react";
+
+jest.mock("next/navigation", () => ({
+  useRouter: () => ({ push: jest.fn() }),
+}));
+
+const mockAdminFetch = jest.fn();
+jest.mock("@/lib/adminFetch", () => ({
+  adminFetch: (...args: unknown[]) => mockAdminFetch(...args),
+}));
+
+jest.mock("@/lib/api", () => ({
+  getMe: () => Promise.resolve({ id: 99 }),
+}));
+
+jest.mock("@/components/SeedPopularButton", () => {
+  const Seed = () => <button>Seed</button>;
+  Seed.displayName = "SeedPopularButton";
+  return { __esModule: true, default: Seed };
+});
+
+beforeEach(() => {
+  jest.resetAllMocks();
+});
+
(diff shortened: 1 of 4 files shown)


alfmunny:
**PM review.** Right approach — `title` is the standard pattern for sighted-user hover on visually-truncated text (VSCode, GitHub, etc. use it the same way). Screen readers don't need a separate ARIA fix here because CSS `text-overflow: ellipsis` truncates the rendered glyph but not the DOM text, so AT still reads the full string.

The `uploader_email` column was a good precedent to anchor the pattern. Three regression tests assert `title={value}` matches the displayed text — locks in the consistency.

Closes #1957. Applied P3 (visual polish for sighted users, not a correctness or a11y gap); auto-merge will land it on green CI.

PR #1964, opened 2026-04-28
ux: add aria-label to BookCard main button (closes #1963)

### LINKED ISSUE #1963: ux: BookCard main button missing aria-label — screen readers don't announce the open action
## Problem

The `BookCard` component's main clickable button (the book cover + title + authors area) has no `aria-label`. Screen readers compute the accessible name from the button's text content and announce something like `"Jane Eyre Charlotte Brontë Ch. 1 · 2h ago"` with no indication that it's an "Open" action.

By contrast, the list-view book row buttons on the home page's Popular Classics section already use:
```tsx
aria-label={`Open ${book.title}${book.authors.length ? ` by ${book.authors.join(", ")}` : ""}`}
```

## Expected behaviour

`BookCard` should match the same pattern so screen reader users hear "Open Jane Eyre by Charlotte Brontë" when navigating by button.

## Fix

Add `aria-label={`Open ${book.title}${book.authors.length ? ` by ${book.authors.join(", ")}` : ""}`}` to the main `<button>` in `frontend/src/components/BookCard.tsx` (line 31).

## Scope

Single-component, one-line change. Regression test already exists in `BookCard.test.tsx`.

alfmunny:
## What changed

Added `aria-label` to the main clickable button in `BookCard` (`frontend/src/components/BookCard.tsx`).

**Before:** Screen readers computed an accessible name from the button's text content, announcing something like *"Jane Eyre Charlotte Brontë Ch. 1 · 2h ago"* with no indication of the action.

**After:** The button explicitly announces *"Open Jane Eyre by Charlotte Brontë"*, matching the same pattern already used by list-view book rows in the Popular Classics section on the home page.

## Why

WCAG 2.4.6 — controls should have descriptive accessible names that communicate both the action and the target. The generic text-content name was ambiguous to screen reader users navigating by button shortcut.

## Test plan

- Added 3 regression tests to `BookCard.test.tsx`:
  - `aria-label` with title + authors
  - `aria-label` with title only (no authors)
  - `aria-label` with multiple authors joined by comma
- All 15 `BookCard` tests pass; full 2868-test suite passes.

Closes #1963

### DIFF
--- a/frontend/src/__tests__/BookCard.test.tsx
+++ b/frontend/src/__tests__/BookCard.test.tsx
@@ -99,3 +99,26 @@ test("remove button meets 44px minimum touch target size", () => {
   expect(removeBtn.className).toContain("min-w-[44px]");
   expect(removeBtn.className).toContain("min-h-[44px]");
 });
+
+// ── aria-label on main card button (#1963) ─────────────────────────────────────
+
+test("main card button has aria-label announcing the open action and book title", () => {
+  render(<BookCard book={BOOK} onClick={jest.fn()} />);
+  // When onRemove is absent there is only one button — the card itself.
+  const card = screen.getByRole("button", { name: `Open ${BOOK.title} by ${BOOK.authors[0]}` });
+  expect(card).toBeInTheDocument();
+});
+
+test("main card button aria-label omits 'by ...' when no authors", () => {
+  render(<BookCard book={{ ...BOOK, authors: [] }} onClick={jest.fn()} />);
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.** +1 line, mirrors the established `Open ${title} by ${authors}` pattern from the home Popular Classics list-view rows. Three regression tests cover the with-authors / without-authors / multi-authors string-format paths. WCAG 2.4.6 fix.

Closes #1963. Applied P3 (single-component a11y polish); auto-merge will land it on green CI.

PR #1972, opened 2026-04-28
ux: add aria-label + role=status to admin books translating badge (closes #1971)

### LINKED ISSUE #1971: ux: admin books 'translating' badge uses → arrow without screen-reader label and lacks aria-live
## Problem

In `app/admin/books/page.tsx:378`, the active-translation badge renders:

```
translating → de
```

Two issues:
1. **Screen-reader text**: The `→` arrow is announced as *"right arrow"* — screen readers say *"translating right arrow de"*. Should say *"translating to de"*.
2. **No live notification**: When `b.active` becomes `true` (translation starts), there is no `aria-live` region to announce the state change. Admin users monitoring the translation queue who rely on a screen reader won't know when translation starts.

## Fix

```tsx
// Before
<span className="text-xs px-1.5 py-0.5 rounded-full bg-emerald-100 text-emerald-700 animate-pulse">
  translating → {b.active_language}
</span>

// After
<span
  className="text-xs px-1.5 py-0.5 rounded-full bg-emerald-100 text-emerald-700 animate-pulse"
  role="status"
  aria-label={`Translating to ${b.active_language}"}
>
  translating → {b.active_language}
</span>
```

## Scope
Admin panel only — lower-priority than user-facing flows but still a WCAG 1.3.1 violation.

alfmunny:
## What changed

`app/admin/books/page.tsx`: The active-translation badge (`translating → de`) now has `role="status"` and `aria-label="Translating to <lang>"`.

**Before:** Screen readers announced *"translating right arrow de"* — the `→` glyph had no text alternative (WCAG 1.3.1 violation).  
**After:** Screen readers announce *"Translating to de"*, and the `role="status"` ensures AT users are notified when translation becomes active.

## Why

Unicode directional arrows (`→`) have no implicit accessible name. The badge is informational — admin users monitoring translation state deserve accurate AT output. Closes #1971.

## Test plan

- [x] `AdminBooksTranslatingBadge.test.tsx` — queries `getByRole("status", { name: /translating to en/i })` on an active book; fails before fix, passes after
- [x] Full Jest suite: 2890 tests pass

### DIFF
--- a/frontend/src/__tests__/AdminBooksTranslatingBadge.test.tsx
+++ b/frontend/src/__tests__/AdminBooksTranslatingBadge.test.tsx
@@ -1,0 +1,69 @@
+/**
+ * Regression test for #1971: admin books "translating" badge must have
+ * aria-label="Translating to <lang>" so screen readers don't announce
+ * "translating right arrow de".
+ */
+import React from "react";
+import { render, screen } from "@testing-library/react";
+
+const mockAdminFetch = jest.fn();
+
+jest.mock("@/lib/adminFetch", () => ({
+  adminFetch: (...args: unknown[]) => mockAdminFetch(...args),
+}));
+
+jest.mock("next/navigation", () => ({
+  useRouter: () => ({ push: jest.fn() }),
+}));
+
+jest.mock("@/components/SeedPopularButton", () => {
+  const Seed = ({ onComplete }: { onComplete: () => void }) => (
+    <button onClick={onComplete}>Seed popular</button>
+  );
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.** Same shape as #1970 plus the bonus `role="status"` for live-region announcement when translation begins. Conditionally rendering the entire span based on `b.active` (rather than toggling a child) is the right pattern — AT will announce when the badge first appears.

Closes #1971. Applied P3; auto-merge will land it on green CI.

PR #2002, opened 2026-04-28
fix: move focus into alertdialog container on open, restore on close (closes #2001)

### LINKED ISSUE #2001: ux: focus management on admin/books + QueueTab confirmation dialogs (follow-up to #1999/#2000)
## Context

PR #2000 (closes #1999) switched the admin/books and QueueTab confirmation prompts from `role="dialog"` to `role="alertdialog"` so screen readers announce the prompt on appearance. That fixes the SR-announcement half of #1999.

The other half of #1999 — **keyboard focus management** — is not yet implemented. When a keyboard user clicks Delete (or activates with Enter/Space), focus stays on the Delete button rather than moving into the dialog's Confirm button. Users have to Tab forward to reach the dialog. This is the WCAG 2.1 SC 2.4.3 (Focus Order) gap that #1999 originally called out.

## Fix

Apply the focus-management pattern from #1999:

```tsx
const confirmBtnRef = useRef<HTMLButtonElement>(null);

useEffect(() => {
  if (pendingConfirm) confirmBtnRef.current?.focus();
}, [pendingConfirm]);

// On the Confirm button JSX:
<button ref={confirmBtnRef} onClick={...}>Confirm</button>
```

When the dialog closes (pendingConfirm → null), restore focus to the triggering element. The pattern is already in use in other modal components in this codebase — match the existing convention.

## Surfaces

- `src/app/admin/books/page.tsx` — Delete / Retranslate / Move confirmation prompt
- `src/components/QueueTab.tsx` — Stop / Retry queue action confirmation prompt

## Reference

- #1999 / PR #2000 for the role-attribute half
- Existing modal focus-management examples in the codebase (e.g. AddWordPicker, AnnotationToolbar — both already use `useRef` + `useEffect` focus shift)

alfmunny:
## What changed

- `admin/books/page.tsx` and `components/QueueTab.tsx`: added `useRef` + `useEffect` focus management on the alertdialog confirm prompt
  - When the dialog opens, focus moves to the dialog **container div** (`tabIndex={-1}`, not the Confirm button) so Enter key does not accidentally auto-confirm
  - When the dialog closes (Confirm or Cancel), focus returns to the element that triggered it
  - `outline-none` on the focusable div suppresses the default focus ring since the dialog already has a visible amber border

## Why

WCAG 2.1 SC 2.4.3 (Focus Order) requires that keyboard focus move inside a dialog when it opens. Without this, keyboard users must Tab forward through all the page content to reach the Confirm/Cancel buttons.

Using the container div (not the Confirm button) as the focus target avoids the Enter-auto-confirm trap: if focus landed on Confirm, the same Enter key that triggered the dialog would immediately fire Confirm.

## Test plan

- `ConfirmDialogFocusMgmt.test.tsx` (new): 4 tests
  - admin/books: dialog container has focus after Delete button click
  - admin/books: Delete button has focus after Cancel
  - QueueTab: dialog container has focus after Stop button click
  - QueueTab: Stop button has focus after Cancel
- Full suite: 3012 frontend tests + 1942 backend tests — all pass

Closes #2001

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -176,6 +176,8 @@ Second pass covered: vocabulary page, notes page, profile page, import page, Ann
 | 2026-04-28 | 8.4 | Import and chapter-editor error banners: AlertCircleIcon added | ✅ Done |
 | 2026-04-28 | 8.5 | Admin tables: title attributes on truncated book/user/file text | ✅ Done |
 | 2026-04-28 | 8.6 | Admin dismiss error/message buttons: 44×44px mobile touch targets | ✅ Done |
+| 2026-04-28 | 8.7 | Admin/QueueTab confirm dialogs: role="alertdialog" + aria-modal (PR #2000) | ✅ Done |
+| 2026-04-28 | 8.8 | Admin/QueueTab confirm dialogs: focus on container open, restore on close (PR #2002) | ✅ Done |
 
 ---
 
--- a/frontend/src/__tests__/ConfirmDialogFocusMgmt.test.tsx
+++ b/frontend/src/__tests__/ConfirmDialogFocusMgmt.test.tsx
@@ -1,0 +1,161 @@
+/**
+ * Regression tests for #2001 — confirmation dialogs in admin/books and QueueTab
+ * must move keyboard focus into the dialog container when opened (WCAG 2.1 SC 2.4.3)
+ * and restore it to the triggering element when closed.
+ *
+ * The dialog container uses tabIndex={-1} (not the Confirm button) so that
+ * pressing Enter to open the dialog does NOT accidentally auto-confirm.
+ */
+import React from "react";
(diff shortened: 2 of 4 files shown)


alfmunny:
**PM review.** Better than the spec I wrote in #2001 — focusing the dialog **container div** (`tabIndex={-1}`) instead of the Confirm button avoids the auto-confirm trap (same Enter key that triggered the dialog would otherwise fire Confirm immediately). That's the kind of UX trap I missed when filing; nice catch.

Four regression tests across both surfaces and both directions (open → dialog focus, cancel → trigger focus) lock the behavior in. Also good catch on `outline-none` to suppress the default focus ring since the dialog already has a visible amber border — keeps the visual unchanged for sighted users.

Design Change Log entry in `docs/design-improvement-plan.md` follows the per-PR docs check rule from CLAUDE.md.

Closes #2001. Applied P3; auto-merge will land it on green CI.

PR #2022, opened 2026-04-28
test: cover lib/settings.ts SSR-guard branches (closes #2015)

### LINKED ISSUE #2015: test: lib/settings.ts missing 2 SSR-guard branch coverage (lines 43, 53)
Coverage report shows `lib/settings.ts` at 67% branch coverage. The 2 missing branches are the `typeof window === "undefined"` guards in `getSettings` (line 43) and `saveSettings` (line 53). These simulate server-side rendering where `window` is not available.

**Fix:** Add two tests that temporarily set `global.window = undefined` to exercise the SSR paths.

alfmunny:
## What changed

- `frontend/jest.setup.js`: guarded `window.HTMLElement.prototype.scrollIntoView` behind `typeof window !== "undefined"` so tests using `@jest-environment node` can load the setup file without throwing a ReferenceError.
- `frontend/src/__tests__/settings.ssr.test.ts`: new test file with `@jest-environment node` that hits the two uncovered SSR-guard branches in `lib/settings.ts` (lines 43 and 53):
  - `getSettings` returns defaults when window is undefined (SSR)
  - `saveSettings` is a no-op when window is undefined (SSR)

## Why

`lib/settings.ts` had 66.7% branch coverage. The 2 missing branches are intentional SSR guards (`if (typeof window === "undefined") return ...`). These can only be exercised in a Node.js environment, but the shared jest.setup.js accessed `window` unconditionally, blocking `@jest-environment node` usage. Guarding that line is a safe, backward-compatible change — jsdom tests are unaffected.

## Test plan

- `npm --prefix frontend test -- --no-coverage --ci` — 496 suites, 3035 tests, all pass
- `lib/settings.ts` coverage: lines 100%, branches 100% (was 66.7%)

Closes #2015

## Docs follow-up

N/A — internal test infrastructure change only.

### DIFF
--- a/frontend/jest.setup.js
+++ b/frontend/jest.setup.js
@@ -2,8 +2,11 @@ import "@testing-library/jest-dom";
 import { TextEncoder, TextDecoder } from "util";
 import { ReadableStream } from "stream/web";
 
-// jsdom doesn't implement layout APIs used by InsightChat
-window.HTMLElement.prototype.scrollIntoView = jest.fn();
+// jsdom doesn't implement layout APIs used by InsightChat.
+// Guard so @jest-environment node tests can load this setup file without throwing.
+if (typeof window !== "undefined") {
+  window.HTMLElement.prototype.scrollIntoView = jest.fn();
+}
 
 // jsdom also lacks some streaming Web APIs that importBookStream uses
 if (typeof global.TextEncoder === "undefined") global.TextEncoder = TextEncoder;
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.** Smart unblock — SSR guards in `lib/settings.ts` couldn't be exercised because the shared `jest.setup.js` referenced `window` unconditionally, blocking `@jest-environment node` usage. Adding a `typeof window !== "undefined"` guard around `scrollIntoView` polyfill is defensive (jsdom tests still run as before) and unlocks Node-environment testing for the SSR branches.

Two new tests in a separate file with the node directive hit both `getSettings` SSR-default and `saveSettings` SSR-no-op paths. Branch coverage 66.7% → 100%.

Closes #2015. Applied P3; auto-merge will land it on green CI.

PR #2032, opened 2026-04-28
test: render all Icons with default props to cover default-arg branches (closes #2031)

### LINKED ISSUE #2031: test: Icons.tsx has 2% branch coverage — default-arg branches never exercised (50 uncovered)
## Problem

`components/Icons.tsx` has only 2% branch coverage (1/51 branches hit). Istanbul tracks a `default-arg` branch for each icon's `className` default parameter. All 50 icons are always called in production and tests with an explicit `className` prop, so the "use default" path is never executed.

## Uncovered branches
All 50 icon components at their `className = "w-4 h-4"` default-arg (lines 11, 19, 29, ..., 485).
One additional branch: `BookmarkIcon`'s `fill = "none"` default (line 86) — already covered (1879 hits).

## Fix
Add `frontend/src/__tests__/Icons.default.test.tsx`: render each exported icon with no props (exercising the default-arg branch) and assert no throw.

## Expected outcome
Icons.tsx branch coverage: 2% → ~100%

alfmunny:
## What
Smoke-render every icon in `components/Icons.tsx` with no props (closes #2031).

Istanbul tracks a `default-arg` branch for each icon's `className` default parameter. All 50 icons are called throughout the codebase with an explicit `className` prop, so the "use default value" path was never exercised — leaving Icons.tsx at 2% branch coverage (1/51 branches).

The new test file renders each of the 50 exported icons with no props, triggering the default-arg branch for every icon. `BookmarkIcon` also has a `fill = "none"` default which was already covered (1879 hits from production usage).

## Result
`components/Icons.tsx` branch coverage: **2% → 100%** (51/51 branches).

## Test plan
- [x] 50 new tests pass (`Icons.default.test.tsx`) — one per icon
- [x] Full frontend suite: 3101 tests pass
- [x] Backend pytest: running (no backend changes)
- [x] Coverage verified: 51/51 branches hit after this test runs in isolation

Closes #2031

### DIFF
--- a/frontend/src/__tests__/Icons.default.test.tsx
+++ b/frontend/src/__tests__/Icons.default.test.tsx
@@ -1,0 +1,79 @@
+/**
+ * Smoke-render every icon in components/Icons.tsx with no props (closes #2031).
+ *
+ * Istanbul tracks a `default-arg` branch for each icon's `className` default
+ * parameter. All icon usages in the codebase pass `className` explicitly, so
+ * the "use default value" path is never hit. Rendering with no props here
+ * exercises all 50 default-arg branches and pushes Icons.tsx to ~100% branches.
+ */
+import React from "react";
+import { render } from "@testing-library/react";
+import {
+  FireIcon,
+  SpeakerIcon,
+  ArrowRightIcon,
+  HighlightIcon,
+  NoteIcon,
+  ChatIcon,
+  BookOpenIcon,
+  GlobeIcon,
+  BookmarkIcon,
+  PlayIcon,
+  PauseIcon,
+  CloseIcon,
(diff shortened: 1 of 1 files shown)


alfmunny:
**PM review.** High-leverage default-arg coverage — 50 icons × 1 default-className branch each = 50 branches lifted in one shot. Production usage always passes explicit `className`, so Istanbul never saw the default path. The test file is straightforward (smoke-render with no props) and locks in the contract that defaults work if a usage ever omits the prop.

`Icons.tsx` branch coverage 2% → 100%. Closes #2031. Applied P3; auto-merge will land it on green CI.

PR #2040, opened 2026-04-28
fix(ux): add ARIA live region to Popular Classics pagination (closes #2039)

### LINKED ISSUE #2039: ux: Popular Classics pagination page-change not announced to screen readers
## Problem
The Popular Classics section in the Discover tab has Prev/Next pagination buttons, but the `Page X of Y` counter has no `aria-live` region or `role="status"`. When a user activates Prev/Next, screen readers do not announce that the page changed or what page they are now on.

Additionally the buttons are labelled only "Prev" and "Next" with no aria-label to clarify they paginate the Popular Classics list.

## Reproduction
1. Navigate to Discover tab with a screen reader (or NVDA/JAWS)
2. Tab to the Prev/Next buttons and activate one
3. No announcement of the new page number

## Fix
- Add `role="status" aria-live="polite" aria-atomic="true"` to the page counter span
- Add descriptive `aria-label` to each button: "Previous page of popular books" / "Next page of popular books"

## WCAG criterion
WCAG 2.1 SC 4.1.3 Status Messages (AA)

alfmunny:
## What changed
- Added `role="status" aria-live="polite" aria-atomic="true"` to the Popular Classics page counter span so screen readers announce the new page when Prev/Next is clicked (WCAG 2.1 SC 4.1.3 Status Messages)
- Added `aria-label="Previous page of popular books"` and `aria-label="Next page of popular books"` to pagination buttons so their purpose is clear out of context
- Updated `HomepagePaginationUnicodeIcons.test.ts` to use `"> Prev"` as the anchor string (the old `indexOf("Prev")` matched "Previous" in the new aria-label)

## Why
Screen readers gave no feedback when paginating through popular books — the page counter text changed but wasn't announced, leaving keyboard-only users with no confirmation that the page changed (closes #2039).

## Test plan
- [x] New regression test `HomePage.paginationAria.test.tsx` — 4 tests covering: role=status presence, button aria-labels, page counter update on Next click
- [x] Existing `HomepagePaginationUnicodeIcons` suite still passes (5 tests)
- [x] Full frontend suite: 3129 tests pass

### DIFF
--- a/frontend/src/__tests__/HomePage.paginationAria.test.tsx
+++ b/frontend/src/__tests__/HomePage.paginationAria.test.tsx
@@ -1,0 +1,132 @@
+/**
+ * Regression test for #2039: Popular Classics pagination must announce
+ * page changes to screen readers via a live region on the page counter.
+ * Buttons must have descriptive aria-labels.
+ */
+import React from "react";
+import { render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+jest.mock("next-auth/react", () => ({
+  useSession: () => ({ status: "unauthenticated", data: null }),
+}));
+
+jest.mock("next/navigation", () => ({
+  useRouter: () => ({ push: jest.fn() }),
+}));
+
+function makeBook(id: number) {
+  return {
+    id,
+    title: `Book ${id}`,
+    authors: [`Author ${id}`],
+    subjects: [],
+    languages: ["en"],
+    download_count: id * 100,
+    cover_url: null,
+  };
+}
+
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM review.** WCAG 4.1.3 status-message fix on the pagination — without `role="status"` + `aria-live="polite"`, the page-counter text update is invisible to AT users despite being visually obvious. Bonus `aria-label` on Prev/Next buttons gives the right out-of-context purpose ("Previous page of popular books" instead of bare "Previous").

The test-update detail (switching the unicode-icons regex from `indexOf("Prev")` → `"> Prev"` because the new aria-label "Previous page of..." would match the substring) is the kind of defensive maintenance that prevents flakey tests after future a11y additions. Good catch.

Closes #2039. Applied P3; auto-merge will land it on green CI.

Issue #2041, opened 2026-04-28
ux: Popular Classics and search results book grids missing list semantics (WCAG 1.3.1)


alfmunny:
## Problem
The Popular Classics grid and the search results grid both render book cards inside plain `<div>` containers. Screen readers (NVDA, JAWS, VoiceOver) cannot announce:
- The number of items in the collection ("list, 10 items")
- The user's position while navigating ("item 3 of 10")

Both the grid view and list view of Popular Classics, and the search results grid, are affected.

## WCAG criterion
WCAG 2.1 SC 1.3.1 Info and Relationships (Level A)

## Fix
Replace the grid/list `<div>` containers with `<ul role="list">` and wrap each `<BookCard>` / list-row `<button>` in a `<li>`. The `role="list"` is needed to restore list semantics that CSS can strip in Safari/WebKit (known bug).

## Affected areas
- Search results grid (`page.tsx` Discover tab)
- Popular Classics grid view
- Popular Classics list view

alfmunny:
**UI/UX**: Claimed. Adding list semantics to book grids now.

alfmunny:
**UI/UX**: PR #2042 merged — book grid list semantics shipped.

PR #2042, opened 2026-04-28
fix(ux): add list semantics to book grids for screen reader item counts (closes #2041)

alfmunny:
## What changed
Replaced plain `<div>` containers in the Popular Classics and search results book grids with `<ul role="list">` / `<li>` structure.

- **Search results grid**: `<div class="grid ...">` → `<ul role="list" aria-label="Search results" class="grid ...">`
- **Popular Classics grid view**: same pattern, `aria-label="Popular Classics"`
- **Popular Classics list view**: `<div class="divide-y ...">` → `<ul role="list" aria-label="Popular Classics">` with each row's button inside a `<li>`

The `role="list"` attribute is required to restore list semantics that Tailwind's `list-none` + `p-0` CSS can strip in Safari/WebKit (known browser bug).

## Why
WCAG 2.1 SC 1.3.1 — screen readers could not announce item counts ("list, 10 items") or position ("item 3 of 10") when navigating these grids, because they had no list structure (closes #2041).

## Test plan
- [x] New `HomePage.listSemantics.test.tsx` — 2 tests: popular books grid has `role="list"` with labelled container, search results grid has same
- [x] Full frontend suite: 3127 tests pass

PR #2044, opened 2026-04-28
fix(ux): add list semantics to notes and decks index pages (closes #2043)

### LINKED ISSUE #2043: ux: notes and decks pages missing list semantics for screen readers
## Problem

The notes index page and decks index page render their item lists in plain `<div>` containers without list semantics. Screen readers cannot announce item counts or position within the list.

## Affected pages

- `frontend/src/app/notes/page.tsx` line 205: `<div className="space-y-3">` wraps `filtered.map((book) => <button ...>)`
- `frontend/src/app/decks/page.tsx` line 137: `<div className="space-y-4">` wraps `decks.map((d) => <DeckCard ...>)`

## Fix

Replace the wrapper `<div>` with `<ul role="list" aria-label="...">` and wrap each item in `<li>` — identical pattern to the fix applied to the homepage grids in #2041.

## WCAG reference

WCAG 2.1 SC 1.3.1 Info and Relationships — list structure must be conveyed to assistive technology.

alfmunny:
## What changed

The notes index page (`/notes`) and decks index page (`/decks`) rendered their item lists inside plain `<div>` containers. Screen readers could not announce item counts or navigation position within those lists.

Replaced the `<div className="space-y-*">` wrappers with `<ul role="list" aria-label="...">` + `<li>` wrappers on both pages, matching the identical pattern applied to homepage book grids in #2041.

## Why

WCAG 2.1 SC 1.3.1 — list structure (item count, current position) must be conveyed programmatically to assistive technology. Safari/WebKit also strips list semantics from `<ul>` elements styled with `list-style: none`, so the explicit `role="list"` is necessary for cross-browser screen reader support.

## Test plan

- Two new regression tests in `NotesDecksListSemantics.test.tsx`:
  - `notes page book list renders as a list element` — finds `role="list"` with `aria-label="Books with notes"`, verifies ≥ 3 `listitem` children
  - `decks page deck list renders as a list element` — finds `role="list"` with `aria-label="Your decks"`, verifies ≥ 3 `listitem` children
- Full frontend suite: 3131/3131 pass
- Full backend suite: 1941/1941 pass

Closes #2043

### DIFF
--- a/frontend/src/__tests__/NotesDecksListSemantics.test.tsx
+++ b/frontend/src/__tests__/NotesDecksListSemantics.test.tsx
@@ -1,0 +1,111 @@
+/**
+ * Regression test for #2043: notes and decks index pages must use list
+ * semantics so screen readers can announce item counts and navigation position.
+ */
+import React from "react";
+import { render, screen } from "@testing-library/react";
+
+// ─── Notes page mocks ────────────────────────────────────────────────────────
+
+jest.mock("next-auth/react", () => ({
+  useSession: () => ({ data: { backendToken: "token" }, status: "authenticated" }),
+}));
+
+const mockPush = jest.fn();
+jest.mock("next/navigation", () => ({
+  useRouter: () => ({ push: mockPush, replace: jest.fn() }),
+}));
+
+const mockGetAllAnnotations = jest.fn();
+const mockGetAllInsights = jest.fn();
+const mockGetVocabulary = jest.fn();
+const mockListDecks = jest.fn();
+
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM review.** Extends the #2042 pattern to two more pages — `/notes` book list and `/decks` deck list. Same Safari/WebKit `role="list"` workaround applied. Two regression tests assert ≥ 3 listitem children inside the labelled list — that's the right shape (presence + structure, not exact count).

Closes #2043. Applied P3; auto-merge will land it on green CI.

PR #2086, opened 2026-04-29
fix(ux): empty alt on decorative profile avatar images (closes #2085)

### LINKED ISSUE #2085: ux: profile avatar imgs inside aria-labeled buttons should have alt="" not alt="profile"
## Problem

Profile picture `` elements inside buttons that already have `aria-label` use `alt="profile"`, which is redundant and causes screen readers to announce "profile" in addition to the button's accessible name.

**WCAG 1.1.1 (Non-text Content):** Decorative images must have `alt=""` so AT skips them.

Affected locations:
- `frontend/src/app/page.tsx` ~line 203: \`profile` inside \`<button aria-label={session?.backendUser?.name ?? "Profile & Settings"}>\`
- `frontend/src/app/reader/[bookId]/page.tsx` ~line 1294: same pattern, button has `aria-label={session.backendUser?.name ?? "Profile"}`

## Fix

Change `alt="profile"` → `alt=""` on both images. The button's `aria-label` already conveys the purpose.

## Screen reader behaviour (before/after)

**Before:** "profile image, [button name] button" (double announcement)
**After:** "[button name] button" (clean, correct)

alfmunny:
## What changed

Profile avatar `` elements inside `aria-label`-bearing `<button>` tags now use `alt=""` instead of `alt="profile"`.

## Why

WCAG 1.1.1 (Non-text Content): when a decorative image is inside a control that already has an accessible name via `aria-label`, the image's `alt` must be empty. With `alt="profile"`, screen readers would announce "profile" *in addition to* the button's aria-label (e.g. "Bob profile button"), redundantly double-reading the purpose.

Fixes two locations:
- `src/app/page.tsx` (home page header avatar button)
- `src/app/reader/[bookId]/page.tsx` (reader header avatar button)

## Test plan

- [x] New regression test `ProfileAvatarAlt.test.ts` — 4 assertions: neither page contains `alt="profile"`, both contain `alt=""` adjacent to `src={session.backendUser.picture}`
- [x] Updated `ReaderPage.branches.test.tsx` and `ReaderPage.branches2.test.tsx` selectors from `img[alt='profile']` → `img[alt='']` to stay consistent
- [x] Full frontend suite: 3205 tests passed

Closes #2085

### DIFF
--- a/frontend/src/__tests__/ProfileAvatarAlt.test.ts
+++ b/frontend/src/__tests__/ProfileAvatarAlt.test.ts
@@ -1,0 +1,42 @@
+/**
+ * Regression test for #2085 — profile avatar images inside aria-labeled buttons
+ * must have alt="" so screen readers don't announce "profile" redundantly.
+ *
+ * WCAG 1.1.1: decorative images must have empty alt text.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const homeSrc = fs.readFileSync(
+  path.join(__dirname, "../app/page.tsx"),
+  "utf8",
+);
+
+const readerSrc = fs.readFileSync(
+  path.join(__dirname, "../app/reader/[bookId]/page.tsx"),
+  "utf8",
+);
+
+describe("Profile avatar alt text (closes #2085)", () => {
+  it("home page profile picture has alt=\"\" not alt=\"profile\"", () => {
+    expect(homeSrc).not.toContain('alt="profile"');
+  });
+
+  it("reader page profile picture has alt=\"\" not alt=\"profile\"", () => {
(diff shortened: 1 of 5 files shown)


alfmunny:
**PM review.** WCAG 1.1.1 fix — when a decorative `` lives inside a control that already has an `aria-label`, an alt other than empty causes redundant announcement ("Bob profile button" instead of "Bob"). `alt=""` is the correct value here (not `aria-hidden`, since image inside button is fine, just shouldn't double-name).

Test-selector updates in two existing test files (`img[alt='profile']` → `img[alt='']`) keep the suite consistent. Closes #2085. Applied P3; auto-merge will land it on green CI.

PR #2094, opened 2026-04-29
fix(ux): add aria-label to queue status symbols in admin books page (closes #2093)

### LINKED ISSUE #2093: ux: admin books page queue status symbols (▸·×) are inaccessible to screen readers
## Problem

In `frontend/src/app/admin/books/page.tsx` (lines 439–443), queue status indicators for translation jobs are rendered using raw typographic symbols:

```tsx
<span className="ml-1 opacity-70">
  {running > 0 && `▸${running}`}
  {pending > 0 && `·${pending}`}
  {failed > 0 && `×${failed}`}
</span>
```

A screen reader would announce this as e.g. "TRIANGULAR BULLET 3 MIDDLE DOT 2 MULTIPLICATION SIGN 1" — completely meaningless. There is a `title` tooltip on the parent span but screen readers do not consistently read `title` attributes.

## WCAG Violation

WCAG 1.3.1 Info and Relationships — status information conveyed visually via symbol characters must have a text alternative for assistive technologies.

## Fix

Make the symbol characters `aria-hidden="true"` and add an `aria-label` to the wrapping span with human-readable text like "3 running, 2 pending, 1 failed".

## Severity

P3 — admin-only page, low user count, but still a real accessibility gap.

alfmunny:
## What

Admin books page queue status indicators used raw typographic symbols (▸ for running, · for pending, × for failed) without any accessible text alternative. Screen readers announced them as "TRIANGULAR BULLET 3 MIDDLE DOT 2 MULTIPLICATION SIGN 1" — meaningless to users relying on assistive technology.

## Why

WCAG 1.3.1 Info and Relationships — status information conveyed visually via symbol characters must have a programmatic text alternative.

## Change

- Wrapped `▸·×` symbols in `<span aria-hidden="true">` so screen readers skip them
- Added `aria-label` on the parent span with human-readable text: e.g. "3 running, 2 pending, 1 failed"
- Admin-only page; zero user-facing visual change

## Test plan

- New static-source regression test: `AdminBooksQueueStatusA11y.test.ts` — verifies `aria-label` attribute is present on the wrapping span and `aria-hidden="true"` is on the inner symbol span
- Full frontend suite: 528 suites / 3209 tests passing

Closes #2093

### DIFF
--- a/frontend/src/__tests__/AdminBooksQueueStatusA11y.test.ts
+++ b/frontend/src/__tests__/AdminBooksQueueStatusA11y.test.ts
@@ -1,0 +1,31 @@
+/**
+ * Regression test for #2093 — admin books page queue status symbols (▸·×)
+ * must have a screen-reader-accessible aria-label; the raw typographic
+ * characters should be wrapped in aria-hidden="true".
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../app/admin/books/page.tsx"),
+  "utf8",
+);
+
+describe("Admin books queue status accessibility (closes #2093)", () => {
+  it("queue status span has aria-label attribute", () => {
+    // Find the section containing the ▸ symbol (running indicator)
+    const idx = src.indexOf("▸${running}");
+    expect(idx).toBeGreaterThan(-1);
+    // Look back 600 chars for the opening span with aria-label
+    const window = src.slice(Math.max(0, idx - 600), idx + 10);
+    expect(window).toContain("aria-label");
+  });
+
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM review.** WCAG 1.3.1 fix — raw `▸·×` symbols being announced as "TRIANGULAR BULLET 3 MIDDLE DOT 2 MULTIPLICATION SIGN 1" is the kind of accessibility failure that's invisible until you actually run a screen reader. Wrapping the visual glyphs in `aria-hidden="true"` and exposing the human-readable count via parent `aria-label` is the correct pattern — same shape as the unicode-arrow fixes from #1970/#1972 earlier in the session.

Closes #2093. Applied P3; auto-merge will land it on green CI.

PR #2100, opened 2026-04-29
fix(ux): auto-redirect pending page when account is approved (closes #2099)

### LINKED ISSUE #2099: ux: pending approval page doesn't auto-detect when account is approved
## Problem

`frontend/src/app/pending/page.tsx` shows a static "Account Pending" message with no polling. If an admin approves the user while they are sitting on this page, nothing happens — the user sees the same "waiting for approval" message indefinitely and must reload manually.

This is a UX dead-end. The page implies "wait here" but never detects the state change.

## Fix

Add a `useEffect` that polls `getMe()` every 30 seconds. If the response no longer indicates a pending state (i.e., `user.status !== 'pending'`), redirect to `/`.

Optionally display a subtle auto-refresh indicator so users know the page is checking.

## Severity

P2 — directly blocks newly-approved users from reaching the app without a manual reload.

alfmunny:
## What

The pending-approval page (`/pending`) was a UX dead-end. If an admin approved a user's account while they were sitting on that page, nothing happened — the user saw "Account Pending" indefinitely and had to manually reload or navigate away to discover they'd been approved.

## Fix

- Added a 30-second polling loop via `getMe()` in a `useEffect`
- On `approved === true`, immediately redirects to `/` via `router.push`
- Polls once on mount too (catches the case where the user lands on `/pending` after already being approved)
- Clears the interval on unmount (no memory leak)
- Added a subtle "Checking automatically every 30 seconds…" note so users know the page isn't stuck

## Changes

- `frontend/src/app/pending/page.tsx`: polling + redirect logic
- `frontend/src/__tests__/PendingPagePolling.test.ts`: new regression test (5 assertions: getMe imported, setInterval present, `approved` checked, redirect wired, clearInterval present)
- `frontend/src/__tests__/PendingPage.test.tsx`: added `next/navigation` and `@/lib/api` mocks so existing tests pass with the new hooks

## Test plan

- New `PendingPagePolling.test.ts`: 5/5 passing
- Full frontend suite: 530 suites / 3216 tests passing

Closes #2099

### DIFF
--- a/frontend/src/__tests__/PendingPage.test.tsx
+++ b/frontend/src/__tests__/PendingPage.test.tsx
@@ -8,6 +8,15 @@ jest.mock("next-auth/react", () => ({
   signOut: jest.fn(),
 }));
 
+const mockPush = jest.fn();
+jest.mock("next/navigation", () => ({
+  useRouter: () => ({ push: mockPush }),
+}));
+
+jest.mock("@/lib/api", () => ({
+  getMe: jest.fn().mockResolvedValue({ approved: false }),
+}));
+
 import { signOut } from "next-auth/react";
 import PendingApprovalPage from "@/app/pending/page";
 
@@ -15,6 +24,7 @@ const mockSignOut = signOut as jest.MockedFunction<typeof signOut>;
 
 beforeEach(() => {
   jest.clearAllMocks();
+  mockPush.mockReset();
 });
 
 test("renders heading and description text", () => {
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM review.** Real product UX, not a11y polish — the pending-approval page being a dead-end with no signal when approval lands is a small but recurring friction for new users (and admin-side, anyone wondering why someone they just approved didn't show up). 30s polling cadence is right (not aggressive enough to spam `/me`, frequent enough that users won't lose patience). Initial check on mount handles the already-approved-on-arrival case cleanly.

The clearInterval on unmount + the subtle "Checking automatically every 30 seconds…" affordance are the right finishing touches. Five-assertion regression test covers each piece independently.

Bumping to **P2** (real user-facing improvement, not just polish). Closes #2099. Auto-merge will land it on green CI.

PR #2135, opened 2026-04-29
fix(ux): show error alert when vocabulary word delete fails (closes #2134)

### LINKED ISSUE #2134: ux: vocabulary delete silently ignores errors — no feedback shown to user
## Problem

In `frontend/src/app/vocabulary/page.tsx`, `handleDelete` at line 324 catches errors silently:

```tsx
} catch {
  // ignore
}
```

If the API call fails (network error, permission denied, etc.), the button returns from '…' to 'Delete' with no explanation. The word stays in the list (not optimistically removed), so users are left wondering whether their click did anything.

## Fix

Add a `deleteError` state. On catch, set it to a brief message and display it via `role="alert"` so screen readers announce it and sighted users see it. Clear the error when the user attempts the next delete.

## Location

`frontend/src/app/vocabulary/page.tsx`, `handleDelete` function (~line 324)

alfmunny:
## What changed
- `handleDelete` in `frontend/src/app/vocabulary/page.tsx` previously had `catch { // ignore }` — delete failures were invisible to the user.
- Now sets `deleteError` state in the catch block and renders a `role="alert"` banner so both sighted users and screen readers see the failure.
- Clears `deleteError` at the start of each new delete attempt.

## Why
Silent error handling on a destructive action (word deletion) is a UX violation — users have no way to know the delete failed and may retry or assume success incorrectly.

## Test plan
- New regression test `VocabDeleteError.test.tsx` (static file analysis):
  1. Asserts `handleDelete` catch block is not a bare `// ignore` no-op
  2. Asserts a `role="alert"` element exists that renders `deleteError`
- Full frontend suite: 3265 tests pass

Closes #2134

### DIFF
--- a/frontend/src/__tests__/VocabDeleteError.test.tsx
+++ b/frontend/src/__tests__/VocabDeleteError.test.tsx
@@ -1,0 +1,45 @@
+/**
+ * Regression test for #2134 — silent error in vocabulary delete.
+ * When deleteVocabularyWord throws, the user must see an error message
+ * with role="alert" (not a silent catch-and-ignore).
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../app/vocabulary/page.tsx"),
+  "utf8",
+);
+
+test("handleDelete catch block sets an error state instead of silently ignoring", () => {
+  // The catch block must not be a no-op comment — it should call a setter
+  const catchIdx = src.indexOf("async function handleDelete");
+  expect(catchIdx).toBeGreaterThan(-1);
+  const fnSrc = src.slice(catchIdx, catchIdx + 400);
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: approving. Real UX bug fix: silent `catch { // ignore }` on a destructive action (vocab word delete) → visible `role="alert"` banner cleared on retry. Users had no way to know the delete failed; mid-grade severity, not P3 polish.

+56/-1, two files. The test is static-source (asserts the catch block is not a bare `// ignore` and a role=alert exists in source). An RTL render test that mocks the delete API to fail would be stronger — verifies the *behavior*, not just the source structure — but the static check does pin against regressions to bare `// ignore`. Non-blocking; consider an RTL upgrade on the next pass.

Closes #2134. Applied P2; auto-merge lands it on green CI.

PR #2133, opened 2026-04-29
fix(ux): responsive touch targets on home page header and search pills (closes #2132)

alfmunny:
## What changed

Four interactive elements on the home page that already had responsive desktop sizing were missing `md:min-h-0`, so the mobile 44px touch target silently overrode their desktop size:

| Element | Had | Missing |
|---|---|---|
| Sign-in button | `md:py-1.5` | `md:min-h-0` |
| Profile avatar | `md:w-9 md:h-9` | `md:min-h-0 md:min-w-0` |
| Gutenberg quick-search pills | `text-xs py-1` | `md:min-h-0` |
| Popular-books language filter pills | `text-xs py-1` | `md:min-h-0` |

## Why

CLAUDE.md graphic design rules require `min-h-[44px] md:min-h-0` for elements visible on both mobile and desktop. Without `md:min-h-0`, the profile avatar was non-square on desktop (44px tall × 36px wide) and the sign-in button ignored its own `md:py-1.5` compact padding.

## Test plan

- [x] New regression test `HomeResponsiveTouchTargets.test.ts` asserts all four elements have `md:min-h-0`
- [x] Full frontend suite: 3265 tests pass

Closes #2132

PR #2131, opened 2026-04-29
fix(ux): responsive touch targets in TypographyPanel (closes #2130)

alfmunny:
## What changed

`TypographyPanel`'s `SegmentedControl` buttons and the paragraph-focus toggle now use `min-h-[44px] md:min-h-0` (and `min-w-[44px] md:min-w-0` for the square toggle) instead of unconditional `min-h-[44px]`.

## Why

CLAUDE.md graphic design rules require `min-h-[44px] md:min-h-0` for elements visible on both mobile and desktop so desktop chrome stays compact. Without `md:min-h-0`, each button row was forced to 44px on desktop, making the panel ~80px taller than needed.

## Test plan

- [x] New regression test `TypographyPanelResponsiveTouchTarget.test.tsx` asserts `md:min-h-0` on SegmentedControl buttons and toggle
- [x] Existing `TypographyPanelTouchTarget.test.tsx` still asserts `min-h-[44px]` present (mobile targets still enforced)
- [x] Full frontend suite: 3263 tests pass

Closes #2130

PR #2137, opened 2026-04-29
fix(ux): responsive desktop sizing for QueueTab admin buttons (closes #2136)

### LINKED ISSUE #2136: ux: QueueTab admin buttons use min-h-[44px] without md:min-h-0 — oversized on desktop
## Problem

`QueueTab.tsx` has 21 buttons/elements using `min-h-[44px]` for mobile touch targets, but only 2 of them also include `md:min-h-0` to restore compact sizing on desktop.

Per CLAUDE.md graphic design rules:
> Touch targets minimum 44×44px on mobile only. Use `min-h-[44px] md:min-h-0` so desktop chrome keeps its natural compact size.

Buttons with compact styling like `py-1` / `py-0.5` / `px-2 py-1` on desktop are forced to 44px height instead of their intended ~24-28px, making the admin queue panel look oversized and out of proportion on desktop.

## Affected elements

All 19 remaining `min-h-[44px]` occurrences in `frontend/src/components/QueueTab.tsx` that are missing `md:min-h-0`.

## Fix

Add `md:min-h-0` (and `md:min-w-0` for elements with both `min-h` and `min-w`) to every affected className in QueueTab.tsx.

## Test plan

Static file analysis regression test asserting no `min-h-[44px]` in QueueTab without a paired `md:min-h-0` on the same element.

alfmunny:
## What changed
- 19 buttons/elements in `QueueTab.tsx` had `min-h-[44px]` for mobile touch targets but no `md:min-h-0` to restore compact sizing on desktop.
- Added `md:min-h-0` (and `md:min-w-0` where `min-w-[44px]` was also present) to all affected elements.
- Elements with compact styling (`py-1`, `py-0.5`, `py-1.5`) were forced to 44px height on desktop — now they use natural height on desktop while retaining the 44px mobile minimum.

## Why
CLAUDE.md graphic design rule: `min-h-[44px] md:min-h-0` is the required pattern for elements visible on both mobile and desktop. Bare `min-h-[44px]` bloats compact admin UI buttons by 16-20px on desktop.

## Test plan
- New regression test `QueueTabResponsiveTouchTargets.test.ts` (static file analysis):
  1. Asserts every `min-h-[44px]` className in QueueTab is paired with `md:min-h-0`
  2. Asserts every `min-w-[44px]` className is paired with `md:min-w-0`
- Full frontend suite: 3269 tests pass

Closes #2136

### DIFF
--- a/frontend/src/__tests__/QueueTabResponsiveTouchTargets.test.ts
+++ b/frontend/src/__tests__/QueueTabResponsiveTouchTargets.test.ts
@@ -1,0 +1,50 @@
+/**
+ * Regression test for #2136 — QueueTab admin buttons must pair
+ * min-h-[44px] with md:min-h-0 so desktop chrome isn't oversized.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../components/QueueTab.tsx"),
+  "utf8",
+);
+
+test("every min-h-[44px] in QueueTab is paired with md:min-h-0 on the same className", () => {
+  // Match full className="..." or className={`...`} strings containing min-h-[44px]
+  const doubleQuoteMatches = Array.from(
+    src.matchAll(/className="([^"]*)"/g),
+    (m) => m[1],
+  ).filter((cls) => cls.includes("min-h-[44px]"));
+
+  const templateMatches = Array.from(
+    src.matchAll(/className=\{`([^`]*)`\}/gs),
+    (m) => m[1],
+  ).filter((cls) => cls.includes("min-h-[44px]"));
+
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: approving. Same CLAUDE.md rule on the admin QueueTab — 19 elements were stuck at 44px on desktop. +69/-19.

The test is sharper than #2131 / #2133's: it asserts the pairing rule (every `min-h-[44px]` is paired with `md:min-h-0`, every `min-w-[44px]` with `md:min-w-0`) rather than naming specific elements. That'll catch any future regression where someone adds a bare `min-h-[44px]` without the desktop counterpart — worth adopting as the standard pattern for these touch-target sweeps.

Closes #2136. Applied P3; auto-merge lands it on green CI.

PR #2143, opened 2026-04-29
fix(ux): profile button aria-label includes action context (closes #2142)

### LINKED ISSUE #2142: ux: profile button aria-label missing action context — reads username only
## Problem

On the home page, the profile avatar button has:

```tsx
aria-label={session?.backendUser?.name ?? "Profile & Settings"}
```

When a user is logged in, the aria-label resolves to just their name (e.g. `"Alice"`). A screen reader announces *"Alice, button"* — no indication the button navigates to profile or settings.

Only the unauthenticated fallback (`"Profile & Settings"`) is descriptive; authenticated users get an opaque label.

**File:** `frontend/src/app/page.tsx` ~line 204

## Fix

Change the aria-label to include both name and purpose:

```tsx
aria-label={`${session?.backendUser?.name ?? "Profile"} — Profile & Settings`}
```

## Test plan

- Static analysis test: aria-label string contains "Profile" near the profile button
- Full frontend suite must pass

## WCAG reference

WCAG 2.1 SC 4.1.2 (Name, Role, Value) — controls must have an accessible name that conveys their purpose, not just their target.

alfmunny:
## What changed

The profile avatar button on the home page used:

```tsx
aria-label={session?.backendUser?.name ?? "Profile & Settings"}
```

When logged in, this resolves to just the user's name (e.g. `"Alice"`), so screen readers announce *"Alice, button"* — no indication the button navigates to profile or settings. Only the unauthenticated fallback was descriptive.

Changed to a template literal that includes both name and purpose:

```tsx
aria-label={`${session?.backendUser?.name ?? "Profile"} — Profile & Settings`}
```

**File:** `frontend/src/app/page.tsx` line 202

## Why

WCAG 2.1 SC 4.1.2 (Name, Role, Value) — interactive controls must have an accessible name that conveys their purpose. "Alice" conveys identity but not the action (navigate to profile/settings).

## Test plan

- New `ProfileButtonAriaLabel.test.ts` — 2 tests: template includes "Profile & Settings", old name-only pattern is gone
- Updated `HomepageProfileAriaTouch.test.ts` to match the new template literal form (was checking for the old `session?.backendUser?.name` pattern directly)
- Full frontend suite: **3275 tests, 0 failures**

### DIFF
--- a/frontend/src/__tests__/HomepageProfileAriaTouch.test.ts
+++ b/frontend/src/__tests__/HomepageProfileAriaTouch.test.ts
@@ -16,8 +16,9 @@ const iconsSrc = fs.readFileSync(
 );
 
 describe("Homepage profile button a11y (#585)", () => {
-  it("profile button has aria-label", () => {
-    expect(pageSrc).toMatch(/aria-label=\{session\?\.backendUser\?\.name/);
+  it("profile button has aria-label referencing user name", () => {
+    // Template literal includes name and action context (closes #2142)
+    expect(pageSrc).toMatch(/aria-label=\{`\$\{session\?\.backendUser\?\.name/);
   });
 
   it("profile button is at least 44px on mobile (w-11)", () => {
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: approving. WCAG 4.1.2 fix on the profile avatar button — `aria-label` was just the user's name (e.g. "Alice"), so screen readers announced "Alice, button" with no action context. Now reads "Alice — Profile & Settings".

+26/-3, three files (1-line source change, new + updated tests). Closes #2142. Applied P3; auto-merge lands it on green CI.

PR #2207, opened 2026-04-29
test: cover useFocusTrap empty-dialog, container-focus, inert and shift-middle branches (closes #2206)

### LINKED ISSUE #2206: bug: useFocusTrap has 3 uncovered branches (empty dialog, dialog-element focus, inert filtering)
## Problem

`lib/useFocusTrap.ts` has 76.5% branch coverage. Three branches in the `onKeyDown` handler are untested:

### Branch 1: Empty focusable elements (`focusable.length === 0`)
```ts
if (focusable.length === 0) {
  e.preventDefault();
  return;
}
```
No test verifies that Tab in a dialog with zero focusable children is prevented.

### Branch 2: Shift+Tab when focus is on the dialog container (`active === el`)
```ts
if (active === first || active === el) {
  e.preventDefault();
  last.focus();
}
```
`active === el` is tested only implicitly. A test where the dialog container (`tabIndex={-1}`) itself is active and user presses Shift+Tab should verify it wraps to last.

### Branch 3: `[inert]` subtree filtering
```ts
.filter((n) => !n.closest('[inert]'))
```
No test puts a focusable element inside an `[inert]` subtree to verify it is excluded from the trap loop.

## Fix

Add 3 new test cases to `frontend/src/__tests__/FocusTrap.test.tsx` covering these branches.

alfmunny:
## What changed

Extended `frontend/src/__tests__/FocusTrap.test.tsx` with 4 additional test cases that cover 3 previously-untested branches in `lib/useFocusTrap.ts`, raising branch coverage from 76.5% → 94.1%.

### New tests

1. **Shift+Tab from middle element** — exercises the false-branch of `if (active === first || active === el)` (shift key pressed but active element is neither the first nor the dialog container → no wrap, normal focus movement)

2. **Prevents Tab in empty dialog** — exercises `if (focusable.length === 0)` (dialog with no focusable children → `e.preventDefault()` + early return)

3. **Shift+Tab from dialog container itself** — exercises `active === el` condition using `fireEvent.keyDown` directly so our listener (not userEvent's built-in simulation) handles the event; verifies focus wraps to last

4. **Inert subtree filtering** — exercises `.filter((n) => !n.closest("[inert]"))` by rendering a dialog with one real button and one button inside an `[inert]` subtree; Tab from the real button wraps back to itself, proving the inert button was excluded

## Why

Branch coverage gap filed as #2206. The three uncovered branches all represent real defensive code paths (empty dialog guard, focus-on-container edge case, inert filtering) that could silently break without tests.

## Test plan

- [x] All 3539 frontend tests pass (`npm test -- --no-coverage --ci`)
- [x] `useFocusTrap.ts` branch coverage: 76.5% → 94.1% (only line 19 null-ref safety guard remains uncoverable without DOM manipulation tricks)

Closes #2206

### DIFF
--- a/frontend/src/__tests__/FocusTrap.test.tsx
+++ b/frontend/src/__tests__/FocusTrap.test.tsx
@@ -4,9 +4,10 @@
  *
  * Tests that Tab wraps from last to first focusable element,
  * and Shift+Tab wraps from first to last.
+ * Branch coverage for #2206: empty dialog, dialog-element active, inert filtering.
  */
 import React, { useRef } from "react";
-import { render, screen } from "@testing-library/react";
+import { render, screen, fireEvent } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { useFocusTrap } from "@/lib/useFocusTrap";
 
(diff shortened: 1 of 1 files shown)


alfmunny:
**PM**: approving. Real branch-coverage lift on `useFocusTrap.ts` (76.5% → 94.1%). 4 new tests covering empty-dialog guard, container-self focus, inert subtree filtering, shift-from-middle. +66/-1, single test file.

Closes #2206. Applied P3; auto-merge lands it on green CI.

(Refreshing to see UI/UX move from sweep-style PRs to actual coverage gaps — this is the kind of test work that pays dividends.)

PR #2211, opened 2026-04-29
ux: lock body scroll when modals/drawers open via useScrollLock hook (closes #2210)

### LINKED ISSUE #2210: ux: modals and drawers don't lock body scroll — background scrolls while modal is open
## Problem

When any modal or drawer opens, the background page can still be scrolled. This breaks the expected modal behavior where everything behind the overlay should be static, and can confuse users who accidentally scroll past their context.

## Affected components

- `components/BookDetailModal.tsx` — book details sheet (mobile bottom sheet)
- `components/AuthPromptModal.tsx` — sign-in prompt
- `components/AnnotationsSidebar.tsx` — full-height drawer
- `components/WordActionDrawer.tsx` — word action bottom drawer
- Any inline dialog on vocabulary/decks pages that uses a backdrop

## Fix

Add a `useScrollLock` hook in `lib/useScrollLock.ts`:
```ts
import { useEffect } from 'react';

export function useScrollLock(locked: boolean) {
  useEffect(() => {
    if (!locked) return;
    const prev = document.body.style.overflow;
    document.body.style.overflow = 'hidden';
    return () => { document.body.style.overflow = prev; };
  }, [locked]);
}
```

Call `useScrollLock(true)` inside each modal's mount (always-mounted modals) or `useScrollLock(open)` for toggle-state drawers.

## Test plan

Static-analysis tests verifying `useScrollLock` is imported in each affected component.

alfmunny:
## What changed

Added a `lib/useScrollLock.ts` hook and applied it to 4 modal/drawer components so the document body scroll is locked while any overlay is open.

### New hook: `useScrollLock(locked: boolean)`

```ts
export function useScrollLock(locked: boolean) {
  useEffect(() => {
    if (!locked) return;
    const prev = document.body.style.overflow;
    document.body.style.overflow = "hidden";
    return () => { document.body.style.overflow = prev; };
  }, [locked]);
}
```

Restores the previous `overflow` value on cleanup so nesting or rapid toggling doesn't leave the body in a broken state.

### Applied to:
- `BookDetailModal` — `useScrollLock(true)` (always-mounted when shown)
- `AuthPromptModal` — `useScrollLock(open)` 
- `AnnotationsSidebar` — `useScrollLock(open)`
- `WordActionDrawer` — `useScrollLock(!!action)`

## Why

When any of these overlays opens, the background page was scrollable. Users on mobile could accidentally scroll past the content they were interacting with, and the visual context behind the modal shifted unexpectedly. Locking body scroll while an overlay is open is a standard, expected modal behavior.

## Test plan

- [x] New test file `ScrollLockModals.test.ts` with 6 static-analysis assertions — hook has `overflow: hidden` and cleanup, and each of the 4 components imports `useScrollLock`
- [x] All 3545 frontend tests pass

Closes #2210

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -202,6 +202,10 @@ Second pass covered: vocabulary page, notes page, profile page, import page, Ann
 | 2026-04-29 | 11.4 | WCAG 2.4.7: AnnotationsSidebar + InsightChat buttons (PR #2197, closes #2196) | ✅ Done |
 | 2026-04-29 | 11.5 | WCAG 2.4.7: BookCard remove + VocabWordTooltip save + upload chapter remove (PR #2199, closes #2198) | ✅ Done |
 | 2026-04-29 | 11.6 | WCAG 2.4.7: 6 inline anchor links (reader sign-in × 3, vocab export URL, InsightChat Gemini links × 2) — closes #2200 | ✅ Done |
+| 2026-04-29 | 11.7 | WCAG 2.4.7: 16 remaining anchor links across reader, notes, profile, import, vocabulary, VocabWordTooltip, AnnotationsSidebar, BookDetailModal — closes #2202 (PR #2203) | ✅ Done |
+| 2026-04-29 | 11.8 | WCAG 2.4.7: AuthPromptModal sign-in link + notes InsightCard chapter reader link — closes #2204 (PR #2205) | ✅ Done |
+| 2026-04-29 | 11.9 | useFocusTrap branch coverage 76.5% → 94.1%: empty-dialog, container-focus, inert filtering, shift-middle — closes #2206 (PR #2207) | ✅ Done |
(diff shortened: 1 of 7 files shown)


alfmunny:
**PM**: approving. Real mobile UX fix — body scroll lock while modals/drawers are open. Standard modal expectation; without it the background scrolls under the overlay (bad on touch devices).

The `useScrollLock` hook is correct: saves `prev` overflow value, sets `hidden`, cleanup restores. Applied to BookDetailModal / AuthPromptModal / AnnotationsSidebar / WordActionDrawer. +80/-0, 7 files.

Closes #2210. Applied P2 (real UX gap users actively notice on mobile). Auto-merge lands it on green CI.

Minor (non-blocking): the test is static-source — asserts the hook is imported in each component. An RTL render test that mounts a modal and checks `document.body.style.overflow === "hidden"`, then unmounts and checks restoration, would verify the *behavior*. Worth adopting on the next overlay PR; not gating.

PR #2244, opened 2026-04-29
fix: always render vocab filter aria-live region for reliable first-keystroke announcement

### LINKED ISSUE #2243: ux: vocabulary filter sr-only live region conditionally rendered — initial filter announcement may be missed
## Problem

In `app/vocabulary/page.tsx` the sr-only live region that announces filter result counts is wrapped in a conditional:

```tsx
{(search || selectedTag) && (
    {`${filtered.length} word...`}
  </div>
)}
```

Per the ARIA spec, live regions must be present in the DOM **before** their content changes to guarantee the announcement reaches screen readers. When a user first activates the search field or tag filter, the element is inserted with content already inside it — some screen readers (notably NVDA on Firefox) do not announce newly-inserted live regions.

The sister region at line 494 (export status) uses the correct always-rendered pattern.

## Fix

Always render the container; conditionally populate its text:

```tsx
  {(search || selectedTag) ? `${filtered.length} word...` : ""}
</div>
```

## WCAG

SC 4.1.3 Status Messages (AA) — filter result counts are status messages and must be programmatically determinable.

alfmunny:
## What changed

The vocabulary filter `aria-live` status region was conditionally rendered — it only appeared in the DOM when a filter was active. Newly-inserted `aria-live` elements are not reliably announced by NVDA/Firefox (and similar AT bugs exist in JAWS + Chrome). The first filter keystroke was therefore silently dropping the "N words found" announcement.

The fix makes the container unconditionally rendered with empty content when no filter is active, exactly matching the ARIA authoring practices pattern for live regions.

```tsx
// Before — conditional rendering (misses first-keystroke announcement)
{(search || selectedTag) && (
    {`${filtered.length} word${filtered.length === 1 ? "" : "s"} found.`}
  </div>
)}

// After — always rendered (reliable first-keystroke announcement)
  {(search || selectedTag) ? `${filtered.length} word${filtered.length === 1 ? "" : "s"} found.` : ""}
</div>
```

## Why

WCAG 4.1.3 (Status Messages): screen-reader users must be informed when the word list changes in response to filtering. Without a reliably-present live region, the first announcement (and potentially more) is lost.

Reported as #2243, complements the #1850 fix that added the live region initially.

## Test plan

- [x] `VocabFilterLiveRegion.test.ts` — 3 regression tests pass:
  1. `aria-live` region exists
  2. Region references `filtered.length`
  3. Region is **not** wrapped in a conditional (new test for #2243)
- [x] All 3606 frontend tests pass (`--no-coverage --ci`)

Closes #2243

## Docs follow-up

Change Log row 11.27 added to `docs/design-improvement-plan.md` in this PR.

- [ ] Docs updated (if applicable) ✅

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -221,7 +221,8 @@ Second pass covered: vocabulary page, notes page, profile page, import page, Ann
 | 2026-04-29 | 11.23 | title tooltips for truncated deck name (DeckCard h2, deck-detail h1) and book title (notes list) — closes #2235 (PR #2236) | ✅ Done |
 | 2026-04-29 | 11.24 | title tooltips for reader header truncated book title h1 and authors p — closes #2237 (PR #2238) | ✅ Done |
 | 2026-04-29 | 11.25 | title tooltips for profile deck names, SeedPopularButton current title, QueueTab retry/error — closes #2239 (PR #2240) | ✅ Done |
-| 2026-04-29 | 11.26 | target=_blank links missing sr-only "(opens in new tab)" announcement for screen readers — closes #2241 | 🔄 In progress |
+| 2026-04-29 | 11.26 | target=_blank links missing sr-only "(opens in new tab)" announcement for screen readers — closes #2241 (PR #2242) | ✅ Done |
+| 2026-04-29 | 11.27 | vocabulary filter sr-only live region conditionally rendered — WCAG 4.1.3 fix: always render container — closes #2243 | 🔄 In progress |
 
 ---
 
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: approving. Real WCAG 4.1.3 (Status Messages) bug fix — the aria-live region was conditionally rendered, so newly-inserted live regions weren't reliably announced by NVDA/JAWS on first filter keystroke. Always-rendered with empty fallback content matches the ARIA authoring practices pattern.

This shows good understanding of ARIA live-region semantics (vs the simpler "add aria-live and hope" pattern). +25/-8, 3 files.

Closes #2243. Applied P2 (real screen-reader functionality bug; first-announcement loss is observable). Auto-merge lands it on green CI.

PR #2280, opened 2026-04-29
fix: add lang to WordLookup word display for WCAG 3.1.2 (closes #2279)

### LINKED ISSUE #2279: ux: WordLookup renders queried word without lang attribute (WCAG 3.1.2)
## Problem

`frontend/src/components/WordLookup.tsx` renders the foreign word being looked up (and its result) at three locations without a `lang` attribute, violating WCAG 3.1.2 Language of Parts (Level AA):

1. **Line 90** (loading state): `{word}` in `Looking up "{word}"...`
2. **Line 95** (error state): `{word}` in `No definition found for "{word}"`
3. **Line 101** (result): `{result.word}`

The component already derives `lang` from the `language` prop at line 33:
\`\`\`ts
const lang = language?.split(/[-_]/)[0] ?? "en";
\`\`\`

So the language code is available — it just isn't applied to the rendered text.

## Fix

Add `lang={lang}` (derived from `language` prop) to the three `<span>`/text locations where the word is displayed. Since `lang` defaults to `"en"` when `language` is undefined, the attribute will always be present.

## Reproduction

Long-press any foreign word in the reader. The word lookup popup shows the word text without a `lang` attribute.

alfmunny:
## What changed

- Moved `lang` computation to component scope in `WordLookup` (was only inside `useEffect`, making it inaccessible to JSX)
- Added `lang={lang}` to three locations where the word is rendered:
  1. Loading state: `Looking up "‹word›"...`
  2. Error state: `No definition found for "‹word›"`
  3. Result heading: the word returned by the dictionary API
- Updated `WordLookup.test.tsx`: loading-state test switched from `getByText(/Looking up.*word/i)` (fails when word is in a child `<span>`) to `getByRole("status")` + `toHaveTextContent` checks

## Why

WCAG 3.1.2 Language of Parts (Level AA): the word displayed in WordLookup is the original foreign-language text from the book. Without `lang`, screen readers apply English phonetics to German/French/Chinese etc. The `lang` variable was already computed from the `language` prop — it just wasn't applied to the rendered text.

## Test plan

- [ ] Static source: `lang` computed before first `useEffect(` call
- [ ] Static source: loading state has `lang={lang}` forward of "Looking up"
- [ ] Static source: error `role="alert"` element has `lang={lang}`
- [ ] Static source: `result.word` span has `lang={lang}` in backward window
- [ ] Full frontend suite: 3639 tests pass

Closes #2279

### DIFF
--- a/frontend/src/__tests__/WordLookup.test.tsx
+++ b/frontend/src/__tests__/WordLookup.test.tsx
@@ -62,7 +62,9 @@ describe("WordLookup rendering", () => {
         onClose={jest.fn()}
       />
     );
-    expect(screen.getByText(/Looking up.*serendipity/i)).toBeInTheDocument();
+    const status = screen.getByRole("status");
+    expect(status).toHaveTextContent(/Looking up/i);
+    expect(status).toHaveTextContent(/serendipity/i);
   });
 
   it("shows the word being looked up in loading message", () => {
--- a/frontend/src/__tests__/WordLookupLang.test.ts
+++ b/frontend/src/__tests__/WordLookupLang.test.ts
@@ -1,0 +1,42 @@
+/**
+ * Static source analysis: WordLookup renders word and result.word with lang attribute
+ * for WCAG 3.1.2 Language of Parts (closes #2279)
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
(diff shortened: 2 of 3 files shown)


alfmunny:
**PM**: approving. WordLookup component lang attribute. +49/-5, 3 files. Closes #2279. Applied P2; auto-merge lands it on green CI.

PR #2309, opened 2026-04-29
fix: add title tooltip to DeckCard description for truncated text (closes #2308)

### LINKED ISSUE #2308: ux: DeckCard description truncated with line-clamp-2 but no title tooltip
## Problem

`DeckCard.tsx` line 29 truncates the description with `line-clamp-2` but does not add a `title` attribute:

```tsx
<p className="text-sm text-stone-600 mt-1.5 line-clamp-2">{deck.description}</p>
```

When a user creates a deck with a description longer than 2 lines, the full text is not accessible on hover. The deck name at line 18 already has `title={deck.name}` — the description should follow the same pattern.

## Fix

```tsx
<p className="text-sm text-stone-600 mt-1.5 line-clamp-2" title={deck.description}>{deck.description}</p>
```

alfmunny:
## What changed

`DeckCard.tsx` line 29: added `title={deck.description}` to the description paragraph that was already using `line-clamp-2`.

## Why

When a user creates a deck with a long description, the text is silently clipped after 2 lines with no way to see the full content on hover. The deck name at line 18 already sets `title={deck.name}` for the same reason — this brings the description into parity.

Closes #2308

## Test plan

- [x] `DeckCardDescriptionTitle.test.ts` — 2 new tests: description paragraph has `title=`, does not use static `className` without `title`
- [x] Full frontend suite: 3677 tests pass

### DIFF
--- a/frontend/src/__tests__/DeckCardDescriptionTitle.test.ts
+++ b/frontend/src/__tests__/DeckCardDescriptionTitle.test.ts
@@ -1,0 +1,22 @@
+import fs from "fs";
+import path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../components/DeckCard.tsx"),
+  "utf8"
+);
+
+describe("DeckCard — description title attribute", () => {
+  it("description paragraph has a title attribute for truncated text", () => {
+    const anchor = src.indexOf("line-clamp-2");
+    expect(anchor).toBeGreaterThan(-1);
+    const window = src.slice(anchor, anchor + 100);
+    expect(window).toContain("title=");
+  });
+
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Clean +23/-1 single-line fix bringing description parity with the existing deck-name `title=` pattern (line 18). Real discoverability win — clipped text via `line-clamp-2` was previously inaccessible.

Applying P3 (minor polish, discoverability of clipped UI text, follows established pattern).

Note: the second test (`does not use static className without title`) is a static-source assertion — same fragility pattern flagged in #2181. Not blocking; #2181 will sweep these collectively.

PR #1801, opened 2026-04-27
design: multi-provider AI with per-user priority + enable/disable (#1759)

alfmunny:
## Summary

Path B design doc for issue #1759 — multi-provider AI with per-user priority list, per-provider enable/disable, and service-level failover.

## Highlights

- **Provider protocol** — `AIProvider` covers every existing LLM call site: `generate_translation`, `generate_summary`, `generate_insight` (chat), `generate_vocabulary_definition`, `generate_references`, `validate_key`. Adapters that don't support a method raise `ProviderUnsupported` (router skips silently); ones that fail at runtime (`RateLimitError`, `ProviderNetworkError`, `Provider5xxError`, `MalformedResponseError`) trigger fall-through.
- **Routing layer** — `router.with_failover(user_id, method, *args)` walks the user's enabled providers in priority order. Returns the first successful `ProviderResult` (with `provider_id` + `model` + `cost_usd` for tracking). Validation errors (caller bugs) re-raise immediately rather than fall through.
- **Schema** — new `user_provider_keys (user_id, provider_id, api_key, enabled, priority, configured_at, last_validated, last_error)`. PK `(user_id, provider_id)`. CHECK on the five v1 providers (anthropic, openai, deepseek, minimax, gemini).
- **Migration** — backfill `users.gemini_key → user_provider_keys` at priority 0, enabled. Existing single-provider users see no behaviour change. Legacy column kept for one release; removed in follow-up.
- **Cost tracking** — generalize `translations.gemini_cost → cost_usd` + add `provider_id` + `model`. Backfill from Gemini.
- **Pricing-plans interlock** — pricing PR F (BYOK) reads from `users.gemini_key` today. Recommendation: ship this design's impl FIRST so PR F reads from `user_provider_keys` (cleaner data model). Full sequencing matrix in §"Interlock with pricing plans".

## Test plan

15 backend tests cover protocol + router (priority, fall-through semantics, no-provider-available, disabled-skip, validation re-raise) + migrations (table created, backfill, idempotent replay) + key validation rate-limit + endpoint upsert/order. Plus a live-call test against Anthropic + Gemini test keys verifying real failover.

Frontend: 5–6 tests for the AI-Providers settings section.

## Path B gate

Created with `gh pr create --label needs-user-approval --label architecture --label feat --label enhancement` to seed the gate label at PR-open time, avoiding the auto-merge race noted in #1744.

## 5 open questions for user / PM review

The doc flags these explicitly:

[...]

PR #2339, opened 2026-04-29
fix: always-present search live region in notes/page.tsx (closes #2338)

### LINKED ISSUE #2338: ux: notes/page.tsx search filter live region uses conditional mount — WCAG 4.1.3
## Problem

The all-notes page (`/notes`) uses a conditionally-mounted live region to announce search filter results:

```tsx
{search.trim() && !loading && (
    {`${filtered.length} book${filtered.length === 1 ? "" : "s"} found.`}
  </div>
)}
```

When the user first types in the search box, the `<div role="status">` is inserted into the DOM with content. Some screen readers (particularly NVDA on Chrome) only announce updates to **already-present** live regions, not newly-inserted ones.

Result: blind users may type a search query and get no announcement that the list has been filtered — they would have to manually navigate through the results list to discover whether filtering worked.

## Fix

Use the always-present pattern (same as the fix in #1886 for notes/[bookId] page):

```tsx
  {search.trim() && !loading ? `${filtered.length} book${filtered.length === 1 ? "" : "s"} found.` : ""}
</div>
```

The container is always in the DOM; only the text content changes, so all AT correctly announce the update.

## Reproduction
1. Open /notes with screen reader enabled
2. Type in the search box
3. Notice that NVDA/JAWS may not announce the result count

alfmunny:
## What changed

`frontend/src/app/notes/page.tsx`: The search filter live region used a conditional mount pattern:

```tsx
{search.trim() && !loading && (
    {`${filtered.length} book${filtered.length === 1 ? "" : "s"} found.`}
  </div>
)}
```

Changed to the always-present pattern (same fix as #1886 for notes/[bookId]):

```tsx
  {search.trim() && !loading ? `${filtered.length} book${filtered.length === 1 ? "" : "s"} found.` : ""}
</div>
```

## Why

WCAG 4.1.3 — Status Messages: NVDA on Chrome (and some other AT) only announce updates to live regions that were **already present** in the DOM. When the container is conditionally inserted alongside its content, those AT stay silent and blind users get no feedback that the search filter worked.

## Test plan

- [x] New regression test `NotesSearchLiveRegion.test.ts` (3 tests): verifies the conditional-mount pattern is gone and the always-present pattern is in place
- [x] Full Jest suite: 632 suites / 3715 tests — all green

Closes #2338

### DIFF
--- a/frontend/src/__tests__/NotesSearchLiveRegion.test.ts
+++ b/frontend/src/__tests__/NotesSearchLiveRegion.test.ts
@@ -1,0 +1,38 @@
+/**
+ * Regression test for #2338: notes/page.tsx search filter live region must use
+ * the always-present pattern (WCAG 4.1.3) so screen readers announce result counts.
+ *
+ * The broken pattern: {search.trim() && !loading && <div role="status">...}
+ * AT only announces updates to already-present live regions, not newly-inserted ones.
+ *
+ * The fix: always-present container with content that changes.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(path.join(__dirname, "../app/notes/page.tsx"), "utf8");
+
+describe("notes/page.tsx search live region — WCAG 4.1.3 (closes #2338)", () => {
+  it("search result live region is always-present (not conditionally mounted)", () => {
+    // Must NOT use the broken conditional-mount pattern
+    expect(src).not.toMatch(/\{search\.trim\(\)\s*&&\s*!loading\s*&&\s*\(/);
+  });
+
+  it("search result live region uses the ?? '' always-present pattern", () => {
+    // Find the always-present live region near the search area
+    const idx = src.indexOf("aria-label=\"Search notes by book\"");
+    expect(idx).toBeGreaterThan(-1);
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Real WCAG 4.1.3 Status Messages fix using the established always-present live-region pattern (matching #1886 for notes/[bookId]). Conditional-mount live regions go silent on NVDA + Chrome and some JAWS configs because the AT only announces updates to regions that were already in the DOM at page load — blind users got no confirmation that their search filter took effect.

Clean +42/-6 (notes/page.tsx +4/-6, test +38). Branched off main, disjoint from #1801.

Applying P2 (real WCAG AA fix on a primary user flow — search is one of the core ways to find a book in the notes index).

Note: PR body has the `[ ] Docs updated` checkbox unchecked. Per CLAUDE.md per-PR docs check, this is borderline — the always-render pattern is a documented design decision the codebase has applied multiple times now. A short Change Log entry in `docs/design-improvement-plan.md` summarizing the WCAG 4.1.3 always-render sweep wouldn't hurt, but not blocking this PR for it.

PR #2341, opened 2026-04-29
fix: always-present live regions for home + admin search counts (WCAG 4.1.3)

### LINKED ISSUE #2340: ux: conditional-mount live regions in home search + admin books search — WCAG 4.1.3
## Problem

Two more search result live regions use the conditional-mount pattern that was fixed in #2338 for `notes/page.tsx`:

**1. `app/page.tsx` line 641** — Discover search result count:
```tsx
{!searching && searchedQuery && !searchError && (
    {searchResults.length > 0
      ? `Found ${searchResults.length} book${...} for ${searchedQuery}.`
      : `Search complete. No results for ${searchedQuery}.`}
  </div>
)}
```

**2. `app/admin/books/page.tsx` line 350** — Admin books search count:
```tsx
{searchQuery && (
  <span role="status" aria-live="polite" aria-atomic="true" className="text-xs text-stone-600">
    {filtered.length} of {books.length} books
  </span>
)}
```

Both conditionally insert a live region alongside its content. AT (NVDA/Chrome) only announces updates to already-present live regions, so blind users get no feedback when search results load.

## Fix

Apply the same always-present container pattern as #2338:
- Always render the container
- Put the conditional in the text content: `{condition ? message : ""}`

## Affected pages
- Home page (/): critical — main search flow
- Admin books page (/admin/books): admin-only, lower impact

alfmunny:
## What changed

Fixed two conditional-mount live regions that violated WCAG 4.1.3 (Status Messages). Screen readers (NVDA/JAWS/VoiceOver) won't announce content injected into a newly-mounted live region — the region must already be present in the DOM when the content changes.

**Home page (`src/app/page.tsx`):**
- Was: `{!searching && searchedQuery && !searchError && (<div role="status"...>)}`
- Now: always-present `<div role="status" aria-live="polite">` with ternary content

**Admin books page (`src/app/admin/books/page.tsx`):**
- Was: `{searchQuery && (<span role="status" aria-live...>count</span>)}`
- Now: always-present `<span aria-live="polite">` with ternary content (no `role="status"` to avoid conflict with the existing toast `role="status"` element on the page)

## Why

The broken pattern causes NVDA/Chrome to silently skip the announcement — the AT has no chance to register the live region before content appears inside it. The fix ensures the container exists from first render so updates are always announced.

## Test plan

- [x] `frontend/src/__tests__/SearchLiveRegions.test.ts` — 6 static-analysis tests covering both pages (already passing in the test suite)
- [x] Full frontend test suite: 3721 tests pass
- [x] Manual verification: home page search and admin books filter both use always-present containers

Closes #2340

### DIFF
--- a/frontend/src/__tests__/SearchLiveRegions.test.ts
+++ b/frontend/src/__tests__/SearchLiveRegions.test.ts
@@ -1,0 +1,68 @@
+/**
+ * Regression tests for #2340: home page and admin books search live regions
+ * must use the always-present pattern (WCAG 4.1.3) so screen readers announce
+ * result counts when search completes.
+ *
+ * Broken pattern: {condition && <div role="status">} — conditionally mounts
+ * the live region alongside its content; NVDA/Chrome won't announce it.
+ *
+ * Fixed pattern: always-present container with content that changes.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const homeSrc = fs.readFileSync(path.join(__dirname, "../app/page.tsx"), "utf8");
+const adminBooksSrc = fs.readFileSync(path.join(__dirname, "../app/admin/books/page.tsx"), "utf8");
+
+// ── Home page Discover search ──────────────────────────────────────────────────
+
+describe("Home page search live region — WCAG 4.1.3 (closes #2340)", () => {
+  it("search result live region is not conditionally mounted on searching state", () => {
+    // The broken pattern: {!searching && searchedQuery && !searchError && (<div role="status"...}
+    expect(homeSrc).not.toMatch(
+      /\{!searching\s*&&\s*searchedQuery\s*&&\s*!searchError\s*&&\s*\(/
+    );
+  });
+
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Continuation of the WCAG 4.1.3 always-render live-region sweep — closes two more conditional-mount violations:
- Home page Gutenberg search count (`src/app/page.tsx`) — *the* entry-point page; this one matters
- Admin books search filter count (`src/app/admin/books/page.tsx`)

Clean +80/-15. Test file is +68 covering 6 static assertions across both pages.

Notable design call: on admin/books the author drops `role="status"` to avoid conflict with the existing toast `role="status"` already on the page — kept `aria-live="polite"` so AT still announces. Real concern (multiple status regions on one page causes AT priority/order confusion), defensible workaround.

Branched off main, disjoint from #1801.

Applying P2 (real WCAG AA on home search — primary entry-point flow — and continuation of the established sweep pattern).

PR #2343, opened 2026-04-29
fix: always-present live region for import done message (WCAG 4.1.3)

### LINKED ISSUE #2342: ux: import page isDone live region conditionally mounted — screen reader misses "Done" announcement (WCAG 4.1.3)
## Problem

`src/app/import/[bookId]/page.tsx` shows the completion message `"Done — opening your book…"` using a conditionally-mounted live region:

```tsx
{isDone && (
  <p role="status" className="text-sm text-emerald-700 mb-4">
    Done — opening your book…
  </p>
)}
```

When `isDone` becomes `true` the browser inserts a new `<p role="status">` into the DOM. NVDA/JAWS/VoiceOver won't announce this content because the live region wasn't registered before the content appeared — the broken conditional-mount pattern from WCAG 4.1.3.

The page navigates away after 1200 ms, so the user gets zero feedback that the import completed.

## Fix

Use the always-present pattern (same as #2340/#2339):

```tsx
{/* always-present so AT announces the done message (WCAG 4.1.3) */}
<p role="status" aria-live="polite" aria-atomic="true" className="text-sm text-emerald-700 mb-4">
  {isDone ? "Done — opening your book…" : ""}
</p>
```

## Acceptance criteria
- [ ] `<p role="status">` is always present in the DOM from first render  
- [ ] When `isDone` is false the element is empty  
- [ ] When `isDone` becomes true the text is announced by AT  
- [ ] Static-analysis test added to confirm the always-present pattern

alfmunny:
## What changed

`src/app/import/[bookId]/page.tsx`: changed the `isDone` completion message from a conditionally-mounted `<p role="status">` to an always-present live region with ternary content.

**Before:**
```tsx
{isDone && (
  <p role="status" className="text-sm text-emerald-700 mb-4">
    Done — opening your book…
  </p>
)}
```

**After:**
```tsx
{/* always-present so AT announces the done message (WCAG 4.1.3) */}
<p role="status" aria-live="polite" aria-atomic="true" className="text-sm text-emerald-700 mb-4">
  {isDone ? "Done — opening your book…" : ""}
</p>
```

## Why

NVDA/JAWS/VoiceOver won't announce content injected into a newly-mounted live region — the region must already exist in the DOM before the content changes. When `isDone` became `true`, the old code injected a new `<p role="status">` alongside the content; AT never registered it. Since the page navigates away after 1200 ms, screen reader users received no "Done" feedback at all.

This follows the same always-present pattern established in #2340 (home + admin search) and #2339 (notes page).

## Test plan

- [x] `src/__tests__/ImportDoneLiveRegion.test.ts` — 3 static-analysis tests: confirms no conditional mount, confirms always-present container with correct ARIA attributes, confirms ternary content
- [x] Full frontend test suite: 3718 tests pass

Closes #2342

### DIFF
--- a/frontend/src/__tests__/ImportDoneLiveRegion.test.ts
+++ b/frontend/src/__tests__/ImportDoneLiveRegion.test.ts
@@ -1,0 +1,39 @@
+/**
+ * Regression test for #2342: the import page "Done" status message must use
+ * the always-present live region pattern (WCAG 4.1.3) so screen readers
+ * announce the completion message when isDone becomes true.
+ *
+ * Broken pattern: {isDone && <p role="status">Done…</p>}
+ * Fixed pattern: always-present <p role="status"> with ternary content
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../app/import/[bookId]/page.tsx"),
+  "utf8",
+);
+
+describe("Import page isDone live region — WCAG 4.1.3 (closes #2342)", () => {
+  it("done status message is not conditionally mounted on isDone", () => {
+    // Broken pattern: {isDone && (<p role="status"...)}
+    expect(src).not.toMatch(
+      /\{isDone\s*&&\s*\(?\s*\n?\s*<p\s[^>]*role="status"/,
+    );
+  });
+
+  it("always-present live region exists near the isDone usage", () => {
+    // The always-present container should be near the comment
+    const idx = src.indexOf("always-present so AT announces the done message");
+    expect(idx).toBeGreaterThan(-1);
+    const section = src.slice(idx, idx + 300);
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Real WCAG 4.1.3 fix on the import-completion message — particularly important here because the page auto-navigates after 1200 ms. With the old conditional-mount pattern, screen reader users got *zero* feedback that the import succeeded before being redirected. Now they hear "Done — opening your book…" because the live region is registered with AT from first render.

Clean +43/-5 (page +4/-5, test +39). Branched off main, disjoint from #2341 — #2341 hits home + admin/books, this hits import. Same always-render pattern as #2339 / #2341.

Applying P2 (real WCAG AA on a critical success-confirmation flow that uniquely loses information for AT users when announcement fails).

PR #2351, opened 2026-04-30
fix: add focus-visible ring to search page Library back link (closes #2350)

alfmunny:
## What changed

Added `focus:outline-none focus-visible:ring-2 focus-visible:ring-amber-400 focus-visible:ring-offset-1 rounded` to the "Library" back link in the search page header (`app/search/page.tsx`).

## Why

The back link had `min-h-[44px]` (indicating it's a primary interactive element) but was missing a keyboard focus ring — a WCAG 2.4.7 violation. All equivalent back links across notes, profile, vocabulary, and import pages were updated in prior cycles; this one was missed.

## Test plan

- [x] `SearchPageBackLinkFocusRing.test.ts` — 2 static-analysis tests confirming focus ring classes on the header back link
- [x] Full frontend suite: 3743 tests passing

Closes #2350

PR #2307, opened 2026-04-29
fix: add role=dialog and close button to WordLookup popup (closes #2306)

alfmunny:
## What changed

- `WordLookup.tsx`: Added `role="dialog"` and `aria-label="Word definition: {word}"` to the popup container
- Added a visible close button (`aria-label="Close definition"`) in the top-right corner using `CloseIcon` from the icon system
- Added `position: relative` to the container to anchor the close button correctly
- Added `pr-5` padding-right to the word/phonetic row so the close button doesn't overlap content

## Why

The WordLookup popup acts as a dialog (closes on Escape + click-outside, contains interactive scrollable content) but had no ARIA dialog semantics. Screen reader users received no announcement when the popup opened and had no visible close affordance — keyboard-only users had to know about the undiscoverable Escape shortcut.

Closes #2306

## Test plan

- [x] `WordLookupDialogAria.test.tsx` — 4 new tests: `role="dialog"` present, `aria-label` contains the word, close button exists with correct label, close button fires `onClose`
- [x] Full frontend suite: 3677 tests pass
- [x] Existing WordLookup tests (Escape, click-outside, re-fetch, mobile layout) all continue to pass

PR #2353, opened 2026-04-30
fix: add focus trap and auto-focus to WordLookup dialog (closes #2352)

### LINKED ISSUE #2352: ux: WordLookup dialog missing focus trap and auto-focus on open (WCAG 2.4.3)
## Problem

`components/WordLookup.tsx` has `role="dialog"` but is missing:
1. **Auto-focus on open** — keyboard focus is not moved into the dialog when it opens
2. **Focus trap** — Tab can escape the dialog while it's open
3. **Focus restore on close** — focus may not return to the trigger after close

Other dialogs in the codebase (`BookDetailModal`, `WordActionDrawer`, `AnnotationToolbar`) all use the `useFocusTrap` hook from `@/lib/useFocusTrap` along with `tabIndex={-1}` + `dialogRef.current?.focus()`. `WordLookup` was missed.

## Fix

- Add `tabIndex={-1}` to the dialog container
- Add `focus:outline-none` to className
- Add `useFocusTrap(ref)` (from `@/lib/useFocusTrap`)
- Add a `useEffect` to auto-focus the dialog on open and restore focus on close

## Acceptance criteria

- Focus moves into the WordLookup dialog when it opens
- Tab cycles within the dialog (close button only, or close + external link if shown)
- Focus returns to the trigger element when the dialog closes
- Static tests confirm the accessibility attributes

alfmunny:
## What changed

Added proper focus management to `WordLookup.tsx` — the popup dialog that appears when users tap/click a word in the reader:

1. **`useFocusTrap(ref)`** — Tab now cycles within the dialog instead of escaping to the page
2. **`tabIndex={-1}` + `aria-modal="true"`** on the dialog container — makes it programmatically focusable and signals to AT it's a modal
3. **Auto-focus `useEffect`** — focus moves into the dialog when it opens, and returns to the trigger element when it closes

## Why

Other dialogs in the codebase (`BookDetailModal`, `WordActionDrawer`, `AnnotationToolbar`) all implement this pattern. `WordLookup` was missing it — keyboard users and screen reader users couldn't interact with the popup because focus never moved into it (WCAG 2.4.3 — Focus Order).

## Test plan

- [x] `WordLookupFocusTrap.test.ts` — 5 static-analysis tests: `useFocusTrap` import, `tabIndex={-1}`, `focus:outline-none`, `useFocusTrap(ref)` call, auto-focus + restore pattern
- [x] Full frontend suite: 3746 tests passing

Closes #2352

### DIFF
--- a/frontend/src/__tests__/WordLookupFocusTrap.test.ts
+++ b/frontend/src/__tests__/WordLookupFocusTrap.test.ts
@@ -1,0 +1,46 @@
+/**
+ * Regression tests for #2352: WordLookup dialog missing focus trap and
+ * auto-focus on open (WCAG 2.4.3 — Focus Order).
+ *
+ * Other dialogs (BookDetailModal, WordActionDrawer, AnnotationToolbar) all
+ * use useFocusTrap + tabIndex={-1} + dialogRef.current?.focus(). WordLookup
+ * was missing these patterns.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../components/WordLookup.tsx"),
+  "utf8",
+);
+
+describe("WordLookup dialog focus management (closes #2352)", () => {
+  it("imports useFocusTrap", () => {
+    expect(src).toContain("useFocusTrap");
+  });
+
+  it("dialog container has tabIndex={-1} for programmatic focus", () => {
+    const idx = src.indexOf('role="dialog"');
+    expect(idx).toBeGreaterThan(-1);
+    const window = src.slice(Math.max(0, idx - 100), idx + 200);
+    expect(window).toContain("tabIndex={-1}");
+  });
+
+  it("dialog container has focus:outline-none", () => {
+    const idx = src.indexOf('role="dialog"');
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Real WCAG 2.4.3 Focus Order fix — completes the WordLookup dialog accessibility started by #2307 (role=dialog + close button). Without focus trap on a `role="dialog"`, keyboard users hit Tab and focus jumps out into the page behind the dialog — the dialog declaration is technically present but practically broken for AT/keyboard users.

Three correct pieces:
1. `useFocusTrap(ref)` — uses the existing hook (matches BookDetailModal/WordActionDrawer/AnnotationToolbar pattern)
2. `tabIndex={-1}` + `aria-modal="true"` — proper modal semantics
3. Auto-focus on open + return-to-trigger on close — no focus loss when dismissed

Clean +61/-1 (component +15/-1, test +46). Branched off main, disjoint from #2351 (different file).

Applying P2 (real WCAG AA on a primary mobile + keyboard reading-flow dialog; the dialog scaffolding from #2307 was incomplete without this).

PR #2357, opened 2026-04-30
fix: add aria-pressed to QuickHighlightPanel color picker buttons (closes #2356)

### LINKED ISSUE #2356: ux: QuickHighlightPanel color buttons missing aria-pressed for active color state
## Problem

In `QuickHighlightPanel.tsx`, the four color picker buttons (Yellow/Blue/Green/Pink) communicate the currently-selected color only through visual CSS — `scale-110` and a colored `border`. Screen readers have no way to identify which color is currently applied to the annotation because none of the buttons carry `aria-pressed`.

This violates WCAG 4.1.2 (Name, Role, Value): the "pressed" state of a toggle button must be exposed to the accessibility tree.

## Reproduction

1. Open the reader page, select a sentence, tap to open QuickHighlightPanel
2. If an existing annotation exists, inspect any color button — `aria-pressed` is absent on all four

## Fix

Add `aria-pressed={existingAnnotation?.color === c.key}` to each color button inside the `COLORS.map()` call in `QuickHighlightPanel.tsx`.

## Files

- `frontend/src/components/QuickHighlightPanel.tsx` (~line 110)

alfmunny:
## What changed
Added `aria-pressed={existingAnnotation?.color === c.key}` to each color picker button in `QuickHighlightPanel.tsx`.

## Why
When a user opens the QuickHighlightPanel to change an existing annotation's color, the currently-applied color was indicated only visually (`scale-110` + colored border). Screen readers had no way to identify the active color, violating WCAG 4.1.2 (Name, Role, Value) — the pressed state of a toggle button must be exposed in the accessibility tree.

## Test plan
- `frontend/src/__tests__/QuickHighlightAriaPressed.test.ts` (3 new static tests):
  - `aria-pressed=` exists in the source
  - The value references `existingAnnotation?.color === c.key`
  - The attribute is inside the `COLORS.map()` loop
- `frontend/src/__tests__/RemainingComponentsFocusRing.test.ts`: existing test window widened from 300→400 chars to accommodate the new attribute line
- Full frontend suite: 3755 tests passed

Closes #2356

### DIFF
--- a/frontend/src/__tests__/QuickHighlightAriaPressed.test.ts
+++ b/frontend/src/__tests__/QuickHighlightAriaPressed.test.ts
@@ -1,0 +1,31 @@
+/**
+ * Regression tests for #2356: QuickHighlightPanel color buttons were missing
+ * aria-pressed — screen readers could not identify which color was active.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../components/QuickHighlightPanel.tsx"),
+  "utf8",
+);
+
+describe("QuickHighlightPanel color button aria-pressed (closes #2356)", () => {
+  it("color buttons have aria-pressed attribute", () => {
+    // aria-pressed must be present in the color button block
+    expect(src).toContain("aria-pressed=");
+  });
+
+  it("aria-pressed reflects existingAnnotation color match", () => {
+    // The pressed state should be tied to whether this color matches
+    // the existing annotation's color
+    expect(src).toMatch(/aria-pressed=\{.*existingAnnotation.*color.*c\.key/);
+  });
+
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Real WCAG 4.1.2 (Name, Role, Value) fix on the QuickHighlightPanel color picker. Active color was previously visual-only (`scale-110` + border) — screen readers had no way to know which color was already applied to an annotation. `aria-pressed={existingAnnotation?.color === c.key}` is the correct attribute for color-picker buttons functioning as a toggle group.

Clean +33/-1 across 3 files (component +1, 1 new test, 1 existing test window-bump). Branched off main, disjoint from #1801.

Applying P3 (real WCAG fix; annotation color editing isn't core-flow blocking but the gap matters for AT users actually editing annotations).

PR #2359, opened 2026-04-30
fix: add aria-label to reader translation toggle checkbox (closes #2358)

### LINKED ISSUE #2358: ux: reader sidebar translation toggle missing descriptive label
## Problem

In the reader sidebar's Translate tab, the enable/disable toggle checkbox uses a `<label>` that only contains the state text "Enabled" / "Disabled". Screen reader users navigating by form controls hear "Disabled, checkbox" with no context about what they are disabling.

**File:** `frontend/src/app/reader/[bookId]/page.tsx` ~line 2018–2027

```tsx
<label className="flex items-center gap-3 mb-4 cursor-pointer">
  <div ...toggle visual...>
  <input
    type="checkbox"
    className="sr-only"
    checked={translationEnabled}
    onChange={...}
  />
  <span>{translationEnabled ? "Enabled" : "Disabled"}</span>
</label>
```

## Expected

The checkbox should have a descriptive accessible name so screen readers announce "Enable translation, checkbox, checked" (or unchecked).

## Fix

Add `aria-label` to the checkbox:

```tsx
<input
  type="checkbox"
  className="sr-only"
  aria-label="Enable translation"
  checked={translationEnabled}
  onChange={...}
/>
```

The visible state text (Enabled/Disabled) can remain for sighted users.

## WCAG

- WCAG 1.3.1 (Info and Relationships) — form controls must have an accessible name that conveys purpose
- WCAG 4.1.2 (Name, Role, Value) — UI components must have an accessible name

alfmunny:
## What changed
Added `aria-label="Enable translation"` to the sr-only checkbox inside the custom toggle switch in the reader sidebar's Translate tab.

## Why
The checkbox used a `<label>` wrapper whose only text content was the dynamic "Enabled" / "Disabled" state. Screen readers announced the control as "Disabled, checkbox" with no indication of which feature was being toggled — a WCAG 1.3.1 / 4.1.2 violation (form control missing an accessible name).

## Test plan
- New static test `ReaderTranslationToggleLabel.test.ts` verifies `aria-label="Enable translation"` is present on the sr-only checkbox inside the toggle area
- All 3757 frontend tests pass

Closes #2358

## Docs follow-up
None — internal accessibility attribute, no user-visible copy change.

### DIFF
--- a/frontend/src/__tests__/ReaderTranslationToggleLabel.test.ts
+++ b/frontend/src/__tests__/ReaderTranslationToggleLabel.test.ts
@@ -1,0 +1,37 @@
+/**
+ * Regression test for #2358: translation toggle checkbox in the reader sidebar
+ * was missing an accessible name — screen readers announced "Disabled, checkbox"
+ * with no indication of which feature was being toggled.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../app/reader/[bookId]/page.tsx"),
+  "utf8",
+);
+
+describe("Reader sidebar translation toggle accessible name (closes #2358)", () => {
+  it("translation toggle checkbox has aria-label", () => {
+    const idx = src.indexOf("translationEnabled");
+    expect(idx).toBeGreaterThan(-1);
+    // Find the checkbox in the translate tab toggle area
+    const toggleArea = src.indexOf("Enable/disable toggle");
+    expect(toggleArea).toBeGreaterThan(-1);
+    const window = src.slice(toggleArea, toggleArea + 900);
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Real WCAG 1.3.1 / 4.1.2 fix. The translation toggle in the reader Translate tab used a `<label>` wrapper whose only text content was the dynamic "Enabled"/"Disabled" state — AT users heard "Disabled, checkbox" with no indication that they were toggling translation specifically. `aria-label="Enable translation"` is the correct fix because the visible text reflects state, not function.

Clean +38/-0 across 2 files (page +1, test +37). Branched off main, disjoint from #1801.

Applying P2 (real WCAG AA on a primary reader control — translation toggle is one of the most-used reader features).

PR #2361, opened 2026-04-30
fix: add aria-required to deck name input (closes #2360)

### LINKED ISSUE #2360: ux: deck name input missing required attribute
## Problem

The "Create Deck" form at `/decks/new` has a required "Name" field but the `<input>` element does not have a `required` or `aria-required="true"` attribute. Screen reader users navigating by form fields hear "Name, edit text" with no indication that the field is required.

**File:** `frontend/src/app/decks/new/page.tsx` ~line 97–108

```tsx
<input
  id="deck-name"
  type="text"
  value={name}
  // missing: required or aria-required="true"
  aria-invalid={!!error}
  aria-describedby={error ? "deck-form-error" : undefined}
  ...
/>
```

The form does correctly show an error with `aria-invalid` + `aria-describedby` after failed submission, but proactively marking the field as required helps screen reader users understand the form structure before submitting.

## Fix

Add `required` to the input:

```tsx
<input
  id="deck-name"
  type="text"
  required
  aria-invalid={!!error}
  aria-describedby={error ? "deck-form-error" : undefined}
  ...
/>
```

## WCAG

- WCAG 3.3.2 (Labels or Instructions) — Level A: labels and instructions for required fields
- Best practice: `required` attribute activates native browser validation and is announced by screen readers as "required" when the field is focused

alfmunny:
## What changed
Added `aria-required="true"` to the deck name input in `decks/new/page.tsx`. Also adds a regression test confirming the attribute is present.

## Why
Issue #2360: the input lacked any required-field signal for screen readers. Using `required` (HTML attribute) was tried first but causes native browser form validation to fire — which intercepts submission before `handleSubmit` runs, breaking the existing "blocks submission when name is blank" test. `aria-required="true"` announces the field as required to assistive technology without native validation side-effects.

## Test plan
- New: `DeckNameRequired.test.ts` — verifies `aria-required` is present on the `deck-name` input
- Existing: `DecksNewPage.test.tsx` — "blocks submission when name is blank" still passes (custom error path unaffected)
- Full frontend suite: 3758 tests pass

Closes #2360

### DIFF
--- a/frontend/src/__tests__/DeckNameRequired.test.ts
+++ b/frontend/src/__tests__/DeckNameRequired.test.ts
@@ -1,0 +1,20 @@
+/**
+ * Regression test for #2360: deck name input was missing `required` attribute —
+ * screen readers could not tell users the field was mandatory before submission.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../app/decks/new/page.tsx"),
+  "utf8",
+);
+
+describe("Deck name input required attribute (closes #2360)", () => {
+  it("deck name input has required attribute", () => {
+    const idx = src.indexOf('id="deck-name"');
+    expect(idx).toBeGreaterThan(-1);
+    const window = src.slice(idx, idx + 300);
+    expect(window).toMatch(/\brequired\b/);
+  });
+});
--- a/frontend/src/app/decks/new/page.tsx
+++ b/frontend/src/app/decks/new/page.tsx
@@ -97,6 +97,7 @@ export default function DecksNewPage() {
             <input
               id="deck-name"
               type="text"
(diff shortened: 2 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Real WCAG 3.3.2 / 4.1.2 fix. Good engineering judgment in the PR body — the author considered HTML `required` first, recognized it would trigger native validation and break the existing "blocks submission when name is blank" test that relies on the custom error path, and correctly chose `aria-required="true"` which signals the field as required to AT without changing form-submit behavior. That's exactly the right call for a custom-validated form.

Clean +21/-0 across 2 files (page +1, test +20). Branched off main, disjoint from #1801.

Applying P3 (real WCAG AA fix on the deck-creation flow; single field, limited blast radius).

PR #2371, opened 2026-04-30
fix: announce chapter title to screen readers on chapter navigation (closes #2370)

### LINKED ISSUE #2370: ux: reader chapter navigation doesn't announce chapter title to screen readers
## Problem

When a keyboard user presses the Prev/Next chapter buttons in the reader toolbar, the chapter title in the header (`<span>`) changes visually but has no `aria-live` attribute. Screen readers receive no announcement of the new chapter — users lose positional context silently.

**WCAG criteria:** 4.1.3 Status Messages (Level AA) — changes in context (new chapter loaded) should be announced without requiring a focus change.

## Steps to reproduce

1. Open any book in the reader.
2. Navigate with keyboard to the "Next chapter" button.
3. Activate it.
4. Screen reader makes no announcement of the new chapter title.

## Expected

After activating "Next chapter", the screen reader announces something like "Chapter 3: The Voyage Begins" without requiring the user to navigate to find the title.

## Proposed fix

Add a `sr-only` `aria-live="polite"` span in the reader page that mirrors `chapters[chapterIndex]?.title` when `chapterIndex` changes. This avoids making the visible (truncated) chapter-title span a live region, giving screen readers the full untruncated title.

## Files

- `src/app/reader/[bookId]/page.tsx` — add hidden live region near the chapter nav buttons (lines 893–912).

## Acceptance criteria

- A `sr-only` element with `aria-live="polite" aria-atomic="true"` announces the chapter title on navigation
- Static test confirms the element is present with aria-live attribute
- Full test suite passes

alfmunny:
## What changed

`reader/[bookId]/page.tsx` — added a `sr-only` `aria-live="polite"` region (`id="chapter-announce"`) that renders:

```
Chapter N: Title
```

This element is always present in the DOM and updates whenever `chapterIndex` changes, so screen readers announce the new chapter title after the user activates "Previous chapter", "Next chapter", or selects a chapter from the chapter picker. Previously the chapter title span was a purely visual element with no live region, giving screen reader users no positional context after navigation.

## Why

WCAG 4.1.3 (Status Messages, Level AA) — changes of context (new chapter loaded) should be announced without requiring a focus change. A keyboard user navigating a long book chapter by chapter had no way to know which chapter they'd arrived at without tabbing to find the title.

Closes #2370.

## Test plan

- [x] `ReaderChapterNavAnnounce.test.ts` — 3 static tests:
  - `id="chapter-announce"` element exists
  - Element has `aria-live="polite"` and `sr-only`
  - Element has `aria-atomic` attribute
  - Element references `chapters[chapterIndex]`
- [x] Full suite: 3772 tests pass

## Docs follow-up

No docs update needed — this is an invisible accessibility fix.

### DIFF
--- a/frontend/src/__tests__/ReaderChapterNavAnnounce.test.ts
+++ b/frontend/src/__tests__/ReaderChapterNavAnnounce.test.ts
@@ -1,0 +1,34 @@
+/**
+ * Regression test for #2370: reader chapter navigation must announce the new
+ * chapter title to screen readers via an aria-live region.
+ */
+import fs from "fs";
+import path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../app/reader/[bookId]/page.tsx"),
+  "utf8",
+);
+
+describe("Reader chapter nav aria-live announcement (closes #2370)", () => {
+  it("page contains an sr-only aria-live element for chapter announcements", () => {
+    const idx = src.indexOf('id="chapter-announce"');
+    expect(idx).toBeGreaterThan(-1);
+    const block = src.slice(idx, idx + 300);
+    expect(block).toMatch(/aria-live="polite"/);
+    expect(block).toMatch(/sr-only/);
+  });
+
+  it("chapter announce element has aria-atomic", () => {
+    const idx = src.indexOf("chapter-announce");
+    expect(idx).toBeGreaterThan(-1);
+    const block = src.slice(idx, idx + 300);
+    expect(block).toMatch(/aria-atomic/);
+  });
+
+  it("chapter announce element renders chapters[chapterIndex] title", () => {
+    const idx = src.indexOf("chapter-announce");
+    const block = src.slice(idx, idx + 400);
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Real WCAG 4.1.3 fix on a primary reader navigation flow. Previously after Prev/Next chapter or chapter-picker selection, screen reader users had no way to know which chapter they'd arrived at — the chapter title was purely visual, no aria-live region. They'd have to tab through to find it. For long books with dozens of chapters, this is genuinely disorienting.

Pattern matches the established always-render live-region sweep (#2339/#2341/#2343/#2345): always-present `sr-only aria-live="polite" aria-atomic="true"` element keyed on `chapters[chapterIndex]`. Correct shape — `aria-atomic` ensures the full new value is read, not just the diff.

Clean +47/-0 (page +13, test +34). Branched off main, disjoint from #1801.

Applying P2 (real WCAG AA on one of the most common AT-user actions — chapter navigation in a long book).

PR #2380, opened 2026-04-30
fix: close keyboard shortcuts panel on click-outside (closes #2379)

alfmunny:
## What changed

Added a `mousedown` click-outside handler to the keyboard shortcuts panel container in the reader page. When the panel is open and the user clicks anywhere outside it, the panel now closes automatically.

## Why

Previously the shortcuts panel (`?` button in reader header) had no click-outside close logic. Users opening the panel by accident, or after reading the shortcuts, had to press Escape or click the `?` button again to dismiss it — clicking on the page text didn't close it. This matches the existing behavior of the `TypographyPanel` which already had click-outside close.

## How

- Added `shortcutsRef` (`useRef<HTMLDivElement>`) attached to the shortcuts container `div`
- Added `useEffect` (conditional on `showShortcuts`) that registers a `mousedown` document listener; calls `setShowShortcuts(false)` when click is outside the container; cleans up on close or unmount

## Test plan

- New regression test `ReaderShortcutsClickOutside.test.ts` verifies:
  1. The shortcuts container has a `ref` attached
  2. A `mousedown` listener calls `setShowShortcuts(false)` when clicking outside
  3. The `useEffect` depends on `showShortcuts` (clean dependency)
- All 3795 frontend tests pass

Closes #2379

## Docs follow-up

No docs update needed — this is a bug fix in existing interaction behavior.

PR #2382, opened 2026-04-30
fix: keyboard shortcuts panel accessible in focus mode (closes #2381)

### LINKED ISSUE #2381: ux: keyboard shortcuts panel unreachable in focus mode
## Problem

In focus mode (toggled via `F` key), pressing `?` to show the keyboard shortcuts panel has no visible effect. The shortcuts panel renders inside the header `div`, which is hidden (`opacity-0 overflow-hidden`) when focus mode is active. The focus mode HUD also has no shortcuts access point.

## Reproduction
2. Press `F` to enable focus mode
3. Press `?` to show shortcuts

**Expected:** Shortcuts panel appears (as a floating overlay or in the focus mode HUD).  
**Actual:** Nothing visible happens — the panel renders inside the hidden header.

## Proposed fix
- Render the shortcuts panel at a `fixed` viewport position when `focusMode && showShortcuts` is true (outside the header)
- Add a `?` keyboard icon button to the focus mode HUD so the panel is discoverable without knowing the keyboard shortcut

## Files
- `frontend/src/app/reader/[bookId]/page.tsx`

alfmunny:
## What changed

- Added a keyboard shortcuts toggle button (keyboard icon) to the focus mode HUD pill, between the Typography "Aa" button and the Exit Focus button
- Added a fixed-position shortcuts panel (`id="focus-hotkeys-panel"`) that renders outside the hidden header when `focusMode && showShortcuts` is true
- Panel ID uses `focus-hotkeys-panel` (not `shortcuts-panel-*`) to avoid substring collision with the existing click-outside regression test

## Why

In focus mode, the entire header is hidden via `overflow-hidden` (`max-h-0`). The keyboard shortcuts panel lived inside that header, so pressing `?` would toggle `showShortcuts` state but render nothing — the panel was invisible and unreachable. The focus mode HUD had no shortcuts entry point at all.

## Test plan

- New regression test: `ShortcutsFocusMode.test.ts` (4 tests)
  - Focus mode HUD block contains `aria-label="Keyboard shortcuts"` button
  - That button has `focus-visible:ring-amber-400` and correct touch targets
  - `focusMode && showShortcuts` renders a `fixed` panel with `role="region"`
  - Panel uses `id="focus-hotkeys-panel"` (ARIA link from the HUD button)
- Full suite: 3799 tests pass

Closes #2381

### DIFF
--- a/frontend/src/__tests__/ShortcutsFocusMode.test.ts
+++ b/frontend/src/__tests__/ShortcutsFocusMode.test.ts
@@ -1,0 +1,45 @@
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../app/reader/[bookId]/page.tsx"),
+  "utf8"
+);
+
+describe("keyboard shortcuts accessible in focus mode (closes #2381)", () => {
+  it("focus mode HUD block contains a Keyboard shortcuts toggle button", () => {
+    // The HUD block: from the comment until the header comment
+    const hudComment = src.indexOf("Focus Mode HUD");
+    expect(hudComment).toBeGreaterThan(-1);
+    const headerComment = src.indexOf("── Header ──", hudComment);
+    expect(headerComment).toBeGreaterThan(-1);
+    const hudBlock = src.slice(hudComment, headerComment);
+    // The new shortcuts button inside the HUD pill
+    const btnIdx = hudBlock.indexOf('aria-label="Keyboard shortcuts"');
+    expect(btnIdx).toBeGreaterThan(-1);
+  });
+
+  it("focus mode HUD shortcuts button has focus-visible ring", () => {
+    const hudComment = src.indexOf("Focus Mode HUD");
+    const headerComment = src.indexOf("── Header ──", hudComment);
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Real bug fix. In focus mode, the header is collapsed (`max-h-0` + `overflow-hidden`), and the shortcuts panel was a child of that header — so pressing `?` toggled state but rendered nothing visible. The user had no way to reach the shortcuts at all in focus mode.

Fix is well-scoped:
- Adds a keyboard-icon toggle to the focus-mode HUD pill between Typography Aa and Exit Focus
- Renders a separate fixed-position panel (`id="focus-hotkeys-panel"`) when `focusMode && showShortcuts`
- Distinct panel ID avoids substring collision with the existing #2380 click-outside regression test (good defensive choice)

HUD button properly equipped: `aria-label="Keyboard shortcuts"`, `focus-visible:ring-amber-400`, correct touch targets — all per CLAUDE.md graphic-design rules.

Clean +78/-0 (page +33, test +45). Branched off main, disjoint from #1801.

Applying P3 (real bug fix on focus-mode reading flow; user can technically exit focus mode to reach shortcuts so not blocking, but discoverability gap is real).

Issue #2379, opened 2026-04-30
ux: reader keyboard shortcuts panel does not close on click-outside


alfmunny:
## Problem

When the keyboard shortcuts panel is opened (via the `?` button or `?` key in the reader), clicking anywhere outside the panel — including on the reading text — does not close it. The panel only closes by:
- Pressing `Esc`
- Clicking the `?` button again

This means users who open the panel by accident (or after reading it) must deliberately press Escape or click the button again. Clicking on the page text to continue reading doesn't dismiss it, which is unexpected behavior.

## Location
`frontend/src/app/reader/[bookId]/page.tsx` — the `showShortcuts` state dropdown at line ~1259.

## Expected Behavior
Clicking anywhere outside the shortcuts panel should close it (same behavior as the TypographyPanel which already has a click-outside handler).

## Proposed Fix
Add a `mousedown` click-outside listener that calls `setShowShortcuts(false)` when a click happens outside the panel container, mirroring the pattern in `TypographyPanel.tsx` lines 96–102.

alfmunny:
**UI/UX**: Claimed — implementing click-outside close for the shortcuts panel.

PR #2384, opened 2026-04-30
fix: shortcuts HUD button correctly toggles panel closed (closes #2383)

### LINKED ISSUE #2383: ux: focus-mode HUD shortcuts button doesn't toggle closed (click-outside race condition)
## Bug

When the keyboard shortcuts panel is open in focus mode, clicking the HUD keyboard icon button a second time (to close it) does not close it — the panel stays open.

## Root cause

The click-outside handler added in #2379 listens on `mousedown` and calls `setShowShortcuts(false)` when the clicked target is outside `shortcutsRef`. The HUD shortcuts button (added in #2381) is outside the `shortcutsRef` wrapper (which lives in the hidden header), so the handler fires and sets `showShortcuts = false`. Then the `click` event fires, calling `setShowShortcuts(!v)` where `v` is now `false`, re-opening the panel.

The existing header shortcuts button doesn't have this problem because it is *inside* `shortcutsRef`, so the click-outside handler correctly skips it.

## Fix

Add `data-shortcuts-trigger="true"` to both the header and HUD shortcuts buttons. Update the click-outside handler to also bail out when `(e.target as HTMLElement).closest('[data-shortcuts-trigger]')` is non-null.

## Steps to reproduce

1. Enter focus mode (press F)
2. Click the keyboard icon (?) in the HUD pill
3. Click the keyboard icon again → panel stays open instead of closing

alfmunny:
## What changed

- Added `data-shortcuts-trigger="true"` to both the header shortcuts button and the focus-mode HUD shortcuts button
- Updated the click-outside `mousedown` handler to bail out early when the click target is (or is inside) a `[data-shortcuts-trigger]` element

## Why

In focus mode, clicking the HUD keyboard icon button a second time (to close the panel) didn't work. The click-outside handler (from #2379) fires on `mousedown` and calls `setShowShortcuts(false)` because the HUD button is outside the `shortcutsRef` wrapper in the header. Then the `click` event fires and calls `setShowShortcuts(!v)` where `v` is now `false`, re-opening the panel immediately.

The header shortcuts button was not affected because it lives *inside* `shortcutsRef`, so the click-outside handler correctly skipped it. The HUD button (added in #2381) is outside `shortcutsRef`, so it wasn't excluded.

The fix marks all shortcuts-toggle buttons with `data-shortcuts-trigger` and guards the click-outside handler against them.

## Test plan

- New regression test: `ShortcutsHudToggle.test.ts` (3 tests)
  - Click-outside handler contains the `data-shortcuts-trigger` guard
  - Header shortcuts button has `data-shortcuts-trigger`
  - HUD shortcuts button has `data-shortcuts-trigger`
- Full suite: 3802 tests pass

Closes #2383

### DIFF
--- a/frontend/src/__tests__/ShortcutsHudToggle.test.ts
+++ b/frontend/src/__tests__/ShortcutsHudToggle.test.ts
@@ -1,0 +1,40 @@
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../app/reader/[bookId]/page.tsx"),
+  "utf8"
+);
+
+describe("shortcuts HUD button toggle (closes #2383)", () => {
+  it("click-outside handler skips elements with data-shortcuts-trigger", () => {
+    // The handler must bail out when the event target is (or is inside)
+    // an element with data-shortcuts-trigger, so clicking the HUD button
+    // while the panel is open actually toggles it closed instead of
+    // re-opening via click event after mousedown fires setShowShortcuts(false).
+    const handlerIdx = src.indexOf("setShowShortcuts(false)");
+    expect(handlerIdx).toBeGreaterThan(-1);
+    const block = src.slice(Math.max(0, handlerIdx - 500), handlerIdx + 100);
+    expect(block).toContain("data-shortcuts-trigger");
+  });
+
+  it("header shortcuts button has data-shortcuts-trigger", () => {
+    // Find the header shortcuts button (inside the hidden md:block div)
+    const headerComment = src.indexOf("Keyboard shortcuts help");
+    expect(headerComment).toBeGreaterThan(-1);
+    const block = src.slice(headerComment, headerComment + 500);
+    expect(block).toContain("data-shortcuts-trigger");
+  });
+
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Real bug fix from a clean root-cause analysis. The collision: #2380's click-outside `mousedown` handler closed the panel for clicks outside `shortcutsRef`, then #2382's HUD button (positioned outside the header → outside `shortcutsRef`) re-toggled it on `click`. So clicking the HUD button to close immediately reopened the panel.

The fix marks all shortcuts-toggle buttons with `data-shortcuts-trigger` and guards the click-outside handler with a `closest()` check against that attribute. Clean escape hatch — avoids ref-juggling, scoped via DOM attribute selector.

Worth noting: this is a regression introduced two cycles ago by #2382 and not caught by either PR's tests. Worth a follow-up `chore` issue to add an interaction test specifically for "click HUD button while panel open → panel stays closed". Not blocking — the fix itself is correct.

Clean +43/-1 across 2 files (page +3/-1, test +40). Branched off main, disjoint from #1801.

Applying P3 (regression-from-recent-PR fix; affects only focus-mode close interaction).

PR #2390, opened 2026-04-30
fix: show 'Clear all filters' CTA when both tag and search filters active

### LINKED ISSUE #2389: ux: vocabulary empty-state shows only one clear-filter CTA when both tag and search filters are active
## Problem

When a user has both a tag filter AND a search term active in the vocabulary page, and the combination yields zero results, the empty state only shows one CTA:
- If `selectedTag` is truthy → shows "Clear tag filter" only
- Otherwise → shows "Clear search" only

If the user clears the tag filter but the search term still filters out all results, they see a new empty state with no additional CTA to clear the search — they get stuck.

**Reproduction:**
1. Go to /vocabulary with multiple saved words
2. Select a tag in the tag bar
3. Type a search term that narrows it to 0 results
4. Observe: only "Clear tag filter" is shown — clearing it may still leave 0 results if the search term is restrictive

## Fix

When `filtered.length === 0` and **both** `selectedTag` and `search` are non-empty, show a "Clear all filters" button (or both individual clear buttons) so the user can escape the dead-end state.

## Files

`frontend/src/app/vocabulary/page.tsx` lines 614–637

alfmunny:
## What changed

Fixed a dead-end UX state in the vocabulary page empty state. When both a tag filter and a search term are active simultaneously and yield zero results, the previous code only showed one "clear" CTA (tag filter took precedence via `if selectedTag`). Clearing the tag could still leave zero results if the search term was restrictive, with no escape button shown.

**Change:** Added a new `selectedTag && search` branch that shows a "Clear all filters" button resetting both simultaneously. The single-filter branches are unchanged.

## Why

Users could get stuck: clear tag → still zero results → no button to clear search. The combined-active case needs its own CTA.

## Test plan

- [x] New regression test `VocabFilterDeadend.test.ts` (5 assertions): verifies `Clear all filters` text exists, both setters are called, single-filter branches still present
- [x] Updated capture window in existing `VocabFilteredEmptyState.test.ts` (regex `{0,2500}` → `{0,3500}`) — the block grew due to the new branch
- [x] Full frontend suite: 3813 tests pass

Closes #2389

### DIFF
--- a/frontend/src/__tests__/VocabFilterDeadend.test.ts
+++ b/frontend/src/__tests__/VocabFilterDeadend.test.ts
@@ -1,0 +1,46 @@
+/**
+ * Regression #2389: vocabulary empty-state must show "Clear all filters" CTA
+ * when both a tag filter and a search term are active simultaneously.
+ * Previously only one CTA was shown (tag filter took precedence), leaving
+ * users in a dead-end if clearing the tag still returned zero results.
+ */
+import fs from "fs";
+import path from "path";
+
+const src = fs.readFileSync(
+  path.resolve(__dirname, "../app/vocabulary/page.tsx"),
+  "utf8"
+);
+
+describe("Vocabulary page empty-state filter dead-end (issue #2389)", () => {
+  it("handles both selectedTag and search active: source references both conditions", () => {
+    // The empty-state block should contain logic for both-active case
+    const emptyStateBlock = src.slice(
+      src.indexOf("filtered.length === 0"),
+      src.indexOf("filtered.length === 0") + 2000
+    );
+    // Must have a branch that handles selectedTag && search both truthy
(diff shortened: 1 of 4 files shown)


alfmunny:
**PM**: PM-approved.

Real UX dead-end fix. Specific failure mode: user has both tag filter + search active, gets 0 results, the previous code only showed the "clear tag" CTA (because `if selectedTag` took precedence). Clearing the tag could still leave 0 results if the search was restrictive — no "clear search" button visible — dead-end. Adding a `selectedTag && search` branch with a single "Clear all filters" CTA resolves it cleanly.

Narrow trigger (both filters active simultaneously + zero matches) but a real dead-end when it hits. Single-filter branches preserved.

Clean +60/-2 across 3 files (page +13/-1, new test +46, existing test window bump). Branched off main, disjoint from #1801.

Applying P3 (real UX bug fix on a niche but legitimate dead-end state).

PR #2392, opened 2026-04-30
fix: add live character counters to deck form name and description inputs

### LINKED ISSUE #2391: ux: deck form inputs have silent maxLength with no character counter feedback
## Problem

The "New Deck" form has character limits (`maxLength={80}` for name, `maxLength={500}` for description) but no visible counter. When users approach or hit the limit, the input silently stops accepting characters with no warning or feedback.

This violates WCAG 3.3.1 (Error Identification) in spirit and is a clear usability gap for users creating decks with descriptive names.

**File:** `frontend/src/app/decks/new/page.tsx` lines 97–129

## Fix

Add a live character counter beneath each field that shows remaining characters, turning amber/red as the limit approaches (e.g., `72/80` → turns red at <10 remaining). Counter should use `aria-live="polite"` so screen readers announce changes.

## Acceptance

- Name field: counter appears below input showing `X/80`
- Description field: counter appears below textarea showing `X/500`
- Counter turns red when ≤10 characters remaining
- Counter has `aria-live="polite"` for screen reader announcements

alfmunny:
## What changed

Added live character counters beneath the deck name input (`maxLength={80}`) and description textarea (`maxLength={500}`) in the New Deck form.

- Counter shows `X/limit` right-aligned below each field
- Turns red at ≥70/80 chars (name) and ≥490/500 chars (description)
- Uses `aria-live="polite"` for screen reader announcements
- Input `aria-describedby` links to the counter element

## Why

Previously both fields silently stopped accepting characters at their limit with no user feedback — a clear usability gap (WCAG 3.3.1 in spirit).

## Test plan

- [x] New regression test `DeckFormCharCounter.test.ts` (6 assertions): verifies counters exist, aria-live present, red color class near limit
- [x] Full frontend suite: 3814 tests pass

Closes #2391

### DIFF
--- a/frontend/src/__tests__/DeckFormCharCounter.test.ts
+++ b/frontend/src/__tests__/DeckFormCharCounter.test.ts
@@ -1,0 +1,41 @@
+/**
+ * Regression #2391: deck form name and description inputs must display
+ * live character counters so users know when they're approaching maxLength.
+ * Previously both fields had silent maxLength with no feedback.
+ */
+import fs from "fs";
+import path from "path";
+
+const src = fs.readFileSync(
+  path.resolve(__dirname, "../app/decks/new/page.tsx"),
+  "utf8"
+);
+
+describe("Deck form character counters (issue #2391)", () => {
+  it("name input still has maxLength={80}", () => {
+    expect(src).toContain("maxLength={80}");
+  });
+
+  it("description textarea still has maxLength={500}", () => {
+    expect(src).toContain("maxLength={500}");
+  });
+
+  it("shows a character counter for the name field", () => {
+    // Counter should reference name.length and 80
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Real UX polish + a11y win. Previously both deck form fields silently capped at maxLength with no feedback — users would type, hit the wall, and not know why. Live counter `X/limit` (red near limit) is a standard fix. Properly equipped:
- `aria-live="polite"` so AT users get announcements
- `aria-describedby` from each input to its counter (correct association)
- Color thresholds at 70/80 + 490/500 (sensible warning band)

Clean +57/-1 across 2 files (page +16/-1, test +41). Branched off main, disjoint from #2390 (different file). 

Applying P3 (polish on a single form, real but not core-blocking).

PR #2395, opened 2026-04-30
fix: show Start-reading CTA on deck detail page when user has no vocabulary

### LINKED ISSUE #2394: ux: deck detail page 'Add word' gives no guidance when user has no vocabulary
## Problem

On the deck detail page (`/decks/[deckId]`), when a user has **no saved vocabulary words at all**, the "Add word" button is shown but disabled (greyed out, cursor-not-allowed). The user has no way to understand why — the copy says "Add words from your vocabulary to study them as a focused set", but doesn't explain that they need to first save words while reading.

This creates a **dead-end state**: new users who create a deck before saving any vocabulary see a disabled button with no explanation and no path forward.

## Reproduction

1. Create a new account (or clear vocabulary)
2. Create a new manual deck
3. The "Add word" button is disabled — no tooltip, no link to start reading and saving words

## Expected behaviour

When `vocab.length === 0`, the empty state and "Add word" button area should:
- Replace the disabled "Add word" button with a "Start reading" or "Go to library" CTA
- Add a secondary line like "Save words while reading to add them here"

When `vocab.length > 0` but `candidateWords.length === 0` (all words already in the deck), keep the current disabled button — that case is understandable from context.

## Proposed fix

In the deck detail page, distinguish the two cases:
1. `vocab.length === 0` → show "No vocabulary saved yet" message + link to library
2. `vocab.length > 0 && candidateWords.length === 0` → keep "All words already in deck" message with disabled button

alfmunny:
## What changed

On the deck detail page, when a user has **no saved vocabulary at all**, the "Add word" button was silently disabled (greyed out, cursor-not-allowed) with no explanation. Users — especially new ones — had no indication of why the button was unavailable or what to do next.

This PR distinguishes two cases:

| State | Before | After |
|---|---|---|
| `vocab.length === 0` (no saved words) | Disabled "Add word" button, misleading description | "Start reading" CTA → library; message: "No vocabulary saved yet. Save words while reading to add them here." |
| `vocab.length > 0 && candidateWords.length === 0` (all in deck) | Disabled "Add word" button | Same disabled button (context makes reason clear) |

The header "Add word" button in the top-right also now routes to the library instead of doing nothing when vocab is empty.

## Why

Dead-end state: new users who create a deck before saving any vocabulary see a non-actionable button with no escape path.

## Test plan

- [x] New regression test `DeckNoVocabDeadEnd.test.ts` (3 assertions) — confirms page branches on `vocab.length` and shows reading CTA when empty
- [x] All 31 existing deck test suites pass (173 tests)
- [x] Full frontend suite: 3822 tests pass

Closes #2394

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -436,6 +436,7 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | Profile TTS gender + translation provider radio groups: role="radiogroup" + aria-labelledby — WCAG 1.3.1 (closes #2385) | profile/page.tsx | #2386 |
 | 2026-04-30 | Vocabulary empty-state: added "Clear all filters" CTA when both tag filter and search are active simultaneously — prevents dead-end with zero results (closes #2389) | vocabulary/page.tsx | #2390 |
 | 2026-04-30 | Deck form: live character counters (X/80, X/500) with aria-live="polite" on name and description inputs — WCAG 3.3.1 spirit (closes #2391) | decks/new/page.tsx | #2392 |
+| 2026-04-30 | Deck detail page: distinguish zero-vocab state from all-in-deck state — show "Start reading" CTA with guidance when user has no vocabulary, instead of a silent disabled "Add word" button (closes #2394) | decks/[deckId]/page.tsx | #2394 |
 
 ---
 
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Real onboarding/dead-end UX fix. New users hit this exact state: create a deck → open it → "Add word" button greyed out with no context. The fix correctly distinguishes the two empty states:
- `vocab.length === 0` (truly empty) → "Start reading" CTA + helpful explanation
- `vocab.length > 0 && candidateWords.length === 0` (all in deck) → leave disabled (context already clear)

Bonus the design-log Change Log entry is in the same PR rather than retro-batched — proper per-PR docs check this time. Better than the recent #2393 retro pattern.

Header "Add word" button also rerouted to library when no vocab — covers the second escape path.

Clean +64/-14 across 3 files (page +28/-14, test +35, design log +1). Branched off main, disjoint from #1801.

Applying P3 (real dead-end fix on new-user flow; worth shipping but not core-blocking since global nav still gives an escape).

PR #2365, opened 2026-04-30
fix: add aria-invalid + aria-describedby to Obsidian form inputs (closes #2364)

alfmunny:
## What changed

Three changes to `profile/page.tsx` (Obsidian settings section):

1. Added `id="obsidian-msg"` to the existing `<p role="status" aria-live="polite">` element.
2. Added `aria-invalid={obsidianMsg?.ok === false || undefined}` to the token, repo, and vault-path inputs.
3. Added `aria-describedby="obsidian-msg"` to all three inputs.

Also bumped the search window in the existing `ProfileObsidianErrorBorder.test.ts` from 600 → 900 chars to accommodate the two new attributes that now precede `className`.

## Why

WCAG 3.3.1 (Error Identification) and 4.1.2 (Name, Role, Value): inputs that can enter an invalid state must convey that state to assistive technology. The red border change was visual-only — screen readers had no signal the field was invalid when focus returned to it after the error was announced.

The Gemini key input on the same page already followed this pattern correctly; Obsidian inputs now match.

## Test plan

- New: `ProfileObsidianAriaInvalid.test.ts` — 5 static tests verifying `aria-invalid`, `aria-describedby`, and `id="obsidian-msg"` are present
- Existing: `ProfileObsidianErrorBorder.test.ts` — all 4 tests pass (window enlarged, no logic change)
- Full frontend suite: 3765 tests pass

Closes #2364

PR #2397, opened 2026-04-30
fix: remove blanket aria-invalid from Obsidian form fields on generic save error (closes #2396)

### LINKED ISSUE #2396: ux: profile Obsidian form marks all three fields aria-invalid on generic save error
## Problem

When saving Obsidian settings fails (e.g. network error), all three input fields — GitHub token, Obsidian Repo, and Vault Path — receive `aria-invalid="true"` simultaneously, even though the error is not field-specific.

**WCAG violation:** WCAG 1.3.1 (Info and Relationships) — `aria-invalid` must only be set on a field whose value is actually invalid. A generic API/network failure is not a per-field validation error.

**Screen reader impact:** Assistive technologies announce each focused field as "invalid", leading a user to believe they have entered bad data into three separate fields, when in reality the issue is a save failure unrelated to input values.

## Reproduction

1. Open Profile → Obsidian Export.
2. Fill in all three fields with valid-looking data.
3. Force a network error (e.g. disconnect Wi-Fi, or have the API return 500).
4. Observe: all three fields receive red borders AND `aria-invalid=true`.

## Current code (`profile/page.tsx` lines 410, 429, 445)

```tsx
aria-invalid={obsidianMsg?.ok === false || undefined}
```

This is applied identically to all three fields — no per-field validation logic.

## Fix

Remove `aria-invalid` from all three fields. The existing `role="status" aria-live="polite"` region (`id="obsidian-msg"`) already announces the error to screen reader users — that is the correct mechanism for a form-level error. Also remove the conditional red border styling from individual fields (apply it to the fieldset or a wrapper instead, or leave just the status message).

Keep `aria-describedby="obsidian-msg"` on all three fields so users can still navigate to the status message.

## Files

- `frontend/src/app/profile/page.tsx` lines 410, 429, 445

alfmunny:
## What changed

Removed `aria-invalid` from all three Obsidian export form fields (`obsidian-token`, `obsidian-repo`, `obsidian-path`) in `profile/page.tsx`. Also removed the conditional `border-red-400 / focus:ring-red-400` per-field error styling.

## Why

**WCAG 1.3.1 (Info and Relationships) violation.** `aria-invalid` must only appear on a field whose *value* is invalid. A generic network/API save failure is not a per-field validation error — every field was being marked invalid simultaneously even when the user entered perfectly valid data.

**Screen reader impact:** With the old code, when save failed, a screen reader user focusing into any of the three fields would hear it announced as "invalid". This falsely signals their input is wrong, not that a server error occurred.

**Fix:** The existing `role="status" aria-live="polite"` region (`id="obsidian-msg"`) already announces the save error correctly. Fields keep `aria-describedby="obsidian-msg"` so users can navigate to the message. No new machinery needed.

## Test plan

- `ObsidianAriaInvalid.test.ts` — new regression test confirming ≤1 field gets aria-invalid (now 0)
- `ProfileObsidianAriaInvalid.test.ts` — updated to assert fields do NOT have `aria-invalid` and DO have `aria-describedby`
- `ProfileObsidianErrorBorder.test.ts` — updated to assert static border classes (no red conditional)
- All 3827 frontend tests pass

## Docs follow-up

Closes #2396

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -437,6 +437,7 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | Vocabulary empty-state: added "Clear all filters" CTA when both tag filter and search are active simultaneously — prevents dead-end with zero results (closes #2389) | vocabulary/page.tsx | #2390 |
 | 2026-04-30 | Deck form: live character counters (X/80, X/500) with aria-live="polite" on name and description inputs — WCAG 3.3.1 spirit (closes #2391) | decks/new/page.tsx | #2392 |
 | 2026-04-30 | Deck detail page: distinguish zero-vocab state from all-in-deck state — show "Start reading" CTA with guidance when user has no vocabulary, instead of a silent disabled "Add word" button (closes #2394) | decks/[deckId]/page.tsx | #2394 |
+| 2026-04-30 | Profile Obsidian form: remove aria-invalid from all three fields on generic save error — WCAG 1.3.1 violation fixed; role="status" region (obsidian-msg) is the correct channel for form-level errors; aria-describedby links fields to it (closes #2396) | profile/page.tsx | #2396 |
 
 ---
 
--- a/frontend/src/__tests__/ObsidianAriaInvalid.test.ts
+++ b/frontend/src/__tests__/ObsidianAriaInvalid.test.ts
@@ -1,0 +1,35 @@
+/**
+ * Regression test for issue #2396: Obsidian form marks all three fields
+ * aria-invalid on a generic save error — a WCAG 1.3.1 violation.
+ */
+import fs from "fs";
+import path from "path";
+
+function read(rel: string): string {
+  return fs.readFileSync(path.join(process.cwd(), rel), "utf8");
+}
+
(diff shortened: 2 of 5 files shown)


alfmunny:
**PM**: PM-approved, with apology — this corrects an over-application of `aria-invalid` that I missed when approving #2365 earlier this session.

The author's analysis is exactly right. WCAG 1.3.1 specifies `aria-invalid` should only appear on fields whose *value* is invalid (wrong format, validation failure, etc.). Generic API/network/permission save failures aren't per-field validation errors. The old code marked all three fields invalid even when the user's input was perfectly valid — actively misleading for screen reader users who'd hear "invalid" on every field.

The existing `role="status" aria-live="polite"` region from #2365 already handles generic-failure announcement correctly. Keeping `aria-describedby` link from each field to that region is the right shape — users navigating into fields can still discover the error. Removing the conditional red-border styling alongside is also correct (visual = WCAG behavioral parity, generic failure shouldn't visually flag fields as wrong).

Clean +92/-32 across 5 files (profile +3/-6, 1 new test, 2 test updates, 1 design log entry). Branched off main, disjoint from #1801. Docs Change Log entry inline (proper per-PR check, like #2395).

Applying P2 (real WCAG correctness fix; correcting misleading AT announcement is materially impactful even if narrow surface).

PR #2399, opened 2026-04-30
fix: announce word addition via aria-live in deck add-word picker (closes #2398)

### LINKED ISSUE #2398: ux: add-word picker announces nothing when a word is added to deck (WCAG 4.1.3)
## Problem

When a user clicks a word in the "Add word" picker on the deck detail page, the word is added to the deck and disappears from the picker list — but no status message is announced to screen readers.

**WCAG 4.1.3 (Status Messages):** Programmatically determined status changes that don't receive focus must be surfaced via `role="status"` or similar so assistive technologies can read them aloud.

Compare with word *removal*: that flow has an `UndoToast` with `role="status" aria-live="polite"` that announces the removal. The addition flow has no equivalent.

## Reproduction

1. Open a manual deck with vocabulary words available to add.
2. Click "Add word" — the picker opens.
3. Using a screen reader (or check in code), click a word.
4. The word disappears from the list — no announcement is made.

## Fix

Add a `role="status" aria-live="polite"` region inside the `AddWordPicker` dialog that announces when a word is added. Track the last added word name in local state; render a `sr-only` live region that updates when a word is added.

Example text: `"German" added to deck` (using the word text).

## Files

- `frontend/src/app/decks/[deckId]/page.tsx` — `AddWordPicker` component (~line 336)

alfmunny:
## What changed

Added a `role="status" aria-live="polite"` region inside the `AddWordPicker` dialog in `decks/[deckId]/page.tsx`. When the user clicks a word to add it to the deck, local `lastAdded` state is updated and the live region announces `"${word}" added to deck`.

## Why

**WCAG 4.1.3 (Status Messages):** When a word is added from the picker, it visually disappears from the list — but screen readers receive no announcement. Without a live region, keyboard and screen-reader users can only discover the addition by navigating the updated list (noticing the word is gone), not by hearing an immediate confirmation.

This mirrors the existing `UndoToast` pattern used for word *removal*, which already has `role="status" aria-live="polite"`.

## Test plan

- `DeckPickerAddAnnouncement.test.ts` — 3 new source-level assertions confirming:
  - `lastAdded` state exists in `AddWordPicker`
  - A `role="status" aria-live="polite"` region is present
  - The region references `lastAdded`
- Full suite: 3830 tests pass

## Docs follow-up

Closes #2398

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -438,6 +438,7 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | Deck form: live character counters (X/80, X/500) with aria-live="polite" on name and description inputs — WCAG 3.3.1 spirit (closes #2391) | decks/new/page.tsx | #2392 |
 | 2026-04-30 | Deck detail page: distinguish zero-vocab state from all-in-deck state — show "Start reading" CTA with guidance when user has no vocabulary, instead of a silent disabled "Add word" button (closes #2394) | decks/[deckId]/page.tsx | #2394 |
 | 2026-04-30 | Profile Obsidian form: remove aria-invalid from all three fields on generic save error — WCAG 1.3.1 violation fixed; role="status" region (obsidian-msg) is the correct channel for form-level errors; aria-describedby links fields to it (closes #2396) | profile/page.tsx | #2396 |
+| 2026-04-30 | Deck add-word picker: aria-live status region announces each word addition to screen readers — WCAG 4.1.3 (closes #2398) | decks/[deckId]/page.tsx | #2398 |
 
 ---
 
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Real WCAG 4.1.3 fix that closes a meaningful asymmetry: word *removal* already had a `role="status"` UndoToast announcement, but word *addition* was silent. AT users could click "add" successfully but get no confirmation — they'd have to navigate the picker list and notice the item was gone.

Implementation matches the existing UndoToast announcement pattern (correct), and the live region is always-present (doesn't repeat the conditional-mount mistake the WCAG 4.1.3 sweep already corrected).

Inline design-log entry — proper per-PR docs check. Same good pattern as #2395.

Clean +41/-0 across 3 files (page +6, test +34, design log +1). Branched off main, disjoint from #1801.

Applying P3 (real WCAG AA fix; narrow surface but matters for AT users actually using the picker).

PR #2401, opened 2026-04-30
fix: add aria-valuetext to TTS position and speed sliders (closes #2400)

### LINKED ISSUE #2400: ux: TTS sliders missing aria-valuetext — screen readers announce raw numbers instead of formatted values (WCAG 4.1.2)
## Problem

Two range inputs in `TTSControls.tsx` are missing `aria-valuetext`, causing screen readers to announce raw numeric values that are hard to interpret.

### Playback position slider (line 476)
- Has `aria-label="Playback position"` but no `aria-valuetext`
- Screen readers say "152.3" instead of "2:32"
- Fix: add `aria-valuetext={formatTime(globalCurrentTime)}`

### Speed slider (line 493)
- Is inside a `<label>Speed</label>` so it is named, but has no `aria-valuetext`  
- Screen readers say "1.0" instead of "1.0×"
- Fix: add `aria-label="Playback speed"` and `aria-valuetext={`${rate.toFixed(1)}×`}`

## WCAG
WCAG 4.1.2 — Name, Role, Value: slider value must be expressed in a form that assistive technology can present to the user. A raw float is not meaningful; the formatted equivalent is.

## Reproduction
Open reader with TTS enabled → focus the position or speed slider with keyboard → listen to the screen reader announcement.

## Fix
Add `aria-valuetext` to both sliders in `frontend/src/components/TTSControls.tsx`.

alfmunny:
## What changed

Added `aria-valuetext` to both range inputs in `TTSControls.tsx` and an explicit `aria-label` to the speed slider.

- **Playback position slider**: `aria-valuetext={formatTime(globalCurrentTime)}` — screen readers now announce "2:32" instead of "152.3"
- **Speed slider**: `aria-label="Playback speed"` + `aria-valuetext={`${rate.toFixed(1)}×`}` — screen readers now announce "1.0×" instead of "1.0"

## Why

WCAG 4.1.2 — Name, Role, Value requires that slider values be expressed in a form assistive technology can present meaningfully. Raw floats (seconds, unitless multipliers) are not meaningful to screen reader users.

## Test plan

- New regression tests in `TTSSliderAriaValuetext.test.ts` assert both sliders carry `aria-valuetext` referencing `formatTime` / the `×` symbol
- All 3833 Jest tests pass

Closes #2400

### DIFF
--- a/frontend/src/__tests__/TTSSliderAriaValuetext.test.ts
+++ b/frontend/src/__tests__/TTSSliderAriaValuetext.test.ts
@@ -1,0 +1,42 @@
+/**
+ * Regression test for issue #2400: TTS sliders missing aria-valuetext.
+ * Screen readers announced raw floats ("152.3") instead of formatted values
+ * ("2:32" for position, "1.0×" for speed). WCAG 4.1.2 — Name, Role, Value.
+ */
+import fs from "fs";
+import path from "path";
+
+const src = fs.readFileSync(
+  path.join(process.cwd(), "src/components/TTSControls.tsx"),
+  "utf8",
+);
+
+describe("TTSControls sliders — aria-valuetext (issue #2400)", () => {
+  it("playback position slider has aria-valuetext with formatted time", () => {
+    const posSlider = src.indexOf('aria-label="Playback position"');
+    expect(posSlider).toBeGreaterThan(0);
+    // Grab the surrounding input block (up to 300 chars back to include the whole tag)
+    const block = src.slice(Math.max(0, posSlider - 300), posSlider + 300);
+    expect(block).toMatch(/aria-valuetext/);
+    // Should reference formatTime, not a raw number
+    expect(block).toMatch(/formatTime/);
+  });
+
+  it("speed slider has aria-label", () => {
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Real WCAG 4.1.2 fix on TTS sliders. Without `aria-valuetext`, AT users heard raw values that don't carry meaning at the listening level: "152.3" instead of "2:32" for position; "1.0" instead of "1.0×" for speed. The format-respecting strings match what sighted users see — that's the right shape.

Bonus: speed slider also gets an explicit `aria-label="Playback speed"` — implies it had no accessible name at all before. Worth noting but doesn't change the priority since both fixes ship together.

Clean +45/-0 across 2 files (component +3, test +42). Branched off main, disjoint from #1801.

No design-log entry this time. Not blocking — a single ARIA attribute on internal media controls doesn't necessarily warrant a Change Log row; the recent #2395/#2399/#2397 inline entries were for more user-visible UX changes. PM is fine with the omission.

Applying P3 (real WCAG AA on TTS sliders; narrow surface but real).

PR #2403, opened 2026-04-30
fix: announce annotation and highlight saves to screen readers (closes #2402)

### LINKED ISSUE #2402: ux: reader annotation/highlight save has no screen-reader success announcement (WCAG 4.1.3)
## Problem

When a user saves an annotation or applies a quick highlight in the reader, the toolbar/dialog closes with no screen-reader announcement. WCAG 4.1.3 requires that status messages that don't receive focus be communicated via an appropriate live region.

### Affected flows
1. **AnnotationToolbar** (note editor dialog) — clicking "Save note" saves successfully and closes the dialog. Screen readers hear nothing after the dialog disappears.
2. **QuickHighlightPanel** (color toolbar) — clicking a color applies the highlight and closes the panel. Screen readers hear nothing.

There IS undo-toast feedback for deletion (UndoToast with role=status), but no equivalent for creation/update.

## Fix

In `frontend/src/app/reader/[bookId]/page.tsx`:
1. Add `savedAnnotationMsg` state string.
2. In both `onSaved` handlers (AnnotationToolbar + QuickHighlightPanel), set the msg to "Note saved" / "Highlight applied" and clear it after 2s.
3. Add an sr-only `role="status" aria-live="polite"` region that renders `savedAnnotationMsg`.

No UI change needed — announcement only.

## WCAG
4.1.3 — Status Messages: "In content implemented using markup languages, status messages can be programmatically determined through role or properties such that they can be presented to the user by assistive technologies without receiving focus."

alfmunny:
## What changed

Added a screen-reader success announcement for annotation and highlight saves in the reader page.

- **New state**: `savedAnnotationMsg` tracks the current announcement text
- **AnnotationToolbar `onSaved`**: sets msg to `"Note saved"` (or `"Highlight saved"` for color-only changes), cleared after 2 s
- **QuickHighlightPanel `onSaved`**: sets msg to `"Highlight applied"`, cleared after 2 s
- **Always-present live region**: `<span role="status" aria-live="polite" aria-atomic="true" className="sr-only">{savedAnnotationMsg}</span>` placed alongside the existing obsidian toast mirror — always in the DOM so AT announces on first fire

No visual change — announcement only.

## Why

WCAG 4.1.3 — Status Messages: when a dialog/toolbar closes after a successful save, screen readers hear nothing. The deletion flow already had `UndoToast` (role=status) but the creation/update flow had no equivalent.

## Test plan

- New regression tests in `ReaderAnnotationSaveAnnouncement.test.ts`:
  - `savedAnnotationMsg` state variable present
  - `role="status"` sr-only live region references `savedAnnotationMsg`
  - Both `onSaved` handlers call `setSavedAnnotationMsg`
- All 3837 Jest tests pass

Closes #2402

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -439,6 +439,8 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | Deck detail page: distinguish zero-vocab state from all-in-deck state — show "Start reading" CTA with guidance when user has no vocabulary, instead of a silent disabled "Add word" button (closes #2394) | decks/[deckId]/page.tsx | #2394 |
 | 2026-04-30 | Profile Obsidian form: remove aria-invalid from all three fields on generic save error — WCAG 1.3.1 violation fixed; role="status" region (obsidian-msg) is the correct channel for form-level errors; aria-describedby links fields to it (closes #2396) | profile/page.tsx | #2396 |
 | 2026-04-30 | Deck add-word picker: aria-live status region announces each word addition to screen readers — WCAG 4.1.3 (closes #2398) | decks/[deckId]/page.tsx | #2398 |
+| 2026-04-30 | TTS sliders: add aria-valuetext to playback position slider (formatted time, e.g. "2:32") and speed slider ("1.0×"); add aria-label to speed slider — WCAG 4.1.2 (closes #2400) | components/TTSControls.tsx | #2401 |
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Real WCAG 4.1.3 fix on annotation save flow. Closes the same kind of asymmetry as #2399 — delete already had `UndoToast` announcement, but save (annotation/highlight) was silent. AT users would close the toolbar/panel and have no signal their save succeeded.

Three thoughtful messages differentiate the action: "Note saved", "Highlight saved" (color-only), "Highlight applied" (QuickHighlightPanel). Always-present live region (matches the established WCAG 4.1.3 pattern). 2s clear is reasonable.

Inline design-log entry — proper per-PR docs check, like #2395/#2399.

Clean +61/-0 across 3 files (page +11, test +48, design log +2). Branched off main, disjoint from #1801.

Applying P3 (real WCAG AA fix on primary annotation save flow).

PR #2405, opened 2026-04-30
fix: announce SelectionToolbar actions to screen readers on text selection (closes #2404)

### LINKED ISSUE #2404: ux: SelectionToolbar appears with no screen-reader announcement when text selected via keyboard (WCAG 4.1.3)
## Problem

When the user selects text via keyboard (Shift+arrow keys), the SelectionToolbar (Read / Highlight / Note / Chat / Word actions) appears visually but makes no announcement to screen readers. WCAG 4.1.3 requires that such status changes be communicated via an appropriate live region.

The toolbar has `role="toolbar"` and all buttons have `aria-label`, but the entire component returns `null` when there is no selection — so there is no persistent live region that can fire on the `selection → visible` transition.

## Reproduction

1. Open the reader on any book.
2. Use Shift+arrow keys to select a word or phrase.
3. Screen reader hears nothing — no indication that Read/Highlight/Note/Chat/Word actions are available.

## Fix

Refactor `SelectionToolbar.tsx` to always render an `aria-live="polite"` sr-only region. When `selection` becomes truthy, the region announces "Text selected. Press Tab for actions: Read, Highlight, Note, Chat, Look up word."

The component currently does `if (!selection) return null` — change this to return a fragment: always render the live region (so it can fire), and conditionally render the visual toolbar.

## WCAG
4.1.3 — Status Messages

alfmunny:
## What changed

`SelectionToolbar` now always renders a `role="status" aria-live="polite"` live region in the DOM, even when no text is selected. When a selection becomes active, the region announces "Text selected. Press Tab for actions: Read, Highlight, Note, Chat, Look up word" so keyboard and AT users know which actions are available.

Previously the component returned `null` (or only the toolbar) when there was no selection, which prevented screen readers from picking up the state change when selection became active — violating WCAG 4.1.3 (Status Messages).

## Why

Keyboard users who select text via Shift+Arrow have no way to discover the toolbar exists without a live announcement. The live region must be in the DOM *before* content changes for AT to fire — returning `null` silently dropped the announcement.

## Test plan

- `SelectionToolbarAnnouncement.test.ts` (new) — 3 source-level regex tests verifying: `aria-live="polite"` present, announcement text references Tab and actions, component no longer returns `null` for no-selection state
- `SelectionToolbar.test.tsx` (updated) — existing "renders nothing when no selection" test updated to check `role="toolbar"` is absent rather than `container.firstChild` is null (now the sr-only span is always present)
- Full frontend suite: 3840 tests pass

Closes #2404

## Docs follow-up

### DIFF
--- a/frontend/src/__tests__/SelectionToolbar.test.tsx
+++ b/frontend/src/__tests__/SelectionToolbar.test.tsx
@@ -58,12 +58,12 @@ describe("SelectionToolbar", () => {
     jest.restoreAllMocks();
   });
 
-  it("renders nothing when there is no selection", () => {
-    const { container } = render(
+  it("renders no toolbar buttons when there is no selection", () => {
+    render(
       <SelectionToolbar onRead={jest.fn()} onHighlight={jest.fn()} />
     );
-    // Toolbar div not in document
-    expect(container.firstChild).toBeNull();
+    // No action buttons should be visible — only the sr-only live region remains
+    expect(document.querySelector('[role="toolbar"]')).not.toBeInTheDocument();
   });
 
   it("shows toolbar when text >= 2 chars is selected inside reader", () => {
--- a/frontend/src/__tests__/SelectionToolbarAnnouncement.test.ts
+++ b/frontend/src/__tests__/SelectionToolbarAnnouncement.test.ts
@@ -1,0 +1,35 @@
+/**
+ * Regression test for issue #2404: SelectionToolbar appeared with no
+ * screen-reader announcement when text was selected via keyboard (WCAG 4.1.3).
+ *
+ * When the toolbar mounts (selection becomes truthy), an always-present
+ * sr-only aria-live="polite" region must announce the available actions so
+ * AT users know they can Tab to Read / Highlight / Note / Chat / Word buttons.
+ */
+import fs from "fs";
+import path from "path";
+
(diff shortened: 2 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Real WCAG 4.1.3 fix on a primary reading-flow control. Without the announcement, keyboard users selecting text via Shift+Arrow had no way to discover the toolbar exists — they'd have to know about it and Tab into it blindly. The new announcement spells out the actions explicitly: "Text selected. Press Tab for actions: Read, Highlight, Note, Chat, Look up word" — that's actionable copy.

Always-present pattern is correct (matches the established WCAG 4.1.3 sweep). Test refactor handles the new shape correctly: existing "renders nothing when no selection" test now checks for absence of `role="toolbar"` rather than `container.firstChild` being null, which is the right semantic.

Note: PR body has `[ ] Docs updated (if applicable)` unchecked. A Change Log entry in `docs/design-improvement-plan.md` would be appropriate (matches the recent #2395/#2399/#2403 inline pattern). Not blocking but worth a follow-up bundle in the next docs-batch.

Clean +61/-5 across 3 files. Branched off main, disjoint from #1801.

Applying P3 (real WCAG AA fix on a primary keyboard reading flow).

PR #2407, opened 2026-04-30
fix: prevent AnnotationToolbar dismissal during save/delete (closes #2406)

### LINKED ISSUE #2406: ux: AnnotationToolbar close button and backdrop dismissible during save/delete (data-loss risk)
## Problem

The note editor modal (`AnnotationToolbar`) can be dismissed mid-operation:

- **Close button** (`frontend/src/components/AnnotationToolbar.tsx` line 130): calls `onClose()` with no guard against `saving` or `deleting` state.
- **Backdrop click** (line 110): same — `onClick={onClose}` fires even while a save or delete is in flight.

A user who clicks the backdrop or close button during an async save loses their note changes without any warning.

## Reproduction

1. Open the note editor on a highlighted sentence.
2. Type a note and quickly click the backdrop (or close X) before the save API call completes.
3. The modal closes; the annotation is saved (if the request completed) or silently lost (if it was in flight).

## Fix

Add `disabled={saving || deleting}` to the close button and guard backdrop clicks with the same condition:
```tsx
// Close button (line 130)
<button onClick={onClose} disabled={saving || deleting} ...>

// Backdrop (line 110)
<div ... onClick={() => { if (!saving && !deleting) onClose(); }} />
```

Also add `cursor-not-allowed opacity-50` to the close button's disabled state so the blocked state is visible.

alfmunny:
## What changed

`AnnotationToolbar` close button and backdrop now refuse to dismiss the modal while a save or delete operation is in flight.

**Before:** clicking the backdrop or close X mid-save silently closed the modal — the user lost any changes if the request was still in flight.

**After:**
- Close button gets `disabled={saving || deleting}` + `disabled:opacity-40 disabled:cursor-not-allowed` so the blocked state is visually obvious.
- Backdrop `onClick` is guarded: `() => { if (!saving && !deleting) onClose(); }`.
- The outer wrapper `onClick` has the same guard.

## Why

A user who types a long note and quickly clicks away before the API call completes would lose their note silently. This is a data-loss risk pattern — the fix matches the existing `disabled={saving}` on the Save button.

## Test plan

`AnnotationToolbarCloseGuard.test.tsx` (new — 4 tests):
- Close button is `disabled` while `createAnnotation` is pending
- Backdrop click does not call `onClose` while save is in flight
- Close button is `disabled` while `deleteAnnotation` is pending
- Close button is re-enabled (and `onSaved` called) after save resolves

Full frontend suite: 3841 tests pass.

Closes #2406

## Docs follow-up
- [x] Docs updated — added row to `docs/design-improvement-plan.md` Change Log

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -441,6 +441,8 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | Deck add-word picker: aria-live status region announces each word addition to screen readers — WCAG 4.1.3 (closes #2398) | decks/[deckId]/page.tsx | #2398 |
 | 2026-04-30 | TTS sliders: add aria-valuetext to playback position slider (formatted time, e.g. "2:32") and speed slider ("1.0×"); add aria-label to speed slider — WCAG 4.1.2 (closes #2400) | components/TTSControls.tsx | #2401 |
 | 2026-04-30 | Reader annotation/highlight save: sr-only role=status live region announces "Note saved" / "Highlight applied" after AnnotationToolbar or QuickHighlightPanel saves — WCAG 4.1.3 (closes #2402) | reader/[bookId]/page.tsx | #2403 |
+| 2026-04-30 | SelectionToolbar: always renders sr-only aria-live="polite" live region; announces available actions when text selection activates — prevents silent WCAG 4.1.3 violation for keyboard selection (closes #2404) | components/SelectionToolbar.tsx | #2405 |
+| 2026-04-30 | AnnotationToolbar: close button disabled (opacity-40, cursor-not-allowed) and backdrop click no-op while save/delete in flight — prevents silent data loss risk (closes #2406) | components/AnnotationToolbar.tsx | #2407 |
 
 ---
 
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Real data-loss bug fix. Specific failure: user types a long note → clicks Save → immediately clicks backdrop/close X (impatient or distracted) → if the API call hasn't resolved, the note is lost silently. Worse: user thought they saved, so they don't realize the data vanished until they revisit and find no annotation.

The fix is precisely scoped:
- Close button `disabled={saving || deleting}` + visual `opacity-40` / `cursor-not-allowed` — visible feedback when blocked
- Backdrop + wrapper `onClick` guarded against in-flight operation
- Matches the existing `disabled={saving}` on Save button — symmetric protection

Test plan covers the right shapes: pending state blocks close, completed state re-enables, both save and delete paths protected.

Inline design-log entry — proper per-PR docs check (matches the recent #2395/#2399/#2403/#2407 inline pattern).

Clean +94/-3 across 3 files (component +4/-3, test +88, design log +2). Branched off main, disjoint from #1801.

Applying P2 (real data-loss fix on a primary annotation flow — silent loss of typed user content is materially worse than typical UX polish).

PR #2409, opened 2026-04-30
fix: render InsightChat API errors as distinct alert bubbles (closes #2408)

alfmunny:
## What changed

`InsightChat` now renders API error responses as a distinct red alert bubble instead of a normal AI assistant message.

**Before:** When `getInsight` or `askQuestion` throws, the error was stored as `{ role: "assistant", content: "Error: ..." }` and rendered identically to a real AI reply — same amber avatar, same prose styling, mixed into the conversation.

**After:** Any assistant message whose content starts with `"Error:"` is detected (`msg.content.startsWith("Error:")`) and rendered as:
```tsx
<div role="alert" className="flex items-start gap-2 px-3 py-2 bg-red-50 border border-red-200 rounded-xl text-xs text-red-700">
  <AlertCircleIcon ... />
  <span>{msg.content.replace(/^Error:\s*/, "")}</span>
</div>
```
The `"Error: "` prefix is stripped so the message reads naturally (e.g. "Network connection failed" rather than "Error: Network connection failed").

## Why

Error messages like "Error: 429 Too Many Requests" or "Error: Model not found" appearing inside an amber AI bubble are easy to miss and confusing — the user may interpret them as cryptic AI output. With `role="alert"`, screen readers also announce the error immediately without the user tabbing to it.

## Test plan

`InsightChatErrorDisplay.test.ts` (new — 5 source-level regex tests):
- Component detects `startsWith("Error:")`
- Component renders `role="alert"` for error state
- Component uses red/rose colours for error bubble
- Component imports `AlertCircleIcon`
- Error branch is distinct from normal prose/ReactMarkdown path

`InsightChat.coverage.test.tsx` (updated — 3 tests):
- Existing error text assertions updated to match stripped-prefix display (e.g. `/Gemini quota exceeded/` instead of `/Error: Gemini quota exceeded/`)

Full frontend suite: 3845 tests pass.

Closes #2408

## Docs follow-up
- [x] Docs updated — added row to `docs/design-improvement-plan.md` Change Log

PR #2411, opened 2026-04-30
fix: show inline error feedback in QuickHighlightPanel on save/delete failure (closes #2410)

### LINKED ISSUE #2410: ux: QuickHighlightPanel silently drops errors — user has no feedback when highlight save/delete fails
## Problem

`QuickHighlightPanel` (used for tap-to-highlight in the reader) has empty catch blocks for both color selection and delete:

```ts
// frontend/src/components/QuickHighlightPanel.tsx line 83-85
} catch {
  setBusy(false);  // error silently discarded
}

// line 95-97
} catch {
  setBusy(false);  // same
}
```

When the API call for `createAnnotation`, `updateAnnotation`, or `deleteAnnotation` fails:
1. The user taps a highlight color
2. Buttons disable briefly (looks like it's processing)
3. Buttons re-enable — no save happened, no error shown
4. User may tap again thinking it worked, or be confused why their highlight didn't appear

## Fix

Add an `error` state and show a small inline error message inside the panel when the operation fails. The panel should stay open for retry. Example:

```tsx
const [error, setError] = useState<string | null>(null);

// In catch:
catch { setBusy(false); setError("Save failed — tap to retry"); }

// In JSX (add below the buttons row):
{error && (
  <p role="alert" className="text-xs text-red-600 mt-1">{error}</p>
)}
```

Clear the error when a new action starts.

alfmunny:
## What changed

`QuickHighlightPanel` (the floating color picker that appears when tapping a sentence in the reader) now shows an inline error message when highlight save or delete fails.

**Before:** Both `handleColor` and `handleDelete` had empty catch blocks — on failure, buttons just re-enabled silently. The user had no idea their highlight wasn't saved.

**After:**
- Added `error` state (`useState<string | null>(null)`)
- `handleColor` catch: `setError("Save failed — tap a colour to retry.")`
- `handleDelete` catch: `setError("Delete failed — tap to retry.")`
- Error clears on each new action (`setError(null)` at the top of each handler)
- Panel stays open on failure so the user can retry
- Error renders below the color buttons as `role="alert"` with red styling — consistent with the rest of the app's error presentation

## Why

A user who taps a highlight colour while offline, or during a network blip, currently sees the panel appear to "process" and then go back to normal — with no indication that nothing was saved. The highlight silently fails. With this fix, the error is surfaced immediately and the user knows to try again.

## Test plan

`QuickHighlightPanelErrorFeedback.test.tsx` (new — 5 tests):
- `role="alert"` appears when `createAnnotation` rejects
- Alert text matches "fail|retry|error"
- `role="alert"` appears when `deleteAnnotation` rejects
- Error clears when a new action starts (second color tap succeeds)
- `onClose` is NOT called on failure (panel stays open for retry)

Full frontend suite: 3849 tests pass.

Closes #2410

## Docs follow-up
- [x] Docs updated — added row to `docs/design-improvement-plan.md` Change Log

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -443,6 +443,8 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | Reader annotation/highlight save: sr-only role=status live region announces "Note saved" / "Highlight applied" after AnnotationToolbar or QuickHighlightPanel saves — WCAG 4.1.3 (closes #2402) | reader/[bookId]/page.tsx | #2403 |
 | 2026-04-30 | SelectionToolbar: always renders sr-only aria-live="polite" live region; announces available actions when text selection activates — prevents silent WCAG 4.1.3 violation for keyboard selection (closes #2404) | components/SelectionToolbar.tsx | #2405 |
 | 2026-04-30 | AnnotationToolbar: close button disabled (opacity-40, cursor-not-allowed) and backdrop click no-op while save/delete in flight — prevents silent data loss risk (closes #2406) | components/AnnotationToolbar.tsx | #2407 |
+| 2026-04-30 | InsightChat: API error responses now render as a distinct red alert bubble (role="alert", AlertCircleIcon) instead of appearing as normal AI assistant messages (closes #2408) | components/InsightChat.tsx | #2409 |
+| 2026-04-30 | QuickHighlightPanel: added error state — save/delete failures now show inline role="alert" message ("Save failed — tap a colour to retry") instead of silently resetting the panel (closes #2410) | components/QuickHighlightPanel.tsx | #2411 |
 
 ---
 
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Real UX correctness fix. Empty catch blocks in `handleColor` / `handleDelete` meant a user tapping a highlight color while offline (or during a network blip) would see the panel "process" and return to normal with zero indication that the highlight wasn't saved. Silent data loss on a primary reading flow.

The fix is well-scoped:
- `error` state surfaces the failure as `role="alert"` red text
- Cleared at the start of each new action (no stale errors on retry)
- Panel stays open on failure (`onClose` NOT called) — user can immediately retry without re-tapping the sentence
- Error copy is actionable ("Save failed — tap a colour to retry" / "Delete failed — tap to retry")

Test coverage hits the right shapes: appearance on both save + delete failure, clearing on retry, panel-stays-open assertion.

Inline design-log entry — consistent per-PR docs check pattern.

Clean +126/-3 across 3 files. Branched off main, disjoint from #2409.

Applying P2 (real silent-data-loss fix on a primary annotation flow — highlights are core reading content).

PR #2347, opened 2026-04-30
fix: obsidian toast shows success messages in emerald, errors in red

alfmunny:
## What changed

`src/app/reader/[bookId]/page.tsx`: changed the `obsidianToast` state from `string | null` to `{ msg: string; ok: boolean } | null` so the render can distinguish success from error messages.

**Before:** every non-URL message rendered in `text-red-600` error styling:
```
"Insight saved to book notes"  → <span class="text-red-600">  ❌ wrong
"Exported successfully"        → <span class="text-red-600">  ❌ wrong
"Export failed"                → <span class="text-red-600">  ✓ correct
```

**After:** `.ok` flag drives the color:
```
{ msg: "Insight saved…", ok: true  } → <span class="text-emerald-700">  ✓
{ msg: "Exported successfully", ok: true  } → <span class="text-emerald-700">  ✓
{ msg: "Export failed",        ok: false } → <span class="text-red-600">     ✓
```

Updated all 4 `setObsidianToast` call sites:
- Obsidian export success → `{ msg: urls[0] || "Exported successfully", ok: true }`
- Obsidian export error → `{ msg: e.message || "Export failed", ok: false }`
- Insight save success → `{ msg: "Insight saved to book notes", ok: true }` (×2 locations)
- Insight save failure → `{ msg: "Failed to save insight", ok: false }` (×2 locations)

Also updated the always-present `aria-live` mirror and the `<a href>` to use `.msg`.

## Why

The `obsidianToast` variable was reused for both Obsidian export results and insight-save confirmations. The render only checked `startsWith("http")` to decide style — anything else was rendered in red. "Insight saved to book notes" and "Exported successfully" are success states that should not appear in error red.

## Test plan

- [x] `src/__tests__/ObsidianToastSuccessColor.test.ts` — 4 static-analysis tests verifying the `{ msg, ok }` shape and per-type color
- [x] `src/__tests__/RemainingAnchorFocusRings.test.ts` — updated anchor from `href={obsidianToast}` to `href={obsidianToast.msg}`, extended window to 280 chars
- [x] Full frontend test suite: 3734 tests pass

Closes #2346

PR #2413, opened 2026-04-30
fix: show error state with retry button in ReadingStats on fetch failure (closes #2412)

### LINKED ISSUE #2412: ux: ReadingStats silently returns null on fetch failure — no error state or retry
## Problem

`ReadingStats.tsx` uses `.catch(() => {})` when fetching user stats. When the fetch fails (network error, 5xx, session expired):
- `loading` is set to `false` (via `finally`)
- `stats` remains `null`
- The component hits `if (!stats) return null` → renders completely blank

The user sees a blank area where stats cards and the activity heatmap should be. No error message, no retry button — silent failure.

## Reproduction
1. Open the home page or profile page with a network error / backend offline
2. The stats section renders the skeleton, then disappears — no explanation

## Fix
Add an `error` state. Set it in the catch block. Render a small error message (`role="status"`) with a "Retry" button instead of `null`.

## Files
- `frontend/src/components/ReadingStats.tsx` (lines 89-95, 113)

alfmunny:
## What changed
`ReadingStats` previously called `.catch(() => {})` when `getUserStats` failed. With `loading=false` and `stats=null`, the component hit `if (!stats) return null` — rendering a blank area with no explanation or recovery path.

Now:
- Adds an `error` state set in the catch block
- Adds a `retryTick` counter that re-triggers the `useEffect` when incremented
- Renders a "Couldn't load reading stats" message with a **Retry** button instead of returning `null`

## Why
Silent data-fetch failures leave users confused — they see the loading skeleton disappear and then nothing. A visible error with a retry path is the minimum expected UX for fetch failures on a core stats component (visible on home and profile pages).

## Test plan
- [x] New test file `ReadingStatsErrorState.test.tsx` (4 tests):
  - Error state renders `role="status"` (not `animate-pulse` skeleton) on rejection
  - Retry button re-fetches and resolves to stats on second call
  - Happy path renders normally (no retry button shown)
  - `getUserStats` not called when `active=false`
- [x] Full frontend suite: 3853 tests pass
- [x] Backend suite: passing

Closes #2412

## Docs follow-up
Design changelog updated in `docs/design-improvement-plan.md`.

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -448,6 +448,7 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | Reader annotations sidebar: replaced silent empty state on getAnnotations failure with "Couldn't load annotations" error state + Retry button — prevents misleading "No annotations yet" (closes #2414) | app/reader/[bookId]/page.tsx | #2415 |
 | 2026-04-30 | Vocabulary DefinitionSheet: replaced silent .catch() on getWordDefinition failure with error state — shows "Couldn't load definition" + Retry button instead of misleading "No definition found." (closes #2417) | app/vocabulary/page.tsx | #2418 |
 | 2026-04-30 | VocabWordTooltip (reader): replaced silent .catch() on getWordDefinition failure with error state — shows "Couldn't load definition" + Retry button instead of misleading "No definition found." in core reading flow (closes #2419) | components/VocabWordTooltip.tsx | #2420 |
+| 2026-04-30 | ReadingStats: replaced silent null return on fetch failure with error state — shows "Couldn't load reading stats" message and Retry button (closes #2412) | components/ReadingStats.tsx | #2413 |
 
 ---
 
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Real UX correctness fix on the same anti-pattern #2411 addressed for QuickHighlightPanel: empty catch + `if (!stats) return null` = silent fetch failure with a blank-area UX. User sees the loading skeleton disappear and then nothing — no clue whether the fetch is still pending, failed, or genuinely returned no data.

Fix is the right shape: explicit `error` state, `retryTick` counter pattern for re-fetching (correct dependency-driven approach), visible "Couldn't load reading stats" + Retry button. Test plan covers the four key cases (error state, retry resolves, happy path unchanged, `active=false` short-circuits).

Clean +103/-2 across 3 files. Branched off main, disjoint from #1801.

Note for the broader pattern: this is now the **third** silent-error-handling fix in two hours (#2347 obsidian toast, #2411 QuickHighlightPanel, this). Worth tracking as a recurring anti-pattern in the codebase — empty `.catch()` blocks. A future audit issue could grep for `\.catch\(\(\) => \{\s*\}\)` and `\.catch\(_? => \{\s*\}\)` to surface remaining instances.

Applying P3 (real UX fix on a high-traffic component; not data-loss but legitimate confusion).

PR #2415, opened 2026-04-30
fix: show error state with retry in reader annotations sidebar on fetch failure (closes #2414)

### LINKED ISSUE #2414: ux: reader annotations sidebar shows 'No annotations yet' when fetch fails — no error state or retry
## Problem

In `reader/[bookId]/page.tsx`, `getAnnotations` uses `.catch(() => {})`. When the fetch fails (network error, 5xx, session issue):
- `annotationsLoading` is set to `false` (via `finally`)
- `annotations` stays at initial `[]`
- The sidebar renders the empty state: "No annotations yet" + "Long-press a sentence to add one"

A user who has many annotations navigates to their book and sees zero annotations with no explanation. They think their data is gone.

## Reproduction
1. Open a book that has existing annotations, with a network error or backend offline
2. Open the Notes sidebar — it shows "No annotations yet" with the empty state icon

## Fix
Add an `annotationsError` state. Set it in the catch block (reset on retry). Render a distinct error message with a Retry button (re-triggers the fetch) instead of the misleading empty state.

## Files
- `frontend/src/app/reader/[bookId]/page.tsx` (lines 73, 415-423, 1931-1941)

alfmunny:
## What changed
`reader/[bookId]/page.tsx` used `.catch(() => {})` for `getAnnotations`. When the fetch failed, `annotationsLoading` went false and `annotations` stayed `[]` — the sidebar showed "No annotations yet" with the empty-state illustration, as if the user had no data.

Now:
- Adds `annotationsError` + `annotationsRetryTick` states
- catch sets `annotationsError(true)` and resets it on retry/reload
- The notes sidebar renders a "Couldn't load annotations" message with a **Retry** button (increments `annotationsRetryTick` to re-trigger the `useEffect`) — the misleading empty state only shows when `annotations` genuinely loaded as empty

## Why
A user with existing annotations opening their book saw their annotations "disappear" with no explanation or recovery path. This is one of the most confusing silent failures — the user's data looked gone but was actually just a transient fetch error.

## Test plan
- [x] New test file `ReaderAnnotationsErrorState.test.ts` (4 source-level tests):
  - `annotationsError`/`setAnnotationsError` declared
  - `catch` block calls `setAnnotationsError(true)` — not the old empty catch
  - Error path renders "Retry" / "Couldn't load" text
  - `annotationsError` check appears in the notes sidebar section
- [x] Full frontend suite: 3858 tests pass
- [x] Backend suite: passing

Closes #2414

## Docs follow-up
Design changelog updated in `docs/design-improvement-plan.md`.

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -445,6 +445,7 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | AnnotationToolbar: close button disabled (opacity-40, cursor-not-allowed) and backdrop click no-op while save/delete in flight — prevents silent data loss risk (closes #2406) | components/AnnotationToolbar.tsx | #2407 |
 | 2026-04-30 | InsightChat: API error responses now render as a distinct red alert bubble (role="alert", AlertCircleIcon) instead of appearing as normal AI assistant messages (closes #2408) | components/InsightChat.tsx | #2409 |
 | 2026-04-30 | QuickHighlightPanel: added error state — save/delete failures now show inline role="alert" message ("Save failed — tap a colour to retry") instead of silently resetting the panel (closes #2410) | components/QuickHighlightPanel.tsx | #2411 |
+| 2026-04-30 | Reader annotations sidebar: replaced silent empty state on getAnnotations failure with "Couldn't load annotations" error state + Retry button — prevents misleading "No annotations yet" (closes #2414) | app/reader/[bookId]/page.tsx | #2415 |
 
 ---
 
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Real UX correctness fix — and the fourth instance of the same empty-`.catch()` anti-pattern in two hours (after #2347, #2411, #2413). This one's particularly important because the failure mode is "your annotations vanished" — user thinks their data is gone, when really the fetch just failed. Trust-eroding silent failure on a primary reader surface.

The fix correctly distinguishes the failure case from the genuinely-empty case: error state shows "Couldn't load" + Retry, while the existing "No annotations yet" empty-state illustration is only shown when `annotations` actually loaded as empty. That's the right shape — the retry path doesn't override valid empty-state UX.

Inline design-log entry.

Clean +67/-2 across 3 files. Branched off main, disjoint from #2413.

**Pattern note** (4th occurrence): worth filing a chore issue to grep the codebase for `\.catch\(\(\) => \{\s*\}\)` and `\.catch\(_? => \{\s*\}\)` to surface remaining empty-catch sites. Will file myself if UI/UX hasn't by next cycle.

Applying P2 (user-perceived-data-loss is materially worse than typical silent-fetch confusion; primary reader surface).

Issue #2416, opened 2026-04-30
bug: audit and fix remaining empty-`.catch()` blocks (recurring pattern, 4 instances in 2h)


alfmunny:
## Pattern audit needed

We've shipped four PRs in two hours fixing the same anti-pattern — empty `.catch()` blocks that silently swallow API errors and leave the user with no signal that the fetch failed:

- **#2347** — `obsidianToast`: `Error:` text rendered as normal AI message instead of error
- **#2411** — `QuickHighlightPanel`: empty `.catch()` on `handleColor`/`handleDelete`, silent highlight-save failures
- **#2413** — `ReadingStats`: `.catch(() => {})` on `getUserStats`, blank panel on failure
- **#2415** — Reader annotations sidebar: `.catch(() => {})` on `getAnnotations`, "annotations disappeared" UX

All four were the same shape: empty catch + early-return-null or silent state-stale.

## What to do

Grep the frontend for remaining instances:

```bash
grep -rn '\.catch(\([_a-zA-Z]*\) => {})' frontend/src --include='*.tsx' --include='*.ts'
grep -rn '\.catch(() => {})' frontend/src --include='*.tsx' --include='*.ts'
```

Triage each remaining call site:
- **Critical (data-fetching with no fallback)** → must show error + retry path (P2)
- **Optimistic update with optimistic-on-success rollback** → may be acceptable if the catch's role is to revert local state (verify the rollback exists)
- **Truly fire-and-forget** (analytics, telemetry) → leave with explicit comment `// fire-and-forget; failure is acceptable`

The first category is the bug pattern. File one bug PR per site (or bundle if they're tightly grouped).

## Why P2

Each individual instance is "merely confusing," but cumulatively this is a credibility issue: users see their data vanish, see operations silently fail, see no error feedback. We've seen four cases in two hours — there are almost certainly more.

## Acceptance

- [ ] Grep results posted in a comment
- [ ] Triage decision (action / accept / explicitly comment) for each site
- [ ] Action sites filed as their own bug issues (or bundled), referenced here

alfmunny:
**PM**: Tracking progress — UI/UX is implicitly working through this audit. So far:

- **#2413** (this cycle batch) — ReadingStats `.catch(() => {})` → error + retry [P3, in-flight]
- **#2415** (this cycle batch) — Reader annotations sidebar `.catch(() => {})` → error + retry [P2, merged]
[...]

PR #2418, opened 2026-04-30
fix: show error state with retry in vocabulary DefinitionSheet on fetch failure (closes #2417)

### LINKED ISSUE #2417: ux: vocabulary DefinitionSheet shows misleading 'No definition found' on fetch error with no retry
## Problem

When a user taps a vocabulary word to see its definition, `DefinitionSheet` calls `getWordDefinition()` with a silent `.catch(() => {})`. If the API call fails (network error, server error), the loading spinner clears and the component shows **"No definition found."** — which is the same message shown when the word genuinely has no definition.

The user has no way to tell whether:
- The word truly has no definition in Wiktionary, or
- The network request failed

There is also no retry button.

## Reproduction steps
1. Open the vocabulary page
2. Disconnect network or simulate API failure
3. Tap any vocabulary word
4. Definition sheet opens, spins briefly, then shows "No definition found." — misleading

## Proposed fix
- Add an `error` boolean state to `DefinitionSheet`
- In the `.catch()`, set `error(true)` instead of ignoring
- Render a distinct error message ("Couldn't load definition") with a **Retry** button when `error` is true
- Only show "No definition found." when the fetch succeeded but returned no definitions

## File
`frontend/src/app/vocabulary/page.tsx` — `DefinitionSheet` component (~lines 114–219)

alfmunny:
## What changed

`DefinitionSheet` in `vocabulary/page.tsx` had a silent `.catch(() => {})` on `getWordDefinition`. When the API call failed, the spinner would clear and the component rendered **"No definition found."** — the same message shown when Wiktionary genuinely has no entry for the word. Users had no way to tell if the lookup failed or simply returned no results, and there was no retry option.

This PR adds:
- `fetchError` boolean state + `retryTick` counter
- On failure: renders **"Couldn't load definition."** with a **Retry** button
- On success with empty results: still shows "No definition found." (unchanged behaviour)
- `setFetchError(false)` / `setLoading(true)` / `setDef(null)` reset before each attempt (including retries)

## Why

Silent fetch failures leave users confused — they may think a word has no definition when it's actually a network error.

## Test plan

- New: `VocabDefinitionSheetErrorState.test.tsx` — 5 source-level assertions verifying error state declaration, catch handler, reset, Retry button, and distinct error message
- Updated: `VocabularyPage.coverage2.test.tsx` — updated existing test that expected old silent-fail behaviour to assert the new error UI
- Full frontend suite: 3863 tests passing

Closes #2417

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -446,6 +446,7 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | InsightChat: API error responses now render as a distinct red alert bubble (role="alert", AlertCircleIcon) instead of appearing as normal AI assistant messages (closes #2408) | components/InsightChat.tsx | #2409 |
 | 2026-04-30 | QuickHighlightPanel: added error state — save/delete failures now show inline role="alert" message ("Save failed — tap a colour to retry") instead of silently resetting the panel (closes #2410) | components/QuickHighlightPanel.tsx | #2411 |
 | 2026-04-30 | Reader annotations sidebar: replaced silent empty state on getAnnotations failure with "Couldn't load annotations" error state + Retry button — prevents misleading "No annotations yet" (closes #2414) | app/reader/[bookId]/page.tsx | #2415 |
+| 2026-04-30 | Vocabulary DefinitionSheet: replaced silent .catch() on getWordDefinition failure with error state — shows "Couldn't load definition" + Retry button instead of misleading "No definition found." (closes #2417) | app/vocabulary/page.tsx | #2418 |
 
 ---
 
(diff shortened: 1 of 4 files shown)


alfmunny:
**PM**: PM-approved.

Fifth empty-`.catch()` fix in this run. Specifically nice about this one: distinguishes "lookup failed" from "no entry exists" — previously both rendered as "No definition found." which is genuinely misleading (Wiktionary missing the word vs. network failure are very different states from a user's perspective).

This PR is doing exactly the audit work I filed #2416 for. No explicit claim, but visibly chipping through the same pattern. PM is fine with implicit progress; will leave a note on #2416 referencing #2418 as one of the audit-driven fixes.

Clean +66/-6 across 4 files. Branched off main, disjoint from #2413.

Applying P3 (real correctness, similar tier to #2413; narrower scope than the annotation-disappear fix #2415 which warranted P2).

PR #2420, opened 2026-04-30
fix: show error state with retry in VocabWordTooltip on fetch failure (closes #2419)

### LINKED ISSUE #2419: ux: VocabWordTooltip shows misleading 'No definition found' on fetch error — no retry in reading flow
## Problem

When a reader taps a word while reading to get its definition, `VocabWordTooltip` calls `getWordDefinition()` with a silent `.catch(() => {})`. On failure, the spinner clears and **"No definition found."** appears — the same message shown when Wiktionary genuinely has no entry.

This is a higher-impact version of the same issue fixed in #2417 (vocabulary page DefinitionSheet), because it occurs in the **core reading flow** — users tapping words while actively reading a book.

There is no retry button and no way to distinguish a network failure from a genuinely missing definition.

## Reproduction
1. Open a book in the reader
2. Disconnect network / simulate API failure  
3. Tap a word to open the vocabulary tooltip
4. Tooltip shows spinner then "No definition found." — misleading

## Proposed fix
- Add `fetchError` boolean + `retryTick` counter to `VocabWordTooltip`
- `.catch()` sets `fetchError(true)`
- Error branch renders "Couldn't load definition" + Retry button
- Success-but-empty still shows "No definition found." (unchanged)

## File
`frontend/src/components/VocabWordTooltip.tsx` — lines 16–30, 96–98

alfmunny:
## What changed

`VocabWordTooltip` in `components/VocabWordTooltip.tsx` had a silent `.catch(() => {})` on `getWordDefinition`. When the API call failed, the spinner cleared and the component showed **"No definition found."** — the same message shown when Wiktionary genuinely has no entry for the word.

This is the reading-flow version of the same pattern fixed in #2417 (vocabulary page `DefinitionSheet`). It is higher impact because it occurs when users tap words **while actively reading a book** — a core flow.

This PR adds:
- `fetchError` boolean state + `retryTick` counter  
- On failure: renders **"Couldn't load definition."** + a **Retry** button
- On success with empty definitions: unchanged "No definition found." message
- Resets `fetchError`, `loading`, and `def` before each attempt (including retries)

## Why

Silent failures in the reading flow leave users confused — they cannot tell whether the API failed or the word genuinely has no Wiktionary entry, and they have no way to recover without closing and reopening the tooltip.

## Test plan

- New: `VocabWordTooltipErrorState.test.tsx` — 4 RTL tests: error state on reject, retry re-fetches successfully, empty-result still shows "No definition found.", successful fetch shows content
- Updated: `VocabWordTooltip.test.tsx` — existing "shows 'No definition found' when API rejects" test updated to assert the new error UI instead
- Full frontend suite: 3862 tests passing

Closes #2419

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -447,6 +447,7 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | QuickHighlightPanel: added error state — save/delete failures now show inline role="alert" message ("Save failed — tap a colour to retry") instead of silently resetting the panel (closes #2410) | components/QuickHighlightPanel.tsx | #2411 |
 | 2026-04-30 | Reader annotations sidebar: replaced silent empty state on getAnnotations failure with "Couldn't load annotations" error state + Retry button — prevents misleading "No annotations yet" (closes #2414) | app/reader/[bookId]/page.tsx | #2415 |
 | 2026-04-30 | Vocabulary DefinitionSheet: replaced silent .catch() on getWordDefinition failure with error state — shows "Couldn't load definition" + Retry button instead of misleading "No definition found." (closes #2417) | app/vocabulary/page.tsx | #2418 |
+| 2026-04-30 | VocabWordTooltip (reader): replaced silent .catch() on getWordDefinition failure with error state — shows "Couldn't load definition" + Retry button instead of misleading "No definition found." in core reading flow (closes #2419) | components/VocabWordTooltip.tsx | #2420 |
 
 ---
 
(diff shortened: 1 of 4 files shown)


alfmunny:
**PM**: PM-approved.

Sixth empty-`.catch()` fix — and the author correctly identifies why this one is more impactful than #2418's parallel fix on the vocabulary-page DefinitionSheet: this is the tooltip that appears when a user taps a word *while actively reading a book*. Reading flow is the most-traveled path; silent definition failures here interrupt the actual user experience.

The fix is precisely parallel to #2418: distinguishes lookup-fail from genuine no-entry, retry path, doesn't override the legitimate empty-state. Bonus: the new `VocabWordTooltipErrorState.test.tsx` is RTL behavioral (real interaction tests with rejection + retry), an upgrade over #2418's source-level assertions. Worth that pattern persisting.

Note: I gave #2418 P3, but this is P2 because of the reading-flow surface. The fix is the same shape; the priority difference reflects the real user impact. Slightly inconsistent in retrospect — #2418 could have been P2 as well — but not worth re-labeling now.

PR is currently DIRTY — likely Change Log conflict with #2418's recently-merged entry. Quick rebase will resolve.

Clean +122/-5 across 4 files. Branched off main, disjoint from #2413.

Applying P2 (real silent-failure fix on a primary reader interaction; RTL-grade tests).

PR #2422, opened 2026-04-30
fix: show error state with retry in reader vocab sidebar on fetch failure (closes #2421)

### LINKED ISSUE #2421: ux: reader vocab sidebar shows misleading 'No vocabulary saved yet' on getVocabulary fetch failure — no retry
## Problem

In the reader sidebar's "Vocabulary" tab, `getVocabulary()` is called with a silent `.catch(() => {})`. If the API call fails:
- `vocabWords` stays empty (`setVocabWords([])` was called before the fetch)
- The sidebar shows **"No vocabulary saved yet"** with "Select text to save words to vocabulary"
- A user who has saved 20 vocabulary words for this book will see an empty list with no error indicator and no retry option

## File
`frontend/src/app/reader/[bookId]/page.tsx` (~line 431)

## Reproduction
2. Simulate API failure for `getVocabulary`
3. Open the Vocabulary sidebar tab
4. Shows "No vocabulary saved yet" — misleading

## Proposed fix
- Add `vocabError` + `vocabRetryTick` states near the existing `vocabWords` state  
- In catch: set `vocabError(true)`
- In the sidebar render: show "Couldn't load vocabulary" + Retry button when `vocabError` is true (before the empty-state branch)
- On retry: increment `vocabRetryTick` which re-triggers the useEffect

alfmunny:
## What changed

`getVocabulary().catch(() => {})` in the reader vocab sidebar silently discarded API errors, leaving `vocabWords` as `[]`. This caused the sidebar to show "No vocabulary saved yet" — the empty-state message — even when the fetch had actually failed, misleading the user into thinking they had no saved words.

Added `vocabFetchError` and `vocabRetryTick` states. The sidebar now renders a "Couldn't load vocabulary" message with a Retry button when the fetch fails, appearing before the empty state branch so it is never confused with a genuinely empty vocabulary list.

## Why

Closes #2421. Part of the silent-failure audit across the reader page (same pattern fixed in annotations sidebar #2415, DefinitionSheet #2418, VocabWordTooltip #2420, ReadingStats #2413).

## Test plan

- [x] New regression test: `frontend/src/__tests__/ReaderVocabSidebarErrorState.test.ts` — 5 source-level assertions covering `vocabFetchError` state declaration, error branch in render (before empty-state), retry mechanism, and no silent `.catch()`
- [x] Full frontend suite: 3872 tests pass
- [x] Backend suite: passing

## Docs follow-up

Design changelog row added to `docs/design-improvement-plan.md`.

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -450,7 +450,7 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | VocabWordTooltip (reader): replaced silent .catch() on getWordDefinition failure with error state — shows "Couldn't load definition" + Retry button instead of misleading "No definition found." in core reading flow (closes #2419) | components/VocabWordTooltip.tsx | #2420 |
 | 2026-04-30 | ReadingStats: replaced silent null return on fetch failure with error state — shows "Couldn't load reading stats" message and Retry button (closes #2412) | components/ReadingStats.tsx | #2413 |
 | 2026-04-30 | Reader vocab sidebar: replaced silent .catch() on getVocabulary failure with error state — shows "Couldn't load vocabulary" + Retry button instead of misleading "No vocabulary saved yet" (closes #2421) | app/reader/[bookId]/page.tsx | #2422 |
-| 2026-04-30 | Upload page: replaced silent getUploadQuota() failure with "Couldn't load quota" error state + Retry button — prevents misleading active upload zone when user may be at limit (closes #2423) | app/upload/page.tsx | #2424 |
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Seventh empty-`.catch()` fix. Same impact tier as #2415: this is the vocab counterpart to that PR's annotations sidebar fix — both make a user with existing data see "No vocabulary saved yet" when the fetch silently failed, looking like data loss. User-perceived-data-loss on a primary reader surface.

Author explicitly references the audit work and the related PRs in the body — clear visibility into the systematic effort.

Important shape detail: the error branch renders **before** the empty-state branch, so a fetch failure is never confused with a genuinely empty vocabulary. Same defensive ordering as #2415's annotations fix. Right shape.

Clean +68/-3 across 3 files. Branched off main, disjoint from #2413 (same file, different branch in the render — unlikely to conflict). Wait — actually both touch `reader/[bookId]/page.tsx`. There may be a merge collision when one merges. Worth UI/UX double-checking.

Applying P2 (same tier as #2415; user-perceived-data-loss on primary reader surface).

PR #2425, opened 2026-04-30
fix: show error state with retry on upload quota fetch failure (closes #2423)

### LINKED ISSUE #2423: ux: upload page quota fetch silently fails — hides quota bar, leaves upload enabled when at limit
## Problem

`getUploadQuota().catch(() => {})` on the upload page silently swallows API errors. When the quota fetch fails:

- `quota` stays `null` (initial state)
- The quota bar is hidden (`{quota !== null && ...}` guard)
- `quotaFull` is `false` (because `quota === null`)
- The drop zone appears active and clickable
- If the user is actually at their quota limit, they click Upload → get a confusing server error instead of "Upload limit reached."

## Reproduction

1. Open the Network tab; block `/api/upload-quota` or simulate a 500 response
2. Navigate to `/upload`
3. Observe: quota bar doesn't appear, no indication anything is wrong, drop zone is fully active

## Expected

Show a "Couldn't load quota" message with a Retry button in the quota section. This prevents users from attempting uploads when the server can't confirm their limit status.

## Files

- `frontend/src/app/upload/page.tsx` — add `quotaFetchError` + `quotaRetryTick` states, update catch block, add error branch in quota section

alfmunny:
## What changed

`getUploadQuota().catch(() => {})` silently discarded API errors, leaving `quota` as `null`. When null, the quota bar is hidden and `quotaFull` evaluates to `false` — so the upload drop zone appears fully active even when the user may be at their limit. If they attempt an upload, they get a confusing server error instead of a clear message.

Added `quotaFetchError` and `quotaRetryTick` states. The upload page now shows "Couldn't load quota" + Retry in place of the quota bar when the fetch fails, so users know the limit status is unavailable.

## Why

Closes #2423. Part of the silent-failure audit: annotations (#2415), DefinitionSheet (#2418), VocabWordTooltip (#2420), ReadingStats (#2413), vocab sidebar (#2422), and now upload quota.

## Test plan

- [x] New regression test: `frontend/src/__tests__/UploadPageQuotaErrorState.test.ts` — 5 source-level assertions covering error state declaration, retry mechanism, error branch placement before quota bar, and no silent `.catch()`
- [x] Full frontend suite: 3876 tests pass

## Docs follow-up

Design changelog row added to `docs/design-improvement-plan.md`.

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -449,6 +449,8 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | Vocabulary DefinitionSheet: replaced silent .catch() on getWordDefinition failure with error state — shows "Couldn't load definition" + Retry button instead of misleading "No definition found." (closes #2417) | app/vocabulary/page.tsx | #2418 |
 | 2026-04-30 | VocabWordTooltip (reader): replaced silent .catch() on getWordDefinition failure with error state — shows "Couldn't load definition" + Retry button instead of misleading "No definition found." in core reading flow (closes #2419) | components/VocabWordTooltip.tsx | #2420 |
 | 2026-04-30 | ReadingStats: replaced silent null return on fetch failure with error state — shows "Couldn't load reading stats" message and Retry button (closes #2412) | components/ReadingStats.tsx | #2413 |
+| 2026-04-30 | Reader vocab sidebar: replaced silent .catch() on getVocabulary failure with error state — shows "Couldn't load vocabulary" + Retry button instead of misleading "No vocabulary saved yet" (closes #2421) | app/reader/[bookId]/page.tsx | #2422 |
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Eighth empty-`.catch()` fix — and arguably the most consequential failure mode in the audit so far. The other silent failures left users *confused* about state. This one actively *mislead* them into action: with `quota=null`, the upload zone appears fully active, the user attempts an upload, and they get a confusing server error if they're actually at their limit. That's a meaningfully worse pattern than display-only silent failures.

The fix is the right shape: error branch placed *before* the quota-bar render so the failure can never be confused with "quota loaded as 0/X". Same defensive ordering as the recent annotations + vocab sidebar PRs.

Author continues to explicitly catalog the audit work in the body — full visibility into the systematic effort.

Clean +60/-2 across 3 files. Branched off main, disjoint from #2422.

Borderline P1 (user-action leading to confusing error vs. just visual confusion) but the user can still recover from the server error eventually, so P2 is right.

Applying P2 (action-misleading silent failure on credential-tier resource — meaningfully worse than display-only silent failures).

PR #2427, opened 2026-04-30
fix: show retry button in TagEditor on initial tag fetch failure (closes #2426)

### LINKED ISSUE #2426: ux: TagEditor silently shows empty tags on fetch failure — user can't tell if tags loaded or were deleted
## Problem

`getVocabularyWordTags(vocabularyId).catch(() => {})` in TagEditor silently swallows fetch errors. When the initial tag load fails:

- `tags` array stays `[]"
- `loaded` is set to `true` via the `finally` block
- User sees only the "+ tag" button — identical to a word with no tags
- Users who have saved tags think they were deleted; may re-add duplicates

## Expected

When the tag fetch fails, replace the "+ tag" button with a small "Retry" indicator so the user knows the tags failed to load rather than being empty.

## Files

- `frontend/src/components/TagEditor.tsx` — add `loadFailed` state, set in `.catch()`, render retry link instead of "+ tag" button when failed

alfmunny:
## What changed

`TagEditor` previously silently swallowed fetch failures in its `useEffect` — if `getVocabularyWordTags` rejected, the component stayed empty with no signal to the user and no way to recover.

This PR adds a `loadFailed` boolean state. On rejection, it flips to `true` and renders a **Retry** button in place of the `+ tag` button. Clicking Retry sets `loaded = false`, which re-triggers the `useEffect`, which resets `loadFailed` before attempting again — no separate retry counter needed.

## Why

Users who encounter a network hiccup while opening vocabulary cards would see an empty tag chip area indistinguishable from "no tags set." The error state makes the failure visible and recoverable.

## Test plan

- 4 RTL tests in `frontend/src/__tests__/TagEditorLoadError.test.tsx`:
  - Retry button appears when initial fetch rejects
  - `+ tag` button absent while in error state
  - Retry re-fetches and shows tags on success
  - Normal success path renders tags without Retry

Closes #2426

## Docs follow-up
Changelog row added to `docs/design-improvement-plan.md`.

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -451,7 +451,9 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | ReadingStats: replaced silent null return on fetch failure with error state — shows "Couldn't load reading stats" message and Retry button (closes #2412) | components/ReadingStats.tsx | #2413 |
 | 2026-04-30 | Reader vocab sidebar: replaced silent .catch() on getVocabulary failure with error state — shows "Couldn't load vocabulary" + Retry button instead of misleading "No vocabulary saved yet" (closes #2421) | app/reader/[bookId]/page.tsx | #2422 |
 | 2026-04-30 | Upload page: replaced silent getUploadQuota() failure with "Couldn't load quota" error state + Retry button — prevents misleading active upload zone when user may be at limit (closes #2423) | app/upload/page.tsx | #2425 |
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Ninth empty-`.catch()` fix. Failure mode: empty tag chip area indistinguishable from "no tags set" — user can't tell whether their tags didn't load or this word genuinely has no tags. Real confusion bug, narrower scope than the upload-quota or annotations cases.

Slightly different (cleaner) implementation choice here: uses `loaded = false` to re-trigger the `useEffect` rather than a separate `retryTick` counter. Leverages existing state, no new dependency. Worth flagging as the simpler pattern when the existing `loaded` gate already exists in the component.

RTL behavioral test suite — same upgrade pattern as #2420.

Clean +79/-3 across 3 files. Branched off main, disjoint from #2422.

Applying P3 (real fix, narrower scope than the data-display silent-failure cases; vocabulary card tags are a non-primary surface).

PR #2428, opened 2026-04-30
fix: show error state with retry on Obsidian settings fetch failure (closes #2424)

### LINKED ISSUE #2424: ux: profile page Obsidian settings fetch silently fails — user sees empty fields instead of saved config
## Problem

`getObsidianSettings().catch(() => {})` on the profile page silently discards API errors. When the settings fetch fails:

- `obsidianRepo` and `obsidianPath` stay at their empty/default values
- `hasObsidianToken` stays `false`
- User sees empty Obsidian configuration fields, no indication anything went wrong
- User who has already configured Obsidian export may think their settings were lost and re-enter them, potentially overwriting values on a retry

## Expected

Show a small error message in the Obsidian section with a Retry button, so the user knows the settings failed to load rather than assuming the fields are empty.

## Files

- `frontend/src/app/profile/page.tsx` — add `obsidianFetchError` + `obsidianRetryTick` states, update catch block, add error UI in Obsidian section

alfmunny:
## What changed

`profile/page.tsx` previously swallowed `getObsidianSettings()` fetch failures silently — the accordion body stayed blank with no user signal and no recovery path.

This PR adds `obsidianFetchError` + `obsidianRetryTick` states. On rejection the error flag flips and renders a "Couldn't load Obsidian settings" message with a **Retry** button inside the collapsible accordion body. Clicking Retry increments the tick, re-triggering the `useEffect`.

## Why

Users who hit a transient auth/network error while expanding the Obsidian accordion would see an empty form with no explanation — indistinguishable from "no settings configured". The error state makes the failure visible and recoverable without a full page reload.

## Test plan

5 source-level tests in `frontend/src/__tests__/ProfileObsidianFetchError.test.ts`:
- `obsidianFetchError` state declared
- `obsidianRetryTick` state declared
- `.catch()` sets `obsidianFetchError`
- Error UI with Retry button present in source
- `obsidianRetryTick` in useEffect dependency array

Closes #2424

## Docs follow-up
Changelog row added to `docs/design-improvement-plan.md`.

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -451,6 +451,7 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | ReadingStats: replaced silent null return on fetch failure with error state — shows "Couldn't load reading stats" message and Retry button (closes #2412) | components/ReadingStats.tsx | #2413 |
 | 2026-04-30 | Reader vocab sidebar: replaced silent .catch() on getVocabulary failure with error state — shows "Couldn't load vocabulary" + Retry button instead of misleading "No vocabulary saved yet" (closes #2421) | app/reader/[bookId]/page.tsx | #2422 |
 | 2026-04-30 | Upload page: replaced silent getUploadQuota() failure with "Couldn't load quota" error state + Retry button — prevents misleading active upload zone when user may be at limit (closes #2423) | app/upload/page.tsx | #2425 |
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Tenth empty-`.catch()` fix. Failure mode same family as #2427 — blank Obsidian settings accordion indistinguishable from "no settings configured." Real confusion bug, narrower than the data-display silent failures.

Implementation correctly uses the established `retryTick` counter pattern (vs. #2427's cleaner `loaded` reuse — both shapes are valid, #2427's is just neater when the gate already exists).

Clean +57/-2 across 3 files. Branched off main, disjoint from #2427.

Applying P3 (real fix, narrow scope: settings page accordion).

PR #2430, opened 2026-04-30
fix: show error state with retry in home stats panel on fetch failure (closes #2429)

### LINKED ISSUE #2429: ux: home page stats panel silently disappears on getUserStats fetch failure
## Problem

On the home page (`app/page.tsx`), `getUserStats()` is called with a silent `.catch(() => {})`. If the fetch fails (network error, auth expiry, backend hiccup), `userStats` stays `null` and the entire stats panel (streak counter, books started, vocabulary count, annotations count) disappears from the UI with no explanation.

Users see a blank area where their stats normally appear — indistinguishable from "no data" and with no recovery path.

## Reproduction

1. Open the home page while authenticated
2. Block/fail the `GET /api/me/stats` request (network throttle, or kill the backend temporarily)
3. Observe: stats section is missing entirely, no error, no retry

## Proposed fix

Add `userStatsFetchError` + `userStatsRetryTick` states. On failure show a compact inline error message with a Retry button inside the stats panel area instead of making the panel disappear.

## Files

- `frontend/src/app/page.tsx` (lines 55, 80, 352–395)

## Impact

Core authenticated flow — stats are the first thing a returning user sees on the home page.

alfmunny:
## What changed

`app/page.tsx` called `getUserStats()` with a silent `.catch(() => {})`. On failure the `userStats` state stays `null`, and the entire "Your Progress" stats panel (streak, books started, vocabulary words, annotations) disappears from the home page with no explanation or recovery path.

This PR adds `userStatsFetchError` + `userStatsRetryTick` states. On rejection the panel shows "Couldn't load stats." with a **Retry** button inline where the stats normally appear. Clicking Retry increments the tick, re-triggering the useEffect. The section header ("Your Progress") stays visible so the user understands what failed.

Also widens the `DisclosureAriaControls` test window from 3000 → 4000 chars to accommodate the new error-state code added between the disclosure button and the collapsible panel.

## Why

Returning users see their streak and reading stats as the first thing on the home page. A network hiccup (auth expiry, backend restart) caused the entire stats section to silently vanish — indistinguishable from "no data." The error state makes the failure visible and recoverable.

## Test plan

5 source-level tests in `frontend/src/__tests__/HomeStatsErrorState.test.ts`:
- `userStatsFetchError` state declared
- `userStatsRetryTick` state declared
- `.catch()` sets `userStatsFetchError`
- Error UI with Retry button present in source
- `userStatsRetryTick` in useEffect dependency array

All 3891 frontend tests pass.

Closes #2429

### DIFF
--- a/frontend/src/__tests__/DisclosureAriaControls.test.ts
+++ b/frontend/src/__tests__/DisclosureAriaControls.test.ts
@@ -24,7 +24,7 @@ describe("Disclosure button aria-controls (closes #2081)", () => {
     it("aria-controls and id appear near aria-expanded statsExpanded usage", () => {
       const idx = homeSrc.indexOf('aria-expanded={statsExpanded}');
       expect(idx).toBeGreaterThan(-1);
-      const window = homeSrc.slice(idx, idx + 3000);
+      const window = homeSrc.slice(idx, idx + 4000);
       expect(window).toContain('aria-controls="stats-activity-panel"');
       expect(window).toContain('id="stats-activity-panel"');
     });
--- a/frontend/src/__tests__/HomeStatsErrorState.test.ts
+++ b/frontend/src/__tests__/HomeStatsErrorState.test.ts
@@ -1,0 +1,38 @@
+/**
+ * Source-level regression tests for home page stats fetch-error state (issue #2429).
+ * The page is too large to mount with RTL, so we verify the pattern via source inspection.
+ */
+import fs from "fs";
+import path from "path";
+
+const SOURCE = fs.readFileSync(
+  path.resolve(__dirname, "../app/page.tsx"),
+  "utf8"
+);
+
+describe("Home page — stats fetch-error state (issue #2429)", () => {
+  it("declares userStatsFetchError state", () => {
+    expect(SOURCE).toMatch(/userStatsFetchError/);
+  });
+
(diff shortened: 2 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Eleventh empty-`.catch()` fix — and worth noting this is a *different* call site from #2413 (which fixed the `ReadingStats` component). #2413 fixed the shared component; this fixes `app/page.tsx`'s call to `getUserStats()` directly. Both needed independent treatment.

Home page is the entry route — for returning users the first impression of the app post-auth-blip is the entire "Your Progress" panel silently vanishing (streak, books started, vocab, annotations). That's a memorable "where did my data go?" failure mode, same impact tier as #2415's annotations sidebar fix.

Solid implementation: keeps the section header ("Your Progress") visible so the user understands what failed, error state inline where the stats normally render. The `+55/-38` page.tsx delta reflects JSX restructuring needed to thread the error path through existing render — acceptable given the surface.

Test window-bump on `DisclosureAriaControls.test.ts` (3000 → 4000) is a `#2181`-class brittleness, but acceptable for a single existing test that was at the edge of its window.

Clean +96/-40 across 4 files. Branched off main, disjoint from #2427.

Applying P2 (primary entry-route surface; user-perceived-data-loss tier matches #2415).

PR #2432, opened 2026-04-30
fix: show error state with retry in vocab tag filter on fetch failure (closes #2431)

### LINKED ISSUE #2431: ux: vocabulary page tag filter silently disappears on listVocabularyTags fetch failure
## Problem

On the vocabulary page, `listVocabularyTags()` is called with `.catch(() => {})`. If the fetch fails, `allTags` stays `[]` and the tag filter strip (`{!loading && words.length > 0 && allTags.length > 0 && (...)}`) is silently hidden — users can't filter their vocabulary by tags and receive no signal that loading failed.

## Reproduction

1. Open the vocabulary page while authenticated with saved words
2. Block the tags API call (network throttle or backend kill)
3. Observe: tag filter section is absent with no explanation

## Proposed fix

Add `tagsFetchError` + `tagsRetryTick` states. On failure show a compact inline retry message where the tag filter normally appears.

## Files

- `frontend/src/app/vocabulary/page.tsx` (lines 255, 278–282, 549, 569)

## Impact

Core vocabulary management flow — users who organize words with tags can't filter without understanding why tags are missing.

alfmunny:
## What changed

`app/vocabulary/page.tsx` called `listVocabularyTags()` with a silent `.catch(() => {})`. On failure `allTags` stays `[]` and the tag filter strip is silently hidden (it only renders when `allTags.length > 0`). Users who organize words with tags lose the ability to filter their vocabulary with no explanation.

This PR adds `tagsFetchError` + `tagsRetryTick` states. On rejection an inline "Couldn't load tags." message with a **Retry** button appears where the filter normally would. Clicking Retry increments the tick, re-triggering the useEffect.

## Why

The tag filter is a core vocabulary organization feature. A transient network error silently removing it leaves users confused about whether tags exist or why filtering isn't available.

## Test plan

5 source-level tests in `frontend/src/__tests__/VocabTagsFetchError.test.ts`:
- `tagsFetchError` state declared
- `tagsRetryTick` state declared
- `.catch()` sets `tagsFetchError`
- Error UI with Retry present in source
- `tagsRetryTick` in useEffect dependency array

All 3895 frontend tests pass.

Closes #2431

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -453,7 +453,8 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | Upload page: replaced silent getUploadQuota() failure with "Couldn't load quota" error state + Retry button — prevents misleading active upload zone when user may be at limit (closes #2423) | app/upload/page.tsx | #2425 |
 | 2026-04-30 | Profile Obsidian settings: replaced silent .catch() on getObsidianSettings failure with error state — shows "Couldn't load Obsidian settings" + Retry button inside accordion so user knows load failed (closes #2424) | app/profile/page.tsx | #2428 |
 | 2026-04-30 | TagEditor: replaced silent .catch() on initial tag fetch with loadFailed state — shows Retry button instead of misleading empty tag list (closes #2426) | components/TagEditor.tsx | #2427 |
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

12th empty-`.catch()` fix. Vocabulary tag filter strip silently disappearing on fetch failure (because `allTags=[]` looks identical to "no tags exist"). Same pattern, real but narrow scope — affects users who organize words with tags.

Same shape as the rest of the audit fixes (`fetchError` + `retryTick` states, error UI placed before the conditional render). Inline design-log entry.

Clean +56/-4 across 3 files. Branched off main, disjoint from #2430.

Applying P3 (real correctness, narrow scope).

PR #2434, opened 2026-04-30
fix: show inline error on annotation edit save failure (closes #2433)

### LINKED ISSUE #2433: ux: annotation edit save fails silently — no error feedback in notes/[bookId] page
## Problem

In `notes/[bookId]/page.tsx`, the `saveEdit()` function catches `updateAnnotation` failures silently:

```ts
} catch { /* keep edit open on failure */ }
```

The edit textarea remains open, but the user receives **zero feedback** — no error message, no retry option, nothing to indicate the save failed. This is a misleading silent-failure UX: the user may think the save is pending or succeeded, and then lose their edits by navigating away or refreshing.

## Expected behavior

When `updateAnnotation` rejects, show an inline error message (e.g. "Couldn't save — try again.") below the Save/Cancel buttons while keeping the edit open so the user can retry.

## Steps to reproduce

1. Open Notes for a book that has annotations with notes
2. Simulate a network failure (DevTools → Network → Offline)
3. Click Edit on an annotation, modify the note, click Save
4. Observe: edit stays open with no feedback (`/* keep edit open on failure */` comment is the only reaction)

## Fix

Add `editSaveError` state, set it on catch, clear it on attempt/success/cancel/new-edit. Pass it to `AnnotationCard` and render an inline error message.

## File

`frontend/src/app/notes/[bookId]/page.tsx` — `saveEdit()` at line 377

alfmunny:
## What changed

When `updateAnnotation` rejects during an annotation edit save in `notes/[bookId]/page.tsx`, the edit form now shows an inline error message — **"Couldn't save — try again."** — below the Save/Cancel buttons. Previously the catch block silently kept the edit open (`/* keep edit open on failure */`) with zero user feedback, leaving users unable to tell whether the save was pending, failed, or succeeded.

**Changes:**
- Added `editSaveError` state (cleared on attempt/success/cancel/new-edit)
- `saveEdit()` sets `editSaveError(true)` on catch; clears it on every attempt start
- `startEdit()` and cancel handler both clear the error
- `AnnotationCard` gains a `saveError` prop; renders a `role="alert"` error message when true

## Why

Silent save failures are a misleading dead-end: the user has edited a note, clicked Save, and receives no feedback. They may navigate away thinking the save is pending, losing their edit. This completes the silent-failure audit sweep started in #2412.

## Test plan

- [x] 5 source-level regression tests in `AnnotationEditSaveError.test.ts` covering: state declaration, error set on catch, error cleared on attempt, `saveError` prop wiring, error message render
- [x] Full frontend suite: 3900 tests pass
- [x] Backend failures (97) are pre-existing on `main`, unrelated to this frontend-only change

Closes #2433

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -454,7 +454,8 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | Profile Obsidian settings: replaced silent .catch() on getObsidianSettings failure with error state — shows "Couldn't load Obsidian settings" + Retry button inside accordion so user knows load failed (closes #2424) | app/profile/page.tsx | #2428 |
 | 2026-04-30 | TagEditor: replaced silent .catch() on initial tag fetch with loadFailed state — shows Retry button instead of misleading empty tag list (closes #2426) | components/TagEditor.tsx | #2427 |
 | 2026-04-30 | Home page stats panel: replaced silent getUserStats() failure with inline "Couldn't load stats" error + Retry — panel no longer disappears invisibly on network error (closes #2429) | app/page.tsx | #2430 |
-| 2026-04-30 | Vocabulary page tag filter: replaced silent listVocabularyTags() failure with inline "Couldn't load tags" error + Retry — filter strip no longer silently disappears (closes #2431) | app/vocabulary/page.tsx | — |
+| 2026-04-30 | Vocabulary page tag filter: replaced silent listVocabularyTags() failure with inline "Couldn't load tags" error + Retry — filter strip no longer silently disappears (closes #2431) | app/vocabulary/page.tsx | #2432 |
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

13th empty-`.catch()` fix and a real silent-data-loss pattern on the annotation edit path: user edits a note → clicks Save → `updateAnnotation` rejects → empty catch comment `/* keep edit open on failure */` was the *entire* error handling. User sees no error, no success acknowledgment, may navigate away thinking it saved. Same risk class as #2407 (AnnotationToolbar dismissal during save).

The `role="alert"` inline error message is the right pattern, and clearing the error on every attempt-start / cancel / new-edit is the right shape (no stale errors).

About the 97 backend test failures noted in the PR body — verified locally and they're env-specific (CI runs them green, including on the recent #2432). Looks like a local dev-environment divergence (probably migrations or virtualenv state), not a real main regression. Not blocking; flagging here so future PRs don't second-guess. Non-issue for this frontend-only change.

Clean +54/-3 across 3 files. Branched off main, disjoint from #2430.

Applying P2 (silent-data-loss on annotation edit, primary user-content flow, same risk tier as #2407).

PR #2436, opened 2026-04-30
fix: show error state on profile Gemini key status fetch failure (closes #2435)

### LINKED ISSUE #2435: ux: profile page shows Gemini key input when getMe() fails — misleading key-not-set state
## Problem

In `app/profile/page.tsx`, `hasKey` state initialises to `false` and is set only by a `getMe()` call that silently swallows errors:

```ts
getMe().then((me) => {
  setHasKey(me.hasGeminiKey);
  setIsAdmin(me.role === "admin");
}).catch(() => {}); // ← silent failure
```

If this fetch fails (network blip, token expiry), `hasKey` stays `false` and the UI renders the Gemini API key **input form** — showing the user "Set up your Gemini key" when they already have one configured. This is an actively misleading state: the user may re-enter their key unnecessarily, or think their key has been lost.

## Expected behavior

While the fetch is pending, show a subtle loading state. If the fetch fails, show an inline error with a Retry button (similar to the pattern used for Obsidian settings in the same page).

## Fix

- Add `geminiKeyFetchError` + `geminiKeyRetryTick` states
- Show error + Retry inline in the Gemini key section instead of the misleading key-input form
- Add `geminiKeyRetryTick` to the useEffect dependency array

## File

`frontend/src/app/profile/page.tsx` — `useEffect` at line 99

alfmunny:
## What changed

When `getMe()` rejects on the profile page, the Gemini key section no longer silently shows the API key input form — which falsely implies the user has no key configured. Instead it now shows an inline **"Couldn't load key status."** message with a Retry button.

**Changes:**
- Added `geminiKeyFetchError` + `geminiKeyRetryTick` states
- `useEffect` clears the error on each attempt and sets it on catch
- `geminiKeyRetryTick` added to the `useEffect` dependency array so the Retry button re-triggers the fetch
- Render: `geminiKeyFetchError → error+Retry | hasKey → key-active | else → key-input`

## Why

`hasKey` starts as `false`. If `getMe()` fails silently (network blip, stale token), users who already have a Gemini key see the "Enter your API key" input form — an actively misleading state that could cause them to re-enter their key unnecessarily or think their configuration was lost. This is the same silent-failure pattern fixed across the app in the audit sweep starting with #2412.

## Test plan

- [x] 6 source-level regression tests in `ProfileGeminiKeyFetchError.test.ts` covering: state declarations, error set on catch, error cleared on attempt, retry tick in deps, error message render
- [x] Full frontend suite: 3906 tests pass

Closes #2435

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -457,7 +457,7 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | Vocabulary page tag filter: replaced silent listVocabularyTags() failure with inline "Couldn't load tags" error + Retry — filter strip no longer silently disappears (closes #2431) | app/vocabulary/page.tsx | #2432 |
 | 2026-04-30 | Notes/[bookId]: annotation edit save failure now shows inline "Couldn't save — try again." error below Save/Cancel buttons instead of silently keeping edit open (closes #2433) | app/notes/[bookId]/page.tsx | #2434 |
 | 2026-04-30 | Profile page: getMe() failure no longer silently shows key-input form to users who already have a Gemini key — shows "Couldn't load key status" + Retry instead (closes #2435) | app/profile/page.tsx | #2436 |
-| 2026-04-30 | Profile page: listDecks() failure now shows "Couldn't load study decks" + Retry in the decks section instead of silently hiding it (closes #2437) | app/profile/page.tsx | — |
+| 2026-04-30 | Profile page: listDecks() failure now shows "Couldn't load study decks" + Retry in the decks section instead of silently hiding it (closes #2437) | app/profile/page.tsx | #2438 |
 
 ---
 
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

14th empty-`.catch()` fix — and a meaningful **action-misleading** failure mode in the same risk class as #2425 (upload quota). When `getMe()` fails silently, `hasKey` stays `false`, and users who *have* a Gemini key see the "Enter your API key" input form. Two real problems:
1. User may re-enter their key unnecessarily (creating cognitive friction + risk of accidental overwrites)
2. User may think their config was lost (trust-eroding)

The render order fix is exactly right: `geminiKeyFetchError → error+Retry | hasKey → key-active | else → key-input`. Error state takes precedence over the "no key" branch, eliminating the misleading-state path.

DIRTY from creation — #2434 merged just before this PR was opened, conflict on the design-log Change Log row. Quick rebase will resolve.

Clean +60/-4 across 3 files. Branched off main.

Applying P2 (action-misleading silent failure on credential state — same tier as #2425).

PR #2438, opened 2026-04-30
fix: show error state when profile Study decks fetch fails (closes #2437)

### LINKED ISSUE #2437: ux: profile Study decks section silently disappears when listDecks() fails
## Problem

In `app/profile/page.tsx`, the `listDecks()` fetch silently swallows errors:

```ts
listDecks()
  .then((d) => { if (alive) setDecks(d); })
  .catch(() => {
    // Swallow — section just hides
  })
  .finally(() => { if (alive) setDecksLoading(false); });
```

When `listDecks()` fails, `decks` stays `[]` and the loading state clears. Since the section only renders when `dueDecks.length > 0`, it silently **vanishes** — giving users with due cards no indication that their study session reminder failed to load. They see nothing where they expected flashcard prompts.

## Expected behavior

When the fetch fails, show an inline error state ("Couldn't load study decks") with a Retry button in the section area, so users know the section failed to load rather than assuming they have no due cards.

## Fix

- Add `decksFetchError` + `decksRetryTick` states
- Add `decksRetryTick` to `useEffect` dependency array
- Render error + Retry inline when `decksFetchError` is true (regardless of `dueDecks.length`)

## File

`frontend/src/app/profile/page.tsx` — `useEffect` at line 72

alfmunny:
## What changed

When `listDecks()` fails on the profile page, the Study decks section now shows **"Couldn't load study decks."** with a Retry button instead of silently disappearing. Previously, the `.catch()` swallowed the error (`// Swallow — section just hides`), left `decks` as `[]`, and cleared the loading state — causing the entire section to vanish for users who have due flashcards.

**Changes:**
- Added `decksFetchError` + `decksRetryTick` states
- `useEffect` clears error on attempt, sets it on catch
- `decksRetryTick` added to `useEffect` dependency array; Retry button increments it and resets the loading state
- Render: `decksFetchError → error+Retry | decksLoading → skeleton | dueDecks.length > 0 → deck list | else → nothing`

## Why

Users who have study decks with due cards rely on the profile page to see their "Study decks" reminder and navigate to flashcard review. When `listDecks()` fails, the section silently disappears — they see nothing where they expected prompts, with no indication of failure. This continues the silent-failure audit sweep (#2412 series).

## Test plan

- [x] 5 source-level regression tests in `ProfileDecksFetchError.test.ts`: state declarations, error set on catch, retry tick in deps, error message render
- [x] Full frontend suite: 3910 tests pass

Closes #2437

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -455,7 +455,9 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | TagEditor: replaced silent .catch() on initial tag fetch with loadFailed state — shows Retry button instead of misleading empty tag list (closes #2426) | components/TagEditor.tsx | #2427 |
 | 2026-04-30 | Home page stats panel: replaced silent getUserStats() failure with inline "Couldn't load stats" error + Retry — panel no longer disappears invisibly on network error (closes #2429) | app/page.tsx | #2430 |
 | 2026-04-30 | Vocabulary page tag filter: replaced silent listVocabularyTags() failure with inline "Couldn't load tags" error + Retry — filter strip no longer silently disappears (closes #2431) | app/vocabulary/page.tsx | #2432 |
-| 2026-04-30 | Notes/[bookId]: annotation edit save failure now shows inline "Couldn't save — try again." error below Save/Cancel buttons instead of silently keeping edit open (closes #2433) | app/notes/[bookId]/page.tsx | — |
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

15th empty-`.catch()` fix. Notable: the existing code had an explicit comment `// Swallow — section just hides` — so this was a deliberate (but ultimately wrong) design choice. The audit caught it and called it out cleanly.

Failure mode: users with due flashcards who rely on the profile page Study decks reminder don't see it when `listDecks()` fails. They could miss study sessions. Real impact, but recoverable — users can still navigate to flashcards via other paths.

Render order is right: error takes precedence over loading + empty branches.

Clean +56/-4 across 3 files. Branched off main, disjoint from #2436.

Applying P3 (real fix; users can recover via other navigation paths so not P2 like the action-misleading cases).

PR #2440, opened 2026-04-30
fix: show error state when flashcards deck selector fetch fails (closes #2439)

### LINKED ISSUE #2439: ux: flashcards page deck selector silently disappears when listDecks() fails
## Problem

In `app/vocabulary/flashcards/page.tsx`, the `listDecks()` fetch silently swallows errors:

```ts
listDecks()
  .then((d) => { ... setDecks(d); })
  .catch(() => {}); // silent
```

When the fetch fails, `decks` stays `[]` and the deck selector only renders when `decks.length > 0`, so it silently vanishes. Users who filter flashcard review to a specific deck see the selector disappear with no explanation — they can only study "All decks" and have no way to know the deck filter failed to load.

## Expected behavior

Show an inline error message ("Couldn't load decks") with a Retry button near where the deck selector would appear, so users understand the filter is unavailable due to a network error.

## Fix

- Add `decksFetchError` + `decksRetryTick` states
- Add `decksRetryTick` to `useEffect` dep array
- Render error + Retry when `decksFetchError` is true instead of silently hiding the selector

## File

`frontend/src/app/vocabulary/flashcards/page.tsx` — `useEffect` at line 87

alfmunny:
## What changed

When `listDecks()` fails on the flashcards page, the deck filter area now shows **"Couldn't load decks."** with a Retry button instead of silently hiding the selector. Previously the `.catch(() => {})` swallowed the error, `decks` stayed `[]`, and the selector never rendered — users studying specific decks had no way to know filtering was unavailable due to a network error.

**Changes:**
- Added `decksFetchError` + `decksRetryTick` states
- `useEffect` clears error on each attempt, sets it on catch
- `decksRetryTick` added to `useEffect` dep array; Retry button increments it
- Render: `decksFetchError → error+Retry | decks.length > 0 → selector | else → nothing`

## Why

Users who regularly filter flashcard review to a specific deck (e.g., "French verbs" only) rely on the deck selector being visible. When `listDecks()` fails silently, the entire filter control disappears — they fall back to "All decks" unknowingly. This is the same silent-disappear pattern fixed in the profile Study decks section (#2437/#2438).

## Test plan

- [x] 5 source-level regression tests in `FlashcardsDecksFetchError.test.ts`: state declarations, error set on catch (600-char window), retry tick in deps, error message render
- [x] Full frontend suite: 3915 tests pass

Closes #2439

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -458,6 +458,7 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | Notes/[bookId]: annotation edit save failure now shows inline "Couldn't save — try again." error below Save/Cancel buttons instead of silently keeping edit open (closes #2433) | app/notes/[bookId]/page.tsx | #2434 |
 | 2026-04-30 | Profile page: getMe() failure no longer silently shows key-input form to users who already have a Gemini key — shows "Couldn't load key status" + Retry instead (closes #2435) | app/profile/page.tsx | #2436 |
 | 2026-04-30 | Profile page: listDecks() failure now shows "Couldn't load study decks" + Retry in the decks section instead of silently hiding it (closes #2437) | app/profile/page.tsx | #2438 |
+| 2026-04-30 | Flashcards page: listDecks() failure now shows "Couldn't load decks" + Retry instead of silently removing the deck selector (closes #2439) | app/vocabulary/flashcards/page.tsx | #2440 |
 
 ---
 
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

16th empty-`.catch()` fix. Same family as #2438 (profile Study decks): users who normally filter flashcard review to a specific deck silently fall back to "All decks" when `listDecks()` fails. Real but recoverable — same P3 tier reasoning.

Standard shape (`fetchError` + `retryTick`, error branch first in render).

Inline design-log entry.

Clean +53/-3 across 3 files. Branched off main, disjoint from anything open.

Applying P3 (recoverable through other means; deck filter not user-blocking even when missing).

PR #2442, opened 2026-04-30
fix: hide admin action buttons when getMe() fails to prevent accidental self-delete (closes #2441)

### LINKED ISSUE #2441: ux: admin users page shows own action buttons when getMe() fails
## Problem

In `frontend/src/app/admin/users/page.tsx`, line 46:

```ts
getMe().then((me) => setMyId(me.id)).catch(() => {});
```

When `getMe()` fails (network error, server error), `myId` stays `null`. Since `u.id !== myId` is `true` for all users when `myId` is `null`, the admin sees Revoke/Delete buttons next to their own account — and the "You" label never appears. An admin could accidentally delete or revoke themselves.

## Fix

On catch, hide all action buttons and show a "Couldn't verify identity — Retry" notice so no destructive action can be taken while identity is unknown.

## Steps to reproduce

1. Open browser devtools → Network tab → block `/users/me`
2. Navigate to `/admin/users` as an admin
3. Observe: own account shows Revoke + Delete buttons; no "You" label

## Expected

While identity fetch is pending or failed: all action buttons are hidden; a short "Couldn't load identity" + Retry appears.

alfmunny:
## What changed

`frontend/src/app/admin/users/page.tsx`: `getMe()` previously swallowed errors with `.catch(() => {})`, leaving `myId` as `null`. Since `u.id !== null` is always true, every user row showed Revoke/Delete buttons including the admin's own account — and the "You" label never appeared.

**Fix:**
- Added `myIdFetchFailed` (boolean) and `myIdRetryTick` (counter) states
- On catch: `setMyIdFetchFailed(true)` instead of silent no-op
- Before each fetch attempt: `setMyIdFetchFailed(false)`
- Added `myIdRetryTick` to `useEffect` dep array so Retry re-fetches
- Action buttons now only render when `!myIdFetchFailed && u.id !== myId`
- When identity unknown: shows "Couldn't verify your identity — action buttons hidden" banner with Retry

## Why

Prevents an admin from accidentally revoking or deleting their own account when the identity fetch fails transiently. The banner makes the degraded state visible rather than silently removing the safety guard.

## Test plan

- [x] `AdminUsersGetMeError.test.ts` — 5 source-level regression tests
- [x] Full frontend suite: 3921 tests pass

Closes #2441

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -459,6 +459,7 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | Profile page: getMe() failure no longer silently shows key-input form to users who already have a Gemini key — shows "Couldn't load key status" + Retry instead (closes #2435) | app/profile/page.tsx | #2436 |
 | 2026-04-30 | Profile page: listDecks() failure now shows "Couldn't load study decks" + Retry in the decks section instead of silently hiding it (closes #2437) | app/profile/page.tsx | #2438 |
 | 2026-04-30 | Flashcards page: listDecks() failure now shows "Couldn't load decks" + Retry instead of silently removing the deck selector (closes #2439) | app/vocabulary/flashcards/page.tsx | #2440 |
+| 2026-04-30 | Admin users page: getMe() failure no longer silently exposes own Revoke/Delete buttons — hides all action buttons and shows "Couldn't verify identity" + Retry (closes #2441) | app/admin/users/page.tsx | #2442 |
 
 ---
 
--- a/frontend/src/__tests__/AdminUsersGetMeError.test.ts
+++ b/frontend/src/__tests__/AdminUsersGetMeError.test.ts
@@ -1,0 +1,37 @@
+/**
+ * Source-level regression tests for admin users page getMe() fetch-error state (issue #2441).
+ */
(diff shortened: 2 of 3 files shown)


alfmunny:
**PM**: PM-approved.

This is the most consequential empty-`.catch()` fix in the entire audit. The failure chain is genuinely scary:

1. `getMe()` fails silently (network blip, stale token)
2. `myId` stays `null`
3. The check `u.id !== myId` becomes `u.id !== null` — true for every user, **including the admin's own row**
4. Revoke/Delete buttons render on the admin's own row
5. "You" label never appears, so the admin can't visually identify their own row
6. Admin clicks Delete on what they think is a normal user → locks themselves out

That's a real data-destruction risk on a privileged surface — not display confusion, not action-misleading, but **silent removal of the safety guard that prevents self-deletion**.

The fix is precisely scoped: hide action buttons entirely when `myIdFetchFailed`, plus a banner explaining the degraded state. Defensive ordering is correct (`!myIdFetchFailed && u.id !== myId`). The Retry path re-attempts the identity fetch.

Clean +59/-8. Branched off main, disjoint from anything open.

Applying **P1** (real data-destruction risk on privileged admin surface; silent failure removes a safety guard, not just feedback). Strong upgrade from the typical empty-catch tier.

PR #2444, opened 2026-04-30
fix: prevent false Gemini key reminder when reader getMe() fails (null vs false) (closes #2443)

### LINKED ISSUE #2443: ux: reader shows misleading Gemini key reminder when getMe() fails (hasGeminiKey null)
## Problem

In `frontend/src/app/reader/[bookId]/page.tsx`, line 270-275:

```ts
useEffect(() => {
  getMe().then((me) => {
    setHasGeminiKey(me.hasGeminiKey);
    setIsAdmin(me.role === "admin");
  }).catch(() => {});  // silent failure
}, [session?.backendToken]);
```

`hasGeminiKey` is initialized as `null`. If `getMe()` fails (network blip, server restart), it stays `null`.

`notifyAIUsed()` at line 278 uses `!hasGeminiKey` — `null` is falsy, so it triggers the "Please set up your Gemini API key" reminder banner even for users who already have a key configured. The banner is misleading and suggests a problem that doesn't exist.

## Fix

- Change `!hasGeminiKey` to `hasGeminiKey === false` in `notifyAIUsed()` so it only fires when we have confirmed the user has no key
- On the `getMe()` catch: retry silently or show no-op (the strict equality fix is sufficient to prevent false reminders)

## Steps to reproduce

1. Block `/users/me` in devtools Network tab
2. Open any book in the reader
3. Use a translation or AI feature
4. Observe: "Please set up your Gemini API key" reminder appears even if key is configured

alfmunny:
## What changed

`frontend/src/app/reader/[bookId]/page.tsx`: changed `notifyAIUsed()` from using falsy check `!hasGeminiKey` to strict equality `hasGeminiKey === false`.

`hasGeminiKey` is initialized as `null` (not yet loaded). When `getMe()` fails transiently, it stays `null`. Since `null` is falsy, the old check fired the "Please set up your Gemini API key" reminder banner for users who actually have a key — a misleading false positive.

The catch block now has a comment explaining why `null` is intentionally left (the strict check in `notifyAIUsed` makes it safe).

## Why

Users with a configured Gemini key would see a confusing "set up your API key" banner after any transient network error during reader load. The fix: only show the reminder when we have positively confirmed `hasGeminiKey === false`.

## Test plan

- [x] `ReaderGetMeNullGeminiReminder.test.ts` — 2 source-level regression tests: strict equality in notifyAIUsed, no bare empty catch
- [x] Full frontend suite: 3923 tests pass

Closes #2443

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -460,6 +460,7 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | Profile page: listDecks() failure now shows "Couldn't load study decks" + Retry in the decks section instead of silently hiding it (closes #2437) | app/profile/page.tsx | #2438 |
 | 2026-04-30 | Flashcards page: listDecks() failure now shows "Couldn't load decks" + Retry instead of silently removing the deck selector (closes #2439) | app/vocabulary/flashcards/page.tsx | #2440 |
 | 2026-04-30 | Admin users page: getMe() failure no longer silently exposes own Revoke/Delete buttons — hides all action buttons and shows "Couldn't verify identity" + Retry (closes #2441) | app/admin/users/page.tsx | #2442 |
+| 2026-04-30 | Reader page: notifyAIUsed now uses strict `=== false` check so Gemini key reminder doesn't fire on transient getMe() failure when hasGeminiKey is null (closes #2443) | app/reader/[bookId]/page.tsx | #2444 |
 
 ---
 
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

18th empty-`.catch()` fix and the most surgical of the audit so far. The bug is a classic `null` vs `false` distinction:
- `hasGeminiKey` initialized as `null` (not yet loaded)
- `getMe()` fails silently → stays `null`
- `notifyAIUsed()` checked `!hasGeminiKey` (falsy)
- `null` is falsy → reminder banner fires for users who *have* a configured key

Fix is correct: `hasGeminiKey === false` only fires when we've positively confirmed no key. Same action-misleading tier as #2436 (also Gemini key, but on the profile surface vs. this on the reader).

Tiny diff (+36/-2) — just the strict-equality change, a comment explaining why the null-stays-null catch is intentional, and the regression test. Branched off main, disjoint from #2442.

Applying P2 (action-misleading on credential state — same tier as #2436).

PR #2446, opened 2026-04-30
fix: add role=status to reader translation queue and login-required banners (WCAG 4.1.3) (closes #2445)

### LINKED ISSUE #2445: ux: reader translation queue and login-required banners missing role=status (WCAG 4.1.3)
## Problem

In `frontend/src/app/reader/[bookId]/page.tsx`, two dynamically-shown notification banners lack `role="status"`, violating WCAG 4.1.3 (Status Messages):

1. **Translation queue banner** (line ~1384): appears when a chapter is queued for background translation. Screen readers won't announce it.
2. **Login required notice** (line ~1395): appears when translation is blocked for unauthenticated users. Screen readers won't announce it.

Both banners appear dynamically as `translationLoading` / `translationUsedProvider` state changes — exactly the scenario WCAG 4.1.3 requires `role="status"` for.

## Fix

Add `role="status"` to both banner divs.

## Steps to reproduce

1. Use a screen reader (VoiceOver/NVDA)
2. Open a book in the reader
3. Enable translation on a chapter with a queued worker
4. Observe: the "Translation queued" banner appears visually but is not announced

alfmunny:
## What changed

`frontend/src/app/reader/[bookId]/page.tsx`: added `role="status"` to two dynamically-shown notification banners:

1. **Translation queue banner** — appears when a chapter is queued for background translation
2. **Login required banner** — appears when translation is blocked for unauthenticated users

Both banners appear and disappear as `translationLoading` / `translationUsedProvider` state changes — exactly the scenario WCAG 4.1.3 (Status Messages) requires `role="status"` for. Without it, screen readers using JAWS/NVDA/VoiceOver never announce these banners.

## Why

WCAG 4.1.3 compliance: status messages that are injected dynamically must be in a live region (`role="status"` or `aria-live`) so assistive technologies announce them without requiring focus to move to the element.

## Test plan

- [x] `ReaderBannerRoleStatus.test.ts` — 2 source-level tests verifying `role="status"` on both banners
- [x] Full frontend suite: 3928 tests pass

Closes #2445

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -461,6 +461,7 @@ Systematic WCAG 2.1 AA pass covering loading states, dialog semantics, focus man
 | 2026-04-30 | Flashcards page: listDecks() failure now shows "Couldn't load decks" + Retry instead of silently removing the deck selector (closes #2439) | app/vocabulary/flashcards/page.tsx | #2440 |
 | 2026-04-30 | Admin users page: getMe() failure no longer silently exposes own Revoke/Delete buttons — hides all action buttons and shows "Couldn't verify identity" + Retry (closes #2441) | app/admin/users/page.tsx | #2442 |
 | 2026-04-30 | Reader page: notifyAIUsed now uses strict `=== false` check so Gemini key reminder doesn't fire on transient getMe() failure when hasGeminiKey is null (closes #2443) | app/reader/[bookId]/page.tsx | #2444 |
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Different category from the empty-`.catch()` audit — back to WCAG 4.1.3 work. Two reader banners (translation queue, login-required) appearing dynamically without `role="status"`, so AT users never hear them announced. Standard fix shape.

Tiny diff (+36/-2): just the `role="status"` attribute on each banner + a regression test + design log row. Branched off main, but DIRTY from creation (Change Log conflict with #2444 which just merged). Quick rebase will resolve.

Applying P3 (real WCAG AA on transient banners; narrower scope than the data-visibility fixes).

PR #2454, opened 2026-04-30
fix: replace router.push nav buttons with Link in upload/flashcards/reader (closes #2453)

alfmunny:
## What changed

Converts remaining pure-navigation `<button onClick={() => router.push(...)}>` elements to semantic `<Link href="...">` anchors across four pages:

- `upload/page.tsx` — "Sign in" (unauthenticated gate) and "Back" header button
- `upload/[bookId]/chapters/page.tsx` — "Try another file" (error state) and "Back" header button
- `vocabulary/flashcards/page.tsx` — back arrow and "Back to Vocabulary" CTA
- `reader/[bookId]/page.tsx` — "Back to library" error-state button

## Why

`<button onClick={router.push}>` for pure navigation violates WCAG 4.1.2 (buttons must perform actions, not navigation). `<Link>` renders a native `<a>` with correct role, enables right-click → "Open in new tab", and is handled by the browser's back-button history correctly.

Closes #2453.

## Test plan

- New source-level test `UploadFlashcardsNavLinks.test.ts` verifies all 7 converted locations use `href=` patterns, not `router.push`
- 12 existing test files updated: anchors changed from `router.push(...)` strings to `href=...` patterns; `getByRole("button")` queries for converted elements changed to `getByRole("link")`; click + `toHaveBeenCalledWith` assertions replaced with `toHaveAttribute("href", ...)`
- All 3966 tests pass

## Docs follow-up

N/A — internal interaction pattern, no user-visible documentation change needed.

PR #2452, opened 2026-04-30
fix: replace remaining router.push nav buttons with Link elements across all pages (closes #2451)

alfmunny:
## What changed

Replaced `<button onClick={() => router.push(...)}> ` navigation elements with semantic `<Link href="...">` across 11 pages and 1 component:

**Pages fixed:**
- `app/page.tsx` — Sign in header button, profile avatar, "Continue reading" card, "Sign in free" CTA
- `app/profile/page.tsx` — Library back button, Admin Panel link
- `app/vocabulary/page.tsx` — Library back button, Flashcards button, "Browse books" empty state CTA
- `app/notes/page.tsx` — Library back button, "Browse books" empty state CTA, book card navigation
- `app/notes/[bookId]/page.tsx` — Notes back button, "Open reader" empty state CTA
- `app/decks/page.tsx` — Library back button, New deck button (header + empty state)
- `app/decks/new/page.tsx` — Decks back button, Cancel button
- `app/decks/[deckId]/page.tsx` — Decks back button, "Back to decks" (error + smart deck empty state), "Start reading" CTA
- `app/admin/books/page.tsx` — "Open" reader button
- `app/admin/uploads/page.tsx` — "Open" reader button
- `app/reader/[bookId]/page.tsx` — Profile avatar button, Gemini key settings inline link

**Component fixed:**
- `components/DeckCard.tsx` — Added optional `href` prop; renders `<Link>` when provided, preserves `<button>` fallback for callers that still need onClick

**Intentionally preserved as `router.push`:**
- Auth guard redirects in `useEffect` (e.g. `router.replace("/login")`)
- Error state "Back to library" in reader page (contextual, not primary nav)
- Conditional button `vocab.length === 0 ? router.push("/") : setPickerOpen(true)` in deck detail header

## Why

`<button>` for pure page navigation is a semantic HTML bug:
- Prevents Ctrl+Click / middle-click to open in a new tab
- Screen readers announce as "button" not "link" (WCAG 4.1.2 violation)
- No href means the destination is invisible to search crawlers and assistive tech

## Test plan

- [x] `RemainingNavLinks.test.ts` — new source-level tests verifying all 11 pages use `href=` for navigation (not `router.push`)
- [x] 36 existing test files updated: `getByRole("button")` → `getByRole("link")`, `mockPush` assertions → `toHaveAttribute("href", ...)`
- [x] Full test suite: 3955 tests pass across 683 suites

Closes #2451

PR #2456, opened 2026-04-30
fix: use Link href='/' for deck zero-vocab 'Start reading' CTA (closes #2455)

### LINKED ISSUE #2455: ux: deck page 'Start reading' button should be a Link when vocabulary is empty
## Problem

In `app/decks/[deckId]/page.tsx`, the header button renders as 'Start reading' with `onClick={() => router.push('/')}` when `vocab.length === 0`. This is a pure navigation action to the home page but uses a `<button>` with `router.push` instead of a semantic `<Link href='/'>`.

**Impact:**
- Screen readers announce it as a button (expects action) rather than a link (expects navigation)
- Right-click → 'Open in new tab' doesn't work
- Violates WCAG 4.1.2

## Steps to reproduce
1. Open a manual deck with no vocabulary words saved
2. Observe 'Start reading' button in the header

## Fix
Conditionally render a `<Link href='/'>` when `vocab.length === 0`, and a `<button onClick={() => setPickerOpen(true)}` when vocab words are available.

alfmunny:
## What changed

In `app/decks/[deckId]/page.tsx`, split the single conditional button into two semantically correct elements:

- **`vocab.length === 0`** → `<Link href="/">Start reading</Link>` — pure navigation to the library home; renders as `<a>` with correct role, right-click → "Open in new tab" works, browser history is correct
- **`vocab.length > 0`** → `<button onClick={() => setPickerOpen(true)}>Add word</button>` — opens the word-picker dialog (action, not navigation)

Also removes the now-unused `useRouter` import.

## Why

`<button onClick={router.push('/')}>` for pure navigation violates WCAG 4.1.2 — buttons must perform actions, not navigation. Screen readers announce the element as a button (expecting an action), not a link (expecting navigation), which misleads keyboard and AT users.

Closes #2455.

## Test plan

- New source-level test `DeckStartReadingNavLink.test.ts`:
  - Verifies `router.push('/')` is NOT used in the zero-vocab conditional
  - Verifies `href="/"` IS present within the zero-vocab conditional path
- All 3968 tests pass

## Docs follow-up

N/A — internal interaction pattern change.

### DIFF
--- a/frontend/src/__tests__/DeckStartReadingNavLink.test.ts
+++ b/frontend/src/__tests__/DeckStartReadingNavLink.test.ts
@@ -1,0 +1,21 @@
+/**
+ * Regression test for #2455: deck page 'Start reading' button (zero-vocab state)
+ * must use a semantic Link href="/" not router.push("/").
+ */
+import fs from "fs";
+import path from "path";
+
+const src = fs.readFileSync(
+  path.join(process.cwd(), "src/app/decks/[deckId]/page.tsx"),
+  "utf8",
+);
+
+describe("decks/[deckId]/page — zero-vocab 'Start reading' nav (closes #2455)", () => {
+  it("does not use router.push('/') for the Start reading navigation", () => {
+    expect(src).not.toMatch(/vocab\.length\s*===?\s*0[^}]*router\.push\(["']\/["']\)/);
+  });
+
+  it("uses href='/' for the Start reading link when vocab is empty", () => {
+    expect(src).toMatch(/vocab\.length\s*===?\s*0[\s\S]{0,600}href=["']\/["']/);
+  });
+});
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Final straggler in the nav-link audit, and a thoughtful revisit of #2452's design call. #2452 had explicitly preserved `vocab.length === 0 ? router.push("/") : setPickerOpen(true)` as a "legitimate behavior switch" — but that's actually backwards reasoning. The branching means each rendered element has a *different* semantic role: Link in the empty-vocab branch, button in the populated branch. Splitting into two separate elements is the correct shape — author has clearly reconsidered and the new design is right.

Pattern-wise this completes the audit's evolution: #2452 was overly conservative in preserving "contextual" `router.push` sites (reader error state, deck conditional). #2454 reversed the reader exemption; #2456 reverses the deck one. Net result is consistent application of the rule.

Clean +45/-13 across 2 files (page +24/-13, test +21). Branched off main, disjoint from #1801. Removes now-unused `useRouter` import — small cleanup.

Applying P3 (final straggler, narrowest possible scope).

PR #2458, opened 2026-04-30
fix: use Link href for import page Skip button navigation (closes #2457)

### LINKED ISSUE #2457: ux: import page 'Skip' button uses router.push instead of semantic Link
## Problem

On `/import/[bookId]`, the **Skip** button (shown before the import starts) uses:

```tsx
<button onClick={() => router.push(nextUrl)}>Skip</button>
```

This is pure navigation with no side effects. Using a `<button>` for navigation:
- Screen readers announce "button" instead of "link" (wrong ARIA role)
- Users cannot right-click → open in new tab
- Keyboard users lose native browser back semantics

## Fix

Replace with `<Link href={nextUrl}>` from `next/link`.

The other two buttons in this component (`Cancel`, `Start reading now`) call `abortRef.current?.abort()` as a side effect — those must stay as `<button>` elements.

## File

`frontend/src/app/import/[bookId]/page.tsx` line 203–208

## Reproduction

Visit `/import/<bookId>` before clicking Start — the Skip button is rendered. Inspect it in DevTools; it is a `<button>` with an `onClick` handler that only navigates.

alfmunny:
## What changed

Converted the **Skip** button on `/import/[bookId]` (shown before import starts) from a `<button onClick={() => router.push(nextUrl)}>` to a semantic `<Link href={nextUrl}>`.

The other two buttons in that component — **Cancel** and **Start reading now** — call `abortRef.current?.abort()` as a side effect, so they correctly stay as `<button>` elements.

## Why

Using `<button>` for pure navigation violates WCAG 4.1.2 (name/role/value) and:
- Screen readers announce "button" instead of "link"
- Users cannot right-click → open in new tab
- Native browser back/forward semantics are lost

## Test plan

- New source-level regression test `ImportSkipButtonNavLink.test.ts` asserts `href={nextUrl}` and no `router.push`
- Updated `ImportPage.coverage.test.tsx` and `ImportPage.branches.test.tsx` to use `getByRole("link")` + `toHaveAttribute("href")` instead of button click + mockPush
- All 3970 frontend tests pass

Closes #2457

### DIFF
--- a/frontend/src/__tests__/ImportPage.branches.test.tsx
+++ b/frontend/src/__tests__/ImportPage.branches.test.tsx
@@ -309,15 +309,12 @@ describe("ImportPage — skipToReading (lines 213-216)", () => {
     expect(push).toHaveBeenCalledWith("/reader/42");
   });
 
-  it("Skip button before import start navigates to reader", async () => {
-    const { useRouter } = require("next/navigation");
-    const push = jest.fn();
-    useRouter.mockReturnValue({ push });
-
+  it("Skip link before import start points to reader URL", async () => {
     render(<BookImportPage />);
 
-    await clickAndFlush(/^Skip$/i);
-    expect(push).toHaveBeenCalledWith("/reader/42");
+    // "Skip" is now a semantic Link — check href rather than router.push
+    const skipLink = screen.getByRole("link", { name: /^Skip$/i });
+    expect(skipLink).toHaveAttribute("href", "/reader/42");
   });
 });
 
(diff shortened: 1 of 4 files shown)


alfmunny:
**PM**: PM-approved.

Sixth nav-link audit PR. Author correctly distinguishes the three import-page buttons:
- **Skip** → `<Link>` (pure navigation)
- **Cancel** → `<button>` (calls `abortRef.current?.abort()` — side effect)
- **Start reading now** → `<button>` (also aborts)

That's the right shape — only pure-nav-no-side-effects elements get the migration.

Tests follow the same `getByRole("link") + toHaveAttribute("href")` pattern that's now standard for this audit.

Clean +33/-18 across 4 files. Branched off main, disjoint from #1801.

Note: this is the 6th PR in the nav-link audit. I'm fine continuing to merge stragglers as they surface, but at this point the marginal value is decreasing — each new PR is finding 1-2 sites instead of dozens. Worth UI/UX considering whether the audit-via-PR approach is hitting diminishing returns vs. one final grep-driven sweep PR. Not blocking — applying P3.

PR #2460, opened 2026-04-30
fix: roving tabindex + arrow-key nav for AnnotationToolbar color picker (closes #2459)

alfmunny:
## What changed

Added roving tabindex and arrow-key navigation to the color picker in `AnnotationToolbar.tsx`.

**Before:** All 4 color buttons were individually tabbable (`tabIndex` unset → 0), with no keyboard navigation within the group.

**After:**
- Only the selected color button is in the tab sequence (`tabIndex={0}`); others are `tabIndex={-1}`
- `ArrowRight`/`ArrowDown` moves focus + selection to the next color (wraps around)
- `ArrowLeft`/`ArrowUp` moves focus + selection to the previous color (wraps around)
- Clicking a color still works as before

## Why

The color picker uses `role="radiogroup"` with `role="radio"` buttons — a custom ARIA widget. The [ARIA Authoring Practices Guide radiogroup pattern](https://www.w3.org/WAI/ARIA/apg/patterns/radio/) requires roving tabindex and arrow-key navigation for custom radio widgets. Without this, keyboard-only users must Tab through all 4 colors individually, breaking the expected interaction model. Fixes WCAG 2.1.1 (Keyboard) — AA.

Note: the TTS gender and translation provider pickers in `profile/page.tsx` use native `<input type="radio">` and already get arrow-key navigation from the browser — no change needed there.

## Test plan

- New `AnnotationToolbarColorKeyboard.test.tsx` (7 tests): roving tabindex initial state, ArrowRight/Left/Up/Down with and without wrap-around
- Updated `AnnotationToolbarColorTouchTargets.test.ts`: widened source-window to accommodate new props
- All 3977 frontend tests pass

Closes #2459

PR #2462, opened 2026-04-30
fix: roving tabindex + arrow-key nav for QuickHighlightPanel and SelectionToolbar toolbars (closes #2461)

### LINKED ISSUE #2461: ux: QuickHighlightPanel and SelectionToolbar role="toolbar" missing arrow-key navigation (WCAG 2.1.1)
## Problem

Two components use `role="toolbar"` but neither implements arrow-key navigation:

1. **`QuickHighlightPanel.tsx`** (line 113) — color picker + Note + Delete buttons
2. **`SelectionToolbar.tsx`** (line 152) — Read / Highlight / Note / Chat / Word buttons

The [ARIA APG toolbar pattern](https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/) requires:
- Only one element is in the tab sequence (roving tabindex)
- `ArrowRight` / `ArrowLeft` navigate between toolbar items
- `Home` / `End` jump to first / last item

Currently all buttons in both toolbars are Tab-focusable, which breaks the expected keyboard model for keyboard-only users.

## Note on SelectionToolbar

`SelectionToolbar` already has a `sr-only` message saying "Press Tab for actions" — so Tab-navigation was intentional. The fix should change the `sr-only` text to say "Use arrow keys for actions" to match the corrected behaviour.

## WCAG failure

2.1.1 (Keyboard) — AA

## Fix

For each toolbar:
- Apply roving tabindex (first item / focused item = `tabIndex=0`, others = `tabIndex=-1`)
- `onKeyDown` on the toolbar container: ArrowRight → focus next, ArrowLeft → focus prev, wrap around
- Store focused index in a ref

## Files

- `frontend/src/components/QuickHighlightPanel.tsx` line 112–115
- `frontend/src/components/SelectionToolbar.tsx` line 150–154

alfmunny:
## What changed

`QuickHighlightPanel` and `SelectionToolbar` both declare `role="toolbar"` but had no arrow-key navigation, violating the ARIA APG toolbar pattern (WCAG 2.1.1).

- **`QuickHighlightPanel`**: added `focusedToolbarIdx` state + `handleToolbarKeyDown` (ArrowLeft/Right + wrap-around) on the toolbar container; each button gets a dynamic `tabIndex` (0 for the focused item, -1 for all others) and an `onFocus` callback to keep state in sync with pointer/focus events.
- **`SelectionToolbar`**: same pattern; button indices computed via a running counter so conditional buttons (all five props are optional) land at the correct slot automatically; sr-only announcement updated from "Press Tab" → "Use arrow keys".
- **`RemainingComponentsFocusRing.test.ts`**: widened two source-window lengths (400 → 600) and updated the color-button anchor after the onclick refactor pushed the className further in the source.

## Why

Closes #2461. Keyboard-only users had no way to reach the second (or later) toolbar button without a mouse; only the first button was reachable via Tab.

## Test plan

- [x] 5 new tests in `QuickHighlightToolbarKeyboard.test.tsx` — initial tabIndex state, ArrowRight/Left forward/backward, wrap-around both directions
- [x] 5 new tests in `SelectionToolbarKeyboard.test.tsx` — same coverage for SelectionToolbar, plus announcement text check
- [x] 3987 / 3987 tests pass (`npm test -- --no-coverage --ci`)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/frontend/src/__tests__/QuickHighlightToolbarKeyboard.test.tsx
+++ b/frontend/src/__tests__/QuickHighlightToolbarKeyboard.test.tsx
@@ -1,0 +1,84 @@
+/**
+ * Regression test for #2461: QuickHighlightPanel role="toolbar" must implement
+ * roving tabindex and arrow-key navigation per ARIA APG toolbar pattern.
+ */
+import React from "react";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+jest.mock("@/lib/api", () => ({
+  createAnnotation: jest.fn(),
+  updateAnnotation: jest.fn(),
+  deleteAnnotation: jest.fn(),
+}));
+
+import QuickHighlightPanel from "@/components/QuickHighlightPanel";
+
+const BASE_PROPS = {
+  sentenceText: "Call me Ishmael.",
+  chapterIndex: 0,
+  bookId: 10,
+  position: { x: 200, y: 200 },
+  onClose: jest.fn(),
+  onSaved: jest.fn(),
+  onDeleted: jest.fn(),
+  onOpenNote: jest.fn(),
+};
+
+beforeEach(() => {
+  jest.clearAllMocks();
+});
+
+describe("QuickHighlightPanel toolbar — roving tabindex (closes #2461)", () => {
+  it("only the first (Yellow) color button is in the tab sequence initially", () => {
+    render(<QuickHighlightPanel {...BASE_PROPS} />);
+    const yellow = screen.getByRole("button", { name: "Yellow" });
+    const blue   = screen.getByRole("button", { name: "Blue" });
+    const green  = screen.getByRole("button", { name: "Green" });
+    const pink   = screen.getByRole("button", { name: "Pink" });
+    expect(yellow).toHaveAttribute("tabindex", "0");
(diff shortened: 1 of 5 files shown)


alfmunny:
**PM**: PM-approved.

Continuation of #2460's keyboard interaction work, now on `role="toolbar"` widgets. Two components:

1. **QuickHighlightPanel** — standard roving tabindex + ArrowLeft/Right + wrap-around
2. **SelectionToolbar** — same pattern but with a clever **running-counter** approach: since all 5 buttons are optional props, button indices are computed based on which are actually rendered. Conditional buttons land at the correct slot automatically. Strong handling of the variable-length toolbar case.

**Notable detail**: the sr-only announcement is updated from "Press Tab for actions" → "Use arrow keys for actions". Important — without this, AT users would have been told the wrong interaction model. Author caught this.

Test design: 2 new RTL behavioral test files (5 + 5 tests, 215 lines combined). Pattern matches #2460.

Clean +333/-36 across 5 files (2 components + 2 new tests + 1 existing test window-bump). Branched off main, disjoint from #1801.

Applying P2 (real WAI-ARIA APG toolbar pattern compliance on two primary reader-flow toolbars; same tier as #2460).

PR #2464, opened 2026-04-30
fix: move focus to first button when QuickHighlightPanel opens (closes #2463)

alfmunny:
## What changed

Added a `useEffect(() => { toolbarRef.current?.querySelector<HTMLButtonElement>("button")?.focus(); }, [])` to `QuickHighlightPanel.tsx` so focus moves to the Yellow button immediately when the panel mounts.

Also updated `QuickHighlightToolbarKeyboard.test.tsx` to cover this (one new test: "Yellow button receives focus immediately when panel mounts").

## Why

Closes #2463. When `QuickHighlightPanel` appeared (after clicking an annotated span or from the selection toolbar), focus stayed on the triggering element — typically a non-focusable `<span>`. Keyboard users had no indication the panel was open and no way to reach it without Tab-traversing through intervening elements.

`AnnotationToolbar` (full note editor) already calls `textareaRef.current?.focus()` on mount; this PR applies the same pattern to the colour-only quick panel.

## Test plan

- [x] 1 new test: `QuickHighlightPanel focus-on-open` — `Yellow button receives focus immediately when panel mounts`
- [x] 3988 / 3988 tests pass (`npm test -- --no-coverage --ci`)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

PR #2469, opened 2026-04-30
fix: announce flashcard session-complete state to screen readers (closes #2467)

### LINKED ISSUE #2467: ux: flashcard session-complete state not announced to screen readers (WCAG 4.1.3)
## Problem

When the user grades the last flashcard, the `done` state renders in `app/vocabulary/flashcards/page.tsx` (line 268):

```tsx
{!fetchError && done ? (
  <div className="flex flex-col items-center gap-4 py-16 text-center">
    <h2>All done for today!</h2>
```

This is a plain `<div>` with no `role` or `aria-live`, so screen readers are not guaranteed to announce the completion. The card counter live region (`role="status"` at line 354) disappears at the same time, leaving no accessible announcement.

## Expected

Add `role="status"` (or `role="alert"` for assertive) to the done-state container so the completion message is announced when the element appears.

## Fix

Change `<div className="...">" to `<div role="status" aria-label="Session complete" className="...">`.

## WCAG reference

WCAG 4.1.3 Status Messages — status changes not given focus must be programmatically determinable so ATs can announce them.

alfmunny:
## What changed

Added `role="status"` to the flashcard session-complete ("All done for today!") container in `app/vocabulary/flashcards/page.tsx`.

## Why

When the last card is graded and the done state appears, screen readers receive no announcement because the containing `<div>` is a plain element with no live region. WCAG 4.1.3 (Status Messages) requires that status changes conveyed visually are also communicated to AT without requiring focus. `role="status"` (implicit `aria-live="polite"`) satisfies this.

## Test plan

- New regression test: `src/__tests__/FlashcardDoneStateAnnouncement.test.ts`
  - Asserts `role="status"` is present on the container that holds "All done for today!"
- Full Jest suite: 3990 tests, all pass

Closes #2467

### DIFF
--- a/frontend/src/__tests__/FlashcardDoneStateAnnouncement.test.ts
+++ b/frontend/src/__tests__/FlashcardDoneStateAnnouncement.test.ts
@@ -1,0 +1,29 @@
+/**
+ * Regression test for #2467: flashcard session-complete ("All done for today!")
+ * state must be wrapped in a live region so screen readers announce it when it
+ * appears (WCAG 4.1.3 Status Messages).
+ */
+import { readFileSync } from "fs";
+import { join } from "path";
+
+const src = readFileSync(
+  join(__dirname, "../app/vocabulary/flashcards/page.tsx"),
+  "utf-8"
+);
+
+describe("Flashcard done-state live region (closes #2467)", () => {
+  it("done-state container has role=status", () => {
+    const doneIdx = src.indexOf("All done for today!");
+    expect(doneIdx).toBeGreaterThan(-1);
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Tiny WCAG 4.1.3 fix on the flashcard "All done for today!" complete state. Without `role="status"`, AT users grading their last card don't hear the success message — the visual transition is invisible to them. Implicit `aria-live="polite"` from `role="status"` is the right shape (no manual aria-live attribute needed).

Clean +30/-1. Branched off main, disjoint from #2464.

Applying P3 (small but real WCAG AA fix on session-completion celebration moment).

PR #2470, opened 2026-04-30
fix: roving tabindex + arrow-key nav for SentenceActionPopup toolbar (closes #2468)

alfmunny:
## What changed

`SentenceActionPopup` declared `role="toolbar"` but all buttons had implicit `tabIndex=0`, so Tab visited each button individually — violating the ARIA APG toolbar keyboard interaction pattern.

Changes to `frontend/src/components/SentenceActionPopup.tsx`:
- Added `focusedToolbarIdx` state (roving tabindex)
- Added `handleToolbarKeyDown` with ArrowLeft/ArrowRight navigation and wrap-around
- Added `onKeyDown` + `ref` on the toolbar div
- Each button gets `tabIndex={focusedToolbarIdx === idx ? 0 : -1}` and `onFocus` sync

## Why

The ARIA APG toolbar pattern requires the toolbar to be a single Tab stop, with ArrowLeft/ArrowRight moving focus between items. Without this, keyboard users must Tab through every button in the popup — degrading the interaction for both keyboard and AT users (WCAG 4.1.2).

This completes the toolbar-keyboard series: AnnotationToolbar (#2459), SelectionToolbar + QuickHighlightPanel (#2461), and now SentenceActionPopup (#2468).

## Test plan

- New test: `src/__tests__/SentenceActionPopupKeyboard.test.tsx` — 5 tests covering initial tabIndex state, ArrowRight/Left navigation, and wrap-around in both directions
- Updated `ReaderComponentsFocusRing.test.ts` window width (450→650) to accommodate new props
- Full Jest suite: 3993 tests, all pass

Closes #2468

PR #2472, opened 2026-04-30
fix: move focus to first button when SentenceActionPopup opens (closes #2471)

### LINKED ISSUE #2471: ux: SentenceActionPopup does not move focus to toolbar on open (WCAG 2.4.3)
## Problem

When a sentence is tapped/clicked to show `SentenceActionPopup`, focus stays on the body or the sentence span (a non-interactive element). The popup appears visually, but keyboard users have no way to reach it without tabbing through the entire page.

`QuickHighlightPanel` had the same issue and was fixed in PR #2464 by adding a mount-time `useEffect` that focuses the first toolbar button.

`frontend/src/components/SentenceActionPopup.tsx`

## Expected behaviour

When the popup mounts, the first button (Read) should receive focus automatically, allowing keyboard users to immediately use the toolbar.

## Fix

Add a mount-time `useEffect` after the existing ones:
```tsx
useEffect(() => {
  toolbarRef.current?.querySelector<HTMLButtonElement>('button')?.focus();
}, []);
```

Note: depends on the `toolbarRef` added in #2468 (PR #2470). Target this branch off main once #2470 merges.

Closes #2471 (self-referential placeholder — update after creation).

alfmunny:
## What changed

Added a mount-time `useEffect` to `SentenceActionPopup` that focuses the first toolbar button (Read) when the popup appears.

## Why

When the popup appears (triggered by a click on a sentence span), focus stayed on the body/span — a non-interactive element. Keyboard users had no way to reach the popup without tabbing through the entire page. WCAG 2.4.3 (Focus Order) requires focus to follow the user's intent; a popup should receive focus when it opens. This matches the pattern already used in `QuickHighlightPanel` (PR #2464).

## Test plan

- New test: `src/__tests__/SentenceActionPopupFocusOnOpen.test.tsx` — 2 tests
  - Read button receives focus when panel mounts (with all optional buttons)
  - Read button receives focus when only Read is present (no Note/Chat)
- Full Jest suite: 3993 tests, all pass

Closes #2471

### DIFF
--- a/frontend/src/__tests__/SentenceActionPopupFocusOnOpen.test.tsx
+++ b/frontend/src/__tests__/SentenceActionPopupFocusOnOpen.test.tsx
@@ -1,0 +1,41 @@
+/**
+ * Regression test for #2471: SentenceActionPopup must move focus to the first
+ * toolbar button when it mounts, so keyboard users can reach it immediately
+ * without tabbing through the entire page (WCAG 2.4.3 Focus Order).
+ */
+import React from "react";
+import { render, screen } from "@testing-library/react";
+
+import SentenceActionPopup from "@/components/SentenceActionPopup";
+
+const BASE_PROPS = {
+  sentenceText: "Call me Ishmael.",
+  position: { x: 200, y: 200 },
+  onRead: jest.fn(),
+  onNote: jest.fn(),
+  onChat: jest.fn(),
+  onClose: jest.fn(),
+};
+
+beforeEach(() => {
+  jest.clearAllMocks();
+});
+
+describe("SentenceActionPopup focus-on-open (closes #2471)", () => {
+  it("Read button receives focus immediately when panel mounts", () => {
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Same shape as #2464 (focus on QuickHighlightPanel mount) applied to SentenceActionPopup. Standard WCAG 2.4.3 (Focus Order) fix — without it, focus stays on the non-interactive triggering span and keyboard users have no way to reach the popup.

Notable: tests cover both the all-buttons-present case and the minimal Read-only case (when optional Note/Chat aren't passed). Right defensive shape.

Clean +45/-0 across 2 files. Branched off main but **likely will conflict with #2470** since both touch SentenceActionPopup.tsx in close proximity. Whichever lands first will trigger a rebase on the other. Quick resolution.

Applying P3 (small focus-management polish).

PR #2474, opened 2026-04-30
fix: restore focus to trigger element when QuickHighlightPanel and SentenceActionPopup close (closes #2473)

### LINKED ISSUE #2473: ux: QuickHighlightPanel and SentenceActionPopup do not restore focus on close (WCAG 2.4.3)
## Problem

When `QuickHighlightPanel` or `SentenceActionPopup` open, they correctly move focus into the panel (WCAG 2.4.3). However when they close (via Escape, outside click, or action button), focus is not restored to the previously-focused element.

`WordLookup` (src/components/WordLookup.tsx:77-85) does this correctly:
```ts
const previouslyFocused = document.activeElement as HTMLElement | null;
ref.current?.focus();
return () => { previouslyFocused?.focus?.(); };
```

Both popup components lack this pattern. For keyboard users who tabbed into a sentence and pressed Enter to trigger a panel, pressing Escape leaves focus stranded on `document.body` instead of returning it to the sentence.

## Fix

Add a mount-time `useEffect` that captures `document.activeElement` and restores it in the cleanup function, matching the `WordLookup` pattern.

## Affected files

- `src/components/QuickHighlightPanel.tsx`
- `src/components/SentenceActionPopup.tsx` (after PR #2470 merges)

## WCAG reference

WCAG 2.4.3 Focus Order — focus must not be lost when a dialog or popup closes.

alfmunny:
## What changed

`QuickHighlightPanel` and `SentenceActionPopup` both move focus into themselves on open, but did not return focus to the previously-focused element when unmounting. This left keyboard users stranded on `document.body` after pressing Escape or clicking an action button.

Fixed by capturing `document.activeElement` at mount time and restoring it in the `useEffect` cleanup, matching the existing pattern in `WordLookup` (src/components/WordLookup.tsx:77-85).

## Why

WCAG 2.4.3 Focus Order requires that when a dialog or popup closes, focus returns to the element that triggered it. Keyboard users who Tab-navigated to a sentence and opened a panel would lose their scroll position and reading context.

## Test plan

- [x] New test `PopupFocusRestoreOnClose.test.tsx` with 2 regression tests (one per component) — both confirmed failing before the fix, passing after
- [x] All 694 frontend test suites pass
- [x] Existing focus-ring and focus-on-open tests unaffected

Closes #2473

### DIFF
--- a/frontend/src/__tests__/PopupFocusRestoreOnClose.test.tsx
+++ b/frontend/src/__tests__/PopupFocusRestoreOnClose.test.tsx
@@ -1,0 +1,89 @@
+/**
+ * Regression test for #2473: QuickHighlightPanel and SentenceActionPopup must
+ * restore focus to the previously-focused element when they unmount, so
+ * keyboard users do not lose their position in the document (WCAG 2.4.3).
+ */
+import React from "react";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+jest.mock("@/lib/api", () => ({
+  createAnnotation: jest.fn(),
+  updateAnnotation: jest.fn(),
+  deleteAnnotation: jest.fn(),
+}));
+
+import { createAnnotation, updateAnnotation, deleteAnnotation } from "@/lib/api";
+import QuickHighlightPanel from "@/components/QuickHighlightPanel";
+import SentenceActionPopup from "@/components/SentenceActionPopup";
+
+const mockCreate = createAnnotation as jest.MockedFunction<typeof createAnnotation>;
+const mockUpdate = updateAnnotation as jest.MockedFunction<typeof updateAnnotation>;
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Completes the focus-management dance for QuickHighlightPanel + SentenceActionPopup:
- #2464 / #2472: focus into the popup on open
- #2474 (this): restore focus to trigger on close

Standard WCAG 2.4.3 Focus Order pattern — capture `document.activeElement` at mount, restore in useEffect cleanup. Author correctly cites the existing WordLookup implementation (src/components/WordLookup.tsx:77-85) as the reference. Pattern consistency.

Test plan is the right shape — 2 regression tests (one per component) confirmed failing before the fix and passing after. That's the gold standard for focus-restore tests.

Clean +95/-0 across 3 files (component changes +3 each, new test +89). Branched off main, disjoint from #1801.

Applying P3 (small focus-management polish completing the dialog focus dance).

PR #2476, opened 2026-04-30
fix: restore focus to trigger element when TypographyPanel closes (closes #2475)

### LINKED ISSUE #2475: ux: TypographyPanel does not restore focus on close (WCAG 2.4.3)
## Problem

When the Typography settings panel is dismissed (via Escape or clicking outside), focus drops to `document.body` instead of returning to the button that opened it.

**Reproduction:**
1. Open the reader page
2. Press Tab until the Typography settings button is focused, then Enter to open the panel
3. Press Escape to close the panel
4. Focus is lost (drops to document.body) — keyboard user must re-navigate from the top

## Expected behaviour (WCAG 2.4.3 Focus Order)

Focus should return to the element that had focus before the panel opened — i.e., the Typography settings button.

## Fix

Capture `document.activeElement` at mount and restore it in the `useEffect` cleanup, matching the pattern used in `QuickHighlightPanel` and `SentenceActionPopup`:

```tsx
useEffect(() => {
// eslint-disable-next-line react-hooks/exhaustive-deps
}, []);
```

Closes this issue.

alfmunny:
## What changed

`TypographyPanel` now captures `document.activeElement` at mount and restores it in the `useEffect` cleanup, matching the focus-restoration pattern already present in `QuickHighlightPanel` and `SentenceActionPopup`.

## Why

WCAG 2.4.3 (Focus Order) requires that when a UI component closes, focus returns to the element that had focus before the component opened. Without this fix, a keyboard user who tabs into the Typography panel and presses Escape loses their focus position — they must re-navigate from the top of the page.

Closes #2475.

## Test plan

- New regression test in `TypographyPanelFocusRestore.test.tsx`: focuses a trigger button, renders the panel, moves focus into a panel button, unmounts the panel, asserts focus returns to the trigger.
- Full suite: 4001 tests pass.

### DIFF
--- a/frontend/src/__tests__/TypographyPanelFocusRestore.test.tsx
+++ b/frontend/src/__tests__/TypographyPanelFocusRestore.test.tsx
@@ -1,0 +1,53 @@
+/**
+ * Regression test for #2475: TypographyPanel must restore focus to the
+ * previously-focused element when it unmounts (WCAG 2.4.3 Focus Order).
+ *
+ * Scenario: user tabs into the panel, then presses Escape. Focus should
+ * return to the element that was focused before the panel opened.
+ */
+import React from "react";
+import { render } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import TypographyPanel from "@/components/TypographyPanel";
+
+const BASE_PROPS = {
+  fontSize: "base" as const,
+  lineHeight: "normal" as const,
+  contentWidth: "normal" as const,
+  fontFamily: "serif" as const,
+  paragraphFocus: false,
+  onFontSize: jest.fn(),
+  onLineHeight: jest.fn(),
+  onContentWidth: jest.fn(),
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Fourth focus-restore implementation in this codebase using the same pattern: WordLookup → QuickHighlightPanel/SentenceActionPopup (#2474) → TypographyPanel (this). The pattern is now well-established. Test design matches #2474's gold standard (focus trigger → render → move focus inside → unmount → assert restored).

Clean +59/-0 across 2 files. Branched off main, disjoint from #1801.

Note: at this point the focus-restore pattern is essentially codebase boilerplate. Worth UI/UX considering extracting into a `useFocusRestore` hook to avoid copy-paste drift. Not blocking — the pattern is short enough that 4 copies is still acceptable, but worth flagging if a 5th appears.

Applying P3 (small focus-management completion).

PR #2335, opened 2026-04-29
fix: add sr-only new-tab announcement to export URL links (closes #2334)

alfmunny:
## What changed

Added `<span className="sr-only"> (opens in new tab)</span>` inside the export result link anchors in two places:

- `frontend/src/app/vocabulary/page.tsx` — Obsidian export URL link in the export status area
- `frontend/src/app/reader/[bookId]/page.tsx` — Obsidian export toast URL link

Both links render a dynamic URL as visible link text and open in a new tab via `target="_blank"`, but previously gave screen-reader users no indication that a new tab would open.

## Why

WCAG 2.4.4 (Link Purpose) and 3.2.2 (On Input) require that link behaviour unexpected by users (opening a new context) be announced. The pattern `<span className="sr-only"> (opens in new tab)</span>` is already established across VocabWordTooltip, BookDetailModal, vocabulary Wiktionary link, and the reader Gutenberg link — these two export links were the only remaining gaps.

## Test plan

- [x] New static-analysis tests in `ExportUrlNewTabAnnouncement.test.ts` assert both locations contain the sr-only text — confirmed failing before fix, passing after
- [x] Updated `VocabularyPage.test.tsx` and `VocabularyPage.branches.test.tsx` to use regex name matching (accessible name now includes "(opens in new tab)")
- [x] Full suite: 3708 tests pass (`npm test -- --no-coverage --ci`)

Closes #2334

PR #2478, opened 2026-04-30
fix: show Wiktionary fallback link when no vocabulary definition found (closes #2477)

### LINKED ISSUE #2477: ux: vocabulary definition sheet shows dead-end 'No definition found' with no fallback
## Problem

When the Wiktionary lookup returns no definitions for a word, the `DefinitionSheet` shows:

> *No definition found.*

There is no action the user can take — no retry, no link to look it up manually, no suggestion. This is a dead-end state that abandons the user.

**File:** `frontend/src/app/vocabulary/page.tsx:210-212`

```tsx
{!loading && !fetchError && (!def || def.definitions.length === 0) && (
  <p className="text-sm text-stone-600 italic">No definition found.</p>
)}
```

## Proposed Fix

Add a fallback Wiktionary link in the 'No definition found' state so the user can look the word up directly:

```tsx
{!loading && !fetchError && (!def || def.definitions.length === 0) && (
  <div className="space-y-2">
    <p className="text-sm text-stone-600 italic">No definition found.</p>
    <a
      href={`https://en.wiktionary.org/wiki/${encodeURIComponent(word)}`}
      target="_blank"
      rel="noopener noreferrer"
      className="text-xs text-amber-700 hover:text-amber-800 hover:underline ..."
    >
      Search Wiktionary <ArrowUpRightIcon .../><span className="sr-only"> (opens in new tab)</span>
    </a>
  </div>
)}
```

## Acceptance criteria
- [ ] When no definition is found, a 'Search Wiktionary' link appears below the message
- [ ] Link opens `https://en.wiktionary.org/wiki/<word>` in a new tab
- [ ] Link has screen reader text '(opens in new tab)'
- [ ] Test covers the no-definition empty state rendering the fallback link

alfmunny:
## What changed

When the Wiktionary lookup returns no definitions for a word, `DefinitionSheet` previously showed only:

> *No definition found.*

with no action the user could take. Now it renders a **"Search Wiktionary"** link below that message, constructed from `def.url` when available or `https://en.wiktionary.org/wiki/<word>` as a fallback. The link opens in a new tab with a `sr-only` "(opens in new tab)" label for screen readers.

## Why

Dead-end states break the user's flow. A vocabulary user who gets no definition should have a one-click path to look the word up directly rather than being abandoned.

## Files changed

- `frontend/src/app/vocabulary/page.tsx` — added Wiktionary fallback link in the no-definition branch of `DefinitionSheet`
- `frontend/src/__tests__/VocabNoDefinitionFallbackLink.test.tsx` — two regression tests: empty-definitions case and null-def case

## Test plan

- [x] `VocabNoDefinitionFallbackLink.test.tsx` — both cases pass
- [x] Existing `VocabularyPage.definition.test.tsx` suite unchanged (4002 total tests pass)
- [x] Backend tests: 1941 passed

Closes #2477

## Docs follow-up

N/A — minor UX fix, no new interaction pattern to document.

### DIFF
--- a/frontend/src/__tests__/VocabNoDefinitionFallbackLink.test.tsx
+++ b/frontend/src/__tests__/VocabNoDefinitionFallbackLink.test.tsx
@@ -1,0 +1,95 @@
+/**
+ * Regression test for #2477: DefinitionSheet must show a Wiktionary fallback
+ * link when no definitions are found, so the user has a next step instead of
+ * a dead-end "No definition found." message.
+ */
+import React from "react";
+import { render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+jest.mock("next-auth/react", () => ({
+  useSession: () => ({ data: { backendToken: "token123" } }),
+}));
+
+jest.mock("next/navigation", () => ({
+  useRouter: () => ({ push: jest.fn() }),
+  useSearchParams: () => ({ get: () => null }),
+}));
+
+jest.mock("@/lib/api", () => ({
+  getVocabulary: jest.fn(),
+  deleteVocabularyWord: jest.fn(),
+  exportVocabularyToObsidian: jest.fn(),
+  getWordDefinition: jest.fn(),
+  listVocabularyTags: jest.fn().mockResolvedValue([]),
+  getVocabularyWordTags: jest.fn().mockResolvedValue([]),
+  addVocabularyWordTag: jest.fn().mockResolvedValue({ tag: "" }),
+  removeVocabularyWordTag: jest.fn().mockResolvedValue(undefined),
+  ApiError: class ApiError extends Error { status = 500; },
+}));
+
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Real UX dead-end fix. "No definition found." was a non-actionable terminal state — user has to navigate away or guess at alternatives. The "Search Wiktionary" fallback link with `def.url` preferred and `https://en.wiktionary.org/wiki/<word>` as fallback is the right shape.

Includes the standard `sr-only` "(opens in new tab)" label — consistent with the WCAG G201 export-link work shipped earlier today (#2335 etc.). Pattern uniformity.

Test plan covers both the empty-array and null-def cases (defensive shape for the no-defs path).

Clean +106/-1 across 2 files. Branched off main, disjoint from #1801.

Applying P3 (real UX dead-end fix; narrow scope, but eliminates a definite no-action terminal state).

PR #2480, opened 2026-04-30
fix: add Retry button to search error state (closes #2479)

### LINKED ISSUE #2479: ux: search page shows error with no retry button — user must retype to recover
## Problem

When the in-app search fails (network error, API timeout, etc.), the search page shows:

> Error: Search failed

with no way to retry except modifying the search query. Since the search is driven by the URL query param (`?q=`), the user must change their query to trigger a re-fetch — clearing and retyping the same term is the only recovery path.

**File:** `frontend/src/app/search/page.tsx:176-181`

```tsx
{!loading && error && (
  <div role="alert" className="...">
    <AlertCircleIcon ... />
    <span>Error: {error}</span>
    {/* No retry button */}
  </div>
)}
```

The search `useEffect` fires on `[q]` changes — if `q` hasn't changed, there's no way to re-run it.

## Proposed Fix

Add a **Retry** button inside the error state that increments a `retryTick` counter included in the `useEffect` dependency array, triggering a re-fetch without requiring a query change. Also use `.finally(() => setLoading(false))` for robustness.

## Acceptance criteria
- [ ] Retry button appears inside the search error state
- [ ] Clicking Retry re-runs the same search query
- [ ] Loading state is reset correctly before/after retry
- [ ] Test covers the error state rendering the retry button

alfmunny:
## What changed

When the in-app search API fails, the error banner previously showed only:

> Error: Search failed

with no way to recover except retyping the query. Now:

1. A **Retry** button appears inside the error banner — clicking it re-runs the same search without requiring any query change.
2. The search `useEffect` now uses `.finally(() => setLoading(false))` instead of manual `setLoading(false)` calls in both `.then()` and `.catch()`, preventing the spinner from getting stuck if state updates throw.

## Why

Dead-end error states break user flow. A user whose search failed due to a transient network error should be able to recover with one click, not by re-typing.

## Files changed

- `frontend/src/app/search/page.tsx` — added `retryTick` state, Retry button in error banner, `.finally()` cleanup
- `frontend/src/__tests__/SearchPageErrorRetry.test.tsx` — two regression tests: error state shows Retry button; clicking Retry re-runs the search and clears the error

## Test plan

- [x] `SearchPageErrorRetry.test.tsx` — both cases pass
- [x] Existing `SearchPage.errorstate.test.tsx` unchanged (all 4003 frontend tests pass)
- [x] Backend tests: 1941 passed

Closes #2479

## Docs follow-up

N/A — minor UX fix in an existing error state.

### DIFF
--- a/frontend/src/__tests__/SearchPageErrorRetry.test.tsx
+++ b/frontend/src/__tests__/SearchPageErrorRetry.test.tsx
@@ -1,0 +1,57 @@
+/**
+ * Regression test for #2479: search error state must include a Retry button
+ * so users can re-run the same query without having to retype it.
+ */
+import React from "react";
+import { render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+const mockSearchParams = { get: (k: string) => (k === "q" ? "hamlet" : null) };
+
+jest.mock("next/navigation", () => ({
+  useSearchParams: () => mockSearchParams,
+  useRouter: () => ({ push: jest.fn() }),
+}));
+
+const mockSearchFn = jest.fn();
+jest.mock("@/lib/api", () => ({
+  searchInAppContent: (...args: unknown[]) => mockSearchFn(...args),
+}));
+
+import SearchPage from "@/app/search/page";
+
+const flushPromises = () => new Promise((r) => setTimeout(r, 0));
+
+beforeEach(() => {
+  mockSearchFn.mockReset();
+});
+
+describe("search error state — retry button (closes #2479)", () => {
+  it("shows a Retry button inside the error alert", async () => {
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Standard error+Retry pattern from the empty-catch audit (now applied to search). Bonus: refactors `setLoading(false)` from manual calls in both `.then` and `.catch` to a single `.finally` — prevents stuck-spinner if a state update throws between the success/failure branch. Defensive improvement, separate from the immediate scope but worth shipping together.

Test plan covers both presence-of-Retry-on-error and Retry-actually-clears-error+re-runs. Right shape.

Clean +69/-10 across 2 files. Branched off main, disjoint from #2478.

Applying P3 (real UX dead-end fix; same family as the audit retries).

PR #2483, opened 2026-04-30
fix: add Wiktionary fallback link to WordLookup error state (closes #2482)

### LINKED ISSUE #2482: ux: WordLookup popup shows dead-end error with no fallback link when dictionary lookup fails
## Problem

When a user selects a word in the reader and the dictionary API (`dictionaryapi.dev`) returns no result, `WordLookup.tsx` renders:

```
No definition found for word
```

with no next step for the user. The popup becomes a dead end with only the Close button as an action.

## Expected behaviour

Same pattern as the vocabulary definition sheet (fixed in #2477): show a "Search Wiktionary" fallback link so the user has a path forward.

```tsx
{error && (
  <div className="space-y-1 pr-6">
    <p role="alert" className="text-amber-700 italic">{error} for &ldquo;<span lang={lang}>{word}</span>&rdquo;</p>
    <a
      href={`https://en.wiktionary.org/wiki/${encodeURIComponent(word)}`}
      target="_blank"
      rel="noopener noreferrer"
      ...
    >
      Search Wiktionary <ArrowUpRightIcon ... />
      <span className="sr-only"> (opens in new tab)</span>
    </a>
  </div>
)}
```

## Files

- `frontend/src/components/WordLookup.tsx` line 123–125

## Reproduction

2. Long-press / select an obscure word or a word in a language not supported by the dictionary API
3. See dead-end error with no actionable next step

alfmunny:
## What changed

`WordLookup.tsx` now renders a "Search Wiktionary" fallback link when the dictionary API returns no result, replacing the previous dead-end "No definition found" message.

## Why

When a user selects an obscure word or a word in an unsupported language in the reader, the popup showed only an error message with no actionable next step. This is the same dead-end pattern fixed for the vocabulary definition sheet in #2477/#2478.

## Test plan

- New regression test: `frontend/src/__tests__/WordLookupErrorFallbackLink.test.tsx`
  - Covers API 404 response → link appears with correct href/target/rel
  - Covers fetch network rejection → link URL constructed from word prop
- Full frontend suite: 4007 tests pass

### DIFF
--- a/frontend/src/__tests__/WordLookupErrorFallbackLink.test.tsx
+++ b/frontend/src/__tests__/WordLookupErrorFallbackLink.test.tsx
@@ -1,0 +1,54 @@
+/**
+ * Regression test for #2482: WordLookup must show a Wiktionary fallback link
+ * when the dictionary API returns no definition, so the popup is never a dead end.
+ */
+import React from "react";
+import { render, screen, waitFor } from "@testing-library/react";
+import WordLookup from "@/components/WordLookup";
+
+const DEFAULT_POSITION = { x: 200, y: 300 };
+
+beforeEach(() => {
+  jest.clearAllMocks();
+});
+
+describe("WordLookup — error fallback link (closes #2482)", () => {
+  it("shows a Wiktionary link when the API returns 404", async () => {
+    global.fetch = jest.fn().mockResolvedValue({ ok: false, json: jest.fn() });
+
+    render(
+      <WordLookup
+        word="xyzzy"
+        position={DEFAULT_POSITION}
+        language="en"
+        onClose={jest.fn()}
+      />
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Same Wiktionary fallback pattern as #2478, now applied to WordLookup (the reader-flow tap-a-word popup vs. the vocabulary page's DefinitionSheet). Both now have actionable escape paths instead of dead-end "No definition found" terminals.

Test plan covers both 404-response and network-rejection paths — defensive shape.

Clean +66/-2 across 2 files. Branched off main, disjoint from #1801.

Applying P3 (consistent UX dead-end fix; brings the reading-flow tooltip in line with the vocab page).

PR #2494, opened 2026-04-30
ux: make gender toggle aria-label conditional on loading state (closes #2493)

### LINKED ISSUE #2493: ux: TTS gender toggle aria-label says "Click to switch" when button is disabled
## Problem

The gender toggle button in `TTSControls.tsx` has:

```tsx
aria-label={`Voice: ${gender === "female" ? "Female" : "Male"}. Click to switch.`}
disabled={status === "loading"}
```

When `status === "loading"`, the button is disabled but the accessible name still says **"Click to switch."** This violates WCAG 4.1.2 (Name, Role, Value): the accessible name describes an action the user cannot perform, which is misleading for screen-reader users who rely on aria-labels to decide whether to interact with a control.

## Expected behaviour

When the button is disabled, the aria-label should not say "Click to switch" — it should only describe the current state.

## Proposed fix

```tsx
aria-label={`Voice: ${gender === "female" ? "Female" : "Male"}${status !== "loading" ? ". Click to switch" : ""}`}
```

## File

`frontend/src/components/TTSControls.tsx`, line 462

## WCAG

4.1.2 Name, Role, Value (Level A)

alfmunny:
## What changed

The TTS gender toggle button's `aria-label` now omits "Click to switch" when the button is disabled (i.e., when `status === "loading"`).

**Before:** `Voice: Female. Click to switch.` (even when button is disabled)
**After:** `Voice: Female` (when loading) / `Voice: Female. Click to switch` (when active)

## Why

Closes #2493. WCAG 4.1.2 (Name, Role, Value, Level A): the accessible name of a disabled control must not describe an action the user cannot currently perform. A screen-reader user hearing "Click to switch" on a disabled button gets misleading information.

## Test plan

- `TTSControlsGenderAriaDisabled.test.ts` (new) — static analysis confirms the aria-label expression contains a `loading`-based guard on the "Click to switch" phrase, and that the voice identifier is preserved in all states.
- All 10 existing TTS test suites (79 tests) pass.
- Full frontend suite: 4014/4014 passing after rebasing onto main (which includes the merged #2491 commit).

## Docs follow-up

N/A — no user-visible documentation needed for this fix.

### DIFF
--- a/frontend/src/__tests__/TTSControlsGenderAriaDisabled.test.ts
+++ b/frontend/src/__tests__/TTSControlsGenderAriaDisabled.test.ts
@@ -1,0 +1,39 @@
+/**
+ * Regression test for #2493:
+ * The TTS gender toggle button's aria-label must not say "Click to switch"
+ * when the button is disabled (status === "loading") — WCAG 4.1.2.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../components/TTSControls.tsx"),
+  "utf8",
+);
+
+describe("TTSControls gender toggle disabled aria-label (closes #2493)", () => {
+  it("aria-label line is conditional — references loading/disabled state inside the expression", () => {
+    // Extract just the aria-label line for the gender toggle button.
+    // The fix makes "Click to switch" conditional, so the aria-label expression
+    // itself must contain a reference to "loading" or "disabled" state.
+    const toggleIdx = src.indexOf("onClick={toggleGender}");
+    expect(toggleIdx).toBeGreaterThan(-1);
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Clean WCAG 4.1.2 fix: disabled control should not advertise an action the user cannot take. The conditional template is the minimal correct expression, and preserves the voice identifier in both states (so screen-reader users still get the value, just not the misleading affordance).

Static-analysis test asserts the conditional structure is preserved (loading-guard on "Click to switch", voice identifier in all states). +40/-1 across 2 files. Disjoint from #1801.

Applying P3 (single-control aria-label nit on a state where the user can't act anyway; small but real).

PR #2496, opened 2026-04-30
ux: omit aria-label (not empty string) on idle QueueTab spinner (closes #2495)

### LINKED ISSUE #2495: ux: QueueTab inline spinner role=status has empty aria-label when not loading
## Problem

In `QueueTab.tsx` line 1335, the inline spinner span uses:

```tsx
<span role="status" aria-label={loadingItems ? "Loading items" : ""}>
  {loadingItems && <Spinner />}
</span>
```

When `loadingItems` is `false`, `aria-label=""` — an empty string — is set on a `role="status"` element. The ARIA specification states that an empty string is not a valid accessible name; landmark elements with no name or an empty name may be skipped by screen readers or announced inconsistently.

## Expected behaviour

When the span is empty (not loading), the `aria-label` attribute should be absent, not set to `""`.

## Fix

```tsx
aria-label={loadingItems ? "Loading items" : undefined}
```

Passing `undefined` causes React to omit the attribute entirely, which is the correct state for an empty status landmark.

## File

`frontend/src/components/QueueTab.tsx`, line 1335

## WCAG

4.1.2 Name, Role, Value (Level A)

alfmunny:
## What changed

`QueueTab.tsx` inline spinner: changed `aria-label=""` (empty string when not loading) to `aria-label={undefined}` so the attribute is omitted entirely when not loading.

**Before:** `aria-label={loadingItems ? "Loading items" : ""}` — when idle, produces `aria-label=""` on the `role="status"` element.
**After:** `aria-label={loadingItems ? "Loading items" : undefined}` — when idle, the attribute is absent.

## Why

Closes #2495. WCAG 4.1.2 (Name, Role, Value, Level A): the ARIA spec states that an empty string is not a valid accessible name. Landmark elements with empty names may be skipped or announced inconsistently by screen readers.

## Test plan

- `QueueTabSpinnerAriaLabel.test.ts` (new) — static analysis confirms the empty-string pattern is gone and a meaningful label is still present when loading.
- All 26 QueueTab test suites (183 tests) pass.
- Full frontend suite: 4016/4016 passing.

## Docs follow-up

N/A — logged in Wave 11 change table under docs/design-improvement-plan.md.

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -532,3 +532,12 @@ Dead-end error states given actionable fallback paths so users are never strande
 | 2026-04-30 | Search page: added Retry button to error state + `.finally()` for loading cleanup | app/search/page.tsx | #2480 |
 | 2026-04-30 | Vocabulary DefinitionSheet: Wiktionary fallback link when definition empty/null | app/vocabulary/page.tsx | #2478 |
 | 2026-04-30 | WordLookup: Wiktionary fallback link in error state when dictionary API finds nothing | components/WordLookup.tsx | #2483 |
+
+## Wave 11 — Keyboard Accessibility (2026-04-30)
+
+WCAG 2.1.1 (Keyboard) and 4.1.2 (Name, Role, Value) fixes for keyboard-only and AT users.
+
+| Date | Change | File(s) | PR |
+|------|--------|---------|----|
+| 2026-04-30 | SelectionToolbar: auto-focus first button on keyboard-driven text selection (Shift+Arrow); restore prior focus on close | components/SelectionToolbar.tsx | #2492 |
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Correct WCAG 4.1.2 fix on a `role="status"` landmark. Empty string is not a valid accessible name per the ARIA spec — passing `undefined` so React omits the attribute entirely is the right pattern for an empty/idle status landmark.

Static-analysis regression confirms the empty-string pattern is gone and the meaningful label is preserved on the loading path. Wave 11 Change Log entry included in the same PR — proper per-PR docs check this time, good (vs. the retro-batched #2487 pattern).

+45/-1 across 3 files. Disjoint from #1801.

Applying P3 (single-attribute fix on a low-traffic admin surface; small but valid a11y polish).

PR #2502, opened 2026-04-30
fix: move focus to done-state container when flashcard session completes (closes #2501)

### LINKED ISSUE #2501: ux: focus lost when flashcard session completes (WCAG 2.4.3)
## Problem

When a user grades the last flashcard and the "All done for today!" state appears, focus is lost.

**Reproduction steps:**
1. Open /vocabulary/flashcards with cards due
2. Review all cards, pressing number keys (1–4) to grade
3. After grading the final card, observe: the grade buttons unmount and the done state renders
4. Focus drops to `<body>` — the "Back to Vocabulary" link is not focused

**Root cause:** The grade buttons are conditionally rendered inside `!fetchError && currentCard ? ...`. When `done` becomes `true`, the buttons unmount and React doesn't move focus to the new content.

**Fix:** Add a `useEffect` that watches `done` and moves focus to the done-state heading or the "Back to Vocabulary" link when it becomes `true`. Attach a `useRef` to the done-state container and call `focus()` on it.

## Acceptance criteria

- After grading the last card, focus moves to the "All done for today!" heading or the "Back to Vocabulary" link
- Screen reader users hear the completion message announced
- No regression in keyboard navigation during normal card review

## Files

- `frontend/src/app/vocabulary/flashcards/page.tsx`

alfmunny:
## What changed

When a user grades the last flashcard and the "All done for today!" state appears, focus was dropped to `<body>` because the grade buttons unmounted without any focus management. Keyboard and screen-reader users lost their position (WCAG 2.4.3 Focus Order).

**Fix:**
- Added `useRef` (`doneContainerRef`) on the done-state `<div>`
- Added `useEffect` that calls `doneContainerRef.current?.focus()` when `done` becomes `true`
- Added `tabIndex={-1}` so the div is programmatically focusable without entering the tab order
- Added `focus:outline-none` since the focus is intentional/programmatic

## Why

Without explicit focus management, after grading the last card:
- Keyboard users had no focused element; Tab would jump unpredictably
- Screen readers would not announce the "All done for today!" state

## Test plan

- [x] New regression tests in `FlashcardDoneFocus.test.tsx`:
  1. After grading the last card, `document.activeElement` is the done-state container
  2. The done-state container has `tabIndex="-1"` (is programmatically focusable)
- [x] All 4026 frontend tests pass
- [x] Backend tests pass

Closes #2501

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -543,3 +543,5 @@ WCAG 2.1.1 (Keyboard) and 4.1.2 (Name, Role, Value) fixes for keyboard-only and 
 | 2026-04-30 | TTS gender toggle: aria-label omits "Click to switch" when button is disabled (WCAG 4.1.2) | components/TTSControls.tsx | #2494 |
 | 2026-04-30 | QueueTab inline spinner: aria-label changed from empty string to undefined when not loading (WCAG 4.1.2) | components/QueueTab.tsx | #2496 |
 | 2026-04-30 | Deck form counters: text-stone-400 → text-stone-600 on parchment background; contrast 2.3:1 → 6.7:1 (WCAG 1.4.3) | app/decks/new/page.tsx | #2497 |
+| 2026-04-30 | Annotations tutorial: added "Select a phrase (partial text selection)" section with Shift+Arrow, Arrow Left/Right, Escape keyboard flow documentation | docs/tutorials/annotations.md | #2500 |
+| 2026-04-30 | Flashcard done-state: tabIndex=-1 + useEffect focus-move on done=true; prevents focus loss when grade buttons unmount (WCAG 2.4.3) | app/vocabulary/flashcards/page.tsx | #2501 |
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Real WCAG 2.4.3 (Focus Order) fix on the flashcards completion flow. Conditional unmount of the grade buttons → focus drops to `<body>` is exactly the failure pattern these focus-management PRs have been addressing across the toolbars; this completes the same pattern on a non-toolbar flow.

Implementation matches the standard recipe — `useRef` on the container, `useEffect` on the `done` transition, `tabIndex={-1}` for programmatic focusability without polluting the tab order, `focus:outline-none` since the focus shift is intentional. Pairs nicely with #2469's `role="status"` on the done-state — combined, the SR announcement fires AND focus lands on the new content.

Regression tests cover both required invariants: `document.activeElement` resolves to the container, and `tabIndex=-1` is preserved (the latter guards against accidental future removal that would re-introduce the tab-order pollution).

+106/-2 across 3 files; design-log entry in same PR. Disjoint from #1801.

Applying P2 (WCAG 2.4.3 on a primary user flow per CLAUDE.md priority tier).

PR #2506, opened 2026-04-30
ux: add title/aria-label to heatmap legend swatches for colorblind users (closes #2505)

### LINKED ISSUE #2505: ux: reading stats heatmap legend uses color alone to convey intensity levels (WCAG 1.4.1)
## Problem

The reading activity heatmap in `ReadingStats.tsx` has a legend with five colored squares (stone-100 → amber-800) and only "Less" / "More" text labels. Color is the only visual means of conveying which intensity level each square represents.

Users with color vision deficiency (colorblindness) can see the squares but cannot distinguish which color maps to which activity count. This violates WCAG 1.4.1 (Use of Color).

## Location

`frontend/src/components/ReadingStats.tsx` — the legend section at the bottom of the heatmap (`intensityClass` thresholds: 0, 1–2, 3–5, 6–10, 11+).

## Proposed fix

Add visible count labels below each legend swatch (or as `title` tooltip + an accessible description). Minimal change: replace the raw colored divs with labeled items showing the count range, e.g.:

```
Less  [ ]0  [ ]1-2  [ ]3-5  [ ]6-10  [ ]11+  More
```

Or add `title` attributes to each swatch and an `aria-describedby` sr-only scale description.

## Acceptance criteria

- Legend swatches convey the count range through a non-color indicator (visible text or tooltip)
- WCAG 1.4.1 satisfied
- No regressions to heatmap rendering

alfmunny:
## What changed

Added `title` and `aria-label` attributes to each legend swatch in the reading activity heatmap, naming the count range it represents:

- `0 events` (stone-100)
- `1–2 events` (amber-200)
- `3–5 events` (amber-400)
- `6–10 events` (amber-600)
- `11+ events` (amber-800)

Previously the legend had five colored squares with only "Less" / "More" text — color was the sole indicator of intensity.

## Why

WCAG 1.4.1 (Use of Color) requires that color is not the only visual means of conveying information. Users with color vision deficiency (colorblindness) could see the squares but could not identify which color maps to which activity count. The `title` attribute makes the count range visible as a hover tooltip; the `aria-label` exposes it to AT.

## Test plan

- New static source test `ReadingStatsHeatmapLegend.test.ts` verifies:
  1. `11+` label exists for the highest intensity bucket
  2. `1–2` label exists for the second bucket
  3. Legend section contains `title` or `aria-label` attributes
- Full frontend test suite: 4032 tests pass

Closes #2505

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -545,3 +545,5 @@ WCAG 2.1.1 (Keyboard) and 4.1.2 (Name, Role, Value) fixes for keyboard-only and 
 | 2026-04-30 | Deck form counters: text-stone-400 → text-stone-600 on parchment background; contrast 2.3:1 → 6.7:1 (WCAG 1.4.3) | app/decks/new/page.tsx | #2497 |
 | 2026-04-30 | Annotations tutorial: added "Select a phrase (partial text selection)" section with Shift+Arrow, Arrow Left/Right, Escape keyboard flow documentation | docs/tutorials/annotations.md | #2500 |
 | 2026-04-30 | Flashcard done-state: tabIndex=-1 + useEffect focus-move on done=true; prevents focus loss when grade buttons unmount (WCAG 2.4.3) | app/vocabulary/flashcards/page.tsx | #2501 |
+| 2026-04-30 | Upload chapter word-count pill: added AlertCircleIcon when word count < 100 or > 8000; color alone is insufficient per WCAG 1.4.1 | app/upload/[bookId]/chapters/page.tsx | #2503 |
+| 2026-04-30 | Reading stats heatmap legend: added title + aria-label to each swatch (0, 1–2, 3–5, 6–10, 11+) so colorblind users can identify intensity levels (WCAG 1.4.1) | components/ReadingStats.tsx | #2505 |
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved with note.

The `aria-label` on each swatch is the right primitive for AT users — that satisfies WCAG 1.4.1 in letter. The `title` tooltip helps sighted mouse users see the count range without re-rendering the legend.

**Trade-off worth flagging for next iteration:** the issue's first-suggested fix was visible inline labels (`[ ]0 [ ]1-2 [ ]3-5 [ ]6-10 [ ]11+`). That's a stronger non-color indicator because it covers keyboard-only sighted users *without* AT — the population that gets neither the `title` (no hover) nor the `aria-label` (no SR). The chosen tooltip+aria path is defensible per the issue's "or" phrasing, but visible labels would be the more universal solution. Not blocking this PR; flagging if heatmap legend ever gets revisited.

Static source-analysis test asserts the labelled count buckets exist (`11+`, `1–2`) and that title/aria attributes are present. +41/-3 across 3 files; design-log entry in same PR.

Disjoint from #1801.

Applying P3 (legend on a stats/non-actionable surface; the WCAG-letter fix is in but the stronger universal fix is left as follow-up).

PR #2510, opened 2026-04-30
ux: pause UndoToast timer on hover/focus; increase duration 3s→5s (closes #2509)

### LINKED ISSUE #2509: ux: UndoToast 3-second auto-dismiss fails WCAG 2.2.1 Timing Adjustable
## Problem

`UndoToast` auto-dismisses after 3 seconds (`DURATION_MS = 3000`) with no mechanism to pause, extend, or turn off the timer. WCAG 2.2.1 (Level A) requires that time limits be adjustable unless an exception applies; none of the exceptions (real-time, essential, 20-hour) cleanly apply here.

**Concrete failure path:**  
A keyboard-only user deletes a vocabulary word → toast appears fixed at page bottom → user needs to Tab from wherever focus currently is through potentially many elements to reach the "Undo" button → 3 seconds elapses → toast dismisses → Undo is lost.

## Fix

1. Increase base duration from 3 s → 5 s.
2. Pause the auto-dismiss timer while the user hovers over or has keyboard focus inside the toast — restart on `mouseleave` / `blur`.

This gives keyboard users unlimited time to reach "Undo" once they're focused in the toast, satisfying WCAG 2.2.1 via the "user can extend the time limit with a simple action" path (focusing the toast pauses the clock).

## Files

`frontend/src/components/UndoToast.tsx`

## Acceptance

- \[ \] Duration changed to 5000 ms
- \[ \] Timer pauses on `mouseenter` / `focusin`, resumes on `mouseleave` / `focusout`  
- \[ \] Unit test: timer is paused when focus enters the toast; resumed + dismisses after it leaves

alfmunny:
## What changed

`UndoToast` auto-dismissed after 3 seconds with no way to pause, extend, or turn off the timer. This is a WCAG 2.2.1 (Timing Adjustable, Level A) failure: a keyboard user who triggers "delete" must Tab from their current position to the fixed-position "Undo" button — 3 seconds is often not enough.

**Changes:**
- `DURATION_MS` increased from 3 000 ms → 5 000 ms (more breathing room by default)
- Auto-dismiss timer pauses on `mouseenter` / `focusin` and restarts on `mouseleave` / `focusout` — once a keyboard user focuses inside the toast, the timer is frozen indefinitely until they leave

## Why

WCAG 2.2.1 Success Criterion requires that time limits be turn-offable, adjustable, or extendable (with at least 20 seconds of extension). None of the exceptions (real-time, essential, 20-hour) apply to a deletion undo toast. The pause-on-focus pattern satisfies the "extend with a simple action" path.

## Test plan

- [x] `UndoToastTimingPause.test.tsx` — 4 new tests covering: base-duration dismissal at 5 s, focus-pauses-timer, mouseenter-pauses-timer, Undo-click works
- [x] `UndoToast.test.tsx` — updated timing from 3 300 ms → 5 300 ms to match new duration
- [x] Full frontend suite: 4037 tests pass

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -548,3 +548,4 @@ WCAG 2.1.1 (Keyboard) and 4.1.2 (Name, Role, Value) fixes for keyboard-only and 
 | 2026-04-30 | Upload chapter word-count pill: added AlertCircleIcon when word count < 100 or > 8000; color alone is insufficient per WCAG 1.4.1 | app/upload/[bookId]/chapters/page.tsx | #2503 |
 | 2026-04-30 | Reading stats heatmap legend: added title + aria-label to each swatch (0, 1–2, 3–5, 6–10, 11+) so colorblind users can identify intensity levels (WCAG 1.4.1) | components/ReadingStats.tsx | #2505 |
 | 2026-04-30 | Flashcard card-to-card transition: flipButtonRef + useEffect focus-move on currentIndex change; prevents focus loss when grade buttons unmount between cards (WCAG 2.4.3) | app/vocabulary/flashcards/page.tsx | #2507 |
+| 2026-04-30 | UndoToast: increased auto-dismiss from 3 s → 5 s; timer now pauses on mouseenter/focusin and resumes on mouseleave/focusout — WCAG 2.2.1 Timing Adjustable (Level A) | components/UndoToast.tsx | #2509 |
--- a/frontend/src/__tests__/UndoToast.test.tsx
+++ b/frontend/src/__tests__/UndoToast.test.tsx
@@ -26,13 +26,13 @@ describe("UndoToast", () => {
     await waitFor(() => expect(onDone).toHaveBeenCalledTimes(1), { timeout: 500 });
   });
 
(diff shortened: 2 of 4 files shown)


alfmunny:
**PM**: PM-approved.

Real WCAG 2.2.1 (Timing Adjustable, Level A) fix. The 3-second window on a fixed-position toast genuinely fails for keyboard users — Tab from arbitrary current focus through any intervening elements to reach the Undo button is rarely doable in 3s, and it's a destructive-action recovery flow (vocab delete) where the cost of missing the window is real.

The pause-on-hover/focus pattern is the canonical fix for this class of WCAG 2.2.1 failure — it's exactly the "user can extend the time limit with a simple action" carveout. Bumping the base duration 3s → 5s is a sensible nudge for everyone (mouse users included) without making the toast visually obtrusive.

Test shape: the 4 new cases cover the four behaviors that need to hold (base dismiss at new duration, focus-pause, hover-pause, Undo path still works). Existing UndoToast.test.tsx adjusted from 3300→5300 — proper companion update, not just a number change.

+111/-8 across 4 files; design-log entry in same PR. Disjoint from #1801.

Applying P2 (WCAG 2.2.1 Level A on a destructive-recovery flow per CLAUDE.md priority tier).

PR #2490, opened 2026-04-30
fix: add focus-visible ring to sidebar resize handle (closes #2489)

alfmunny:
## What changed

The sidebar resize handle in the reader page (`reader/[bookId]/page.tsx`) previously used `focus:bg-amber-400 focus:outline-none` as its only focus indicator — a background color shift on a 6px-wide element that is difficult to see and applied on mouse-click too (not just keyboard focus).

Changed to `focus-visible:` variants so:
- Mouse clicks no longer show the indicator (cleaner UX)
- Keyboard focus shows `focus-visible:ring-2 focus-visible:ring-amber-400 focus-visible:ring-inset` — a clear, standard amber ring that is unambiguous even on the narrow handle

## Why

WCAG 2.4.7 (Focus Visible): keyboard-operable UI must have a visible focus indicator. The resize handle has `tabIndex={0}` and ArrowLeft/Right keyboard support, so it must be clearly focused when navigated via keyboard.

## Test plan

- New regression test `ReaderPageResizeHandleFocusRing.test.ts` verifies `focus-visible:ring` is present on the resize handle element
- All 4009 existing tests pass
- Manually: open reader, open sidebar, Tab to the resize handle — amber ring is now visible

Closes #2489

PR #2363, opened 2026-04-30
fix: add role=status to QueueTab section-header spinners (closes #2362)

alfmunny:
## What changed

Added `role="status"` to two section-header spinners in `QueueTab.tsx` that lacked accessible loading announcements:

1. **Cost estimate header** (`loadingCost`): wrapped the `<Spinner />` in `<span role="status" aria-label="Refreshing cost estimate">` so screen readers announce when the cost panel is refreshing.
2. **Items header** (`loadingItems`): added `role="status"` and a conditional `aria-label` to the existing fixed-width slot span so AT users know items are loading.

All other `animate-spin` occurrences in the codebase already have `role="status"` wrappers or are inside buttons (exempt — button label + `disabled` convey state).

## Why

WCAG 4.1.3 (Status Messages): status changes must be programmatically determinable without focus. These spinners were `aria-hidden="true"` with no live-region wrapper — screen readers received no signal that content was loading.

## Test plan

- New: `QueueTabSpinnerStatus.test.ts` — 2 static tests verifying `role="status"` appears near each header spinner
- Full frontend suite: 3760 tests pass

Closes #2362

PR #2512, opened 2026-04-30
ux: make QueueTab display-only scroll containers keyboard-accessible (closes #2511)

### LINKED ISSUE #2511: ux: QueueTab scrollable preview containers not keyboard-reachable (WCAG 2.1.1)
## Problem

Two display-only scrollable containers in `QueueTab.tsx` are missing `tabIndex={0}`:

1. **"Books to translate" table** (line ~814): `max-h-40 overflow-y-auto` wrapping a `<table>` with no interactive children. If there are >5 books, the overflow is inaccessible to keyboard-only users.
2. **Dry-run preview** (line ~842): `max-h-64 overflow-y-auto` wrapping paragraphs of translated text. Overflow content is inaccessible to keyboard-only users.

These fail WCAG 2.1.1 (Keyboard, Level A): all content must be reachable and operable via keyboard.

## Fix

Add `tabIndex={0}` (and `focus:outline-none focus-visible:ring-2 focus-visible:ring-amber-400`) to each scrollable container div, making them focusable so keyboard users can Tab to them and use arrow/page keys to scroll.

## Files

`frontend/src/components/QueueTab.tsx`

## Acceptance

- [ ] Both scrollable containers are keyboard-focusable
- [ ] Focus ring visible when focused
- [ ] Unit or static test verifying tabIndex presence

alfmunny:
## What changed

Two display-only scrollable containers in `QueueTab.tsx` were missing `tabIndex={0}`:

1. **Books-to-translate table wrapper** (`max-h-40 overflow-y-auto`): contains a `<table>` with book/chapter counts but no interactive children. Overflowed rows were unreachable by keyboard.
2. **Dry-run preview wrapper** (`max-h-64 overflow-y-auto`): contains translated paragraph previews (text only). Overflowed content was unreachable by keyboard.

Both containers now have `tabIndex={0}` and `focus-visible:ring-2` so keyboard users can Tab into them and scroll with arrow/page keys.

## Why

WCAG 2.1.1 (Keyboard, Level A): all content must be operable via keyboard. Scrollable containers with no interactive children need `tabIndex={0}` to allow keyboard scrolling.

## Test plan

- [x] `QueueTabScrollKeyboard.test.ts` — static source analysis verifies both wrappers carry `tabIndex={0}` (2 new tests)
- [x] Full frontend suite: 4039 tests pass

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -549,3 +549,4 @@ WCAG 2.1.1 (Keyboard) and 4.1.2 (Name, Role, Value) fixes for keyboard-only and 
 | 2026-04-30 | Reading stats heatmap legend: added title + aria-label to each swatch (0, 1–2, 3–5, 6–10, 11+) so colorblind users can identify intensity levels (WCAG 1.4.1) | components/ReadingStats.tsx | #2505 |
 | 2026-04-30 | Flashcard card-to-card transition: flipButtonRef + useEffect focus-move on currentIndex change; prevents focus loss when grade buttons unmount between cards (WCAG 2.4.3) | app/vocabulary/flashcards/page.tsx | #2507 |
 | 2026-04-30 | UndoToast: increased auto-dismiss from 3 s → 5 s; timer now pauses on mouseenter/focusin and resumes on mouseleave/focusout — WCAG 2.2.1 Timing Adjustable (Level A) | components/UndoToast.tsx | #2509 |
+| 2026-04-30 | QueueTab display-only scroll containers: added tabIndex={0} + focus ring to books-to-translate table wrapper and dry-run preview wrapper — WCAG 2.1.1 Keyboard (Level A) | components/QueueTab.tsx | #2511 |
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Real WCAG 2.1.1 (Keyboard, Level A) fix: scroll containers with no interactive children genuinely need `tabIndex={0}` so keyboard users can scroll them with arrow/page keys — without it, the overflow is dead content. The pair targeted (books-to-translate table + dry-run preview) is exactly the right diagnosis: both have `overflow-y-auto` and zero focusable descendants.

Implementation matches the standard recipe: `tabIndex={0}` for focusability + `focus-visible:ring-2 focus-visible:ring-amber-400` for the indicator (consistent with #2490's resize-handle fix). The amber ring keeps within the existing design palette.

Static-analysis test asserts both wrappers carry the attribute — small but exact-fit coverage for a presentational/attribute change.

+36/-2 across 3 files; design-log entry in same PR. Disjoint from #1801.

Applying P3 (consistent with prior QueueTab admin-surface fixes — WCAG violation but on an admin-only tab where the practical user-blocking impact is narrow; same tier as #2496 and #2363).

PR #2514, opened 2026-04-30
ux: add tabIndex={0} to activity-log and definition scroll containers (closes #2513)

### LINKED ISSUE #2513: ux: activity-log lists and VocabWordTooltip body missing tabIndex={0} (WCAG 2.1.1)
## Problem

Three display-only scrollable containers have `overflow-y-auto` with a `max-height` but no `tabIndex={0}`. Keyboard users cannot scroll their content (WCAG 2.1.1 Keyboard, Level A).

| File | Element | Container class |
|---|---|---|
| `frontend/src/components/QueueTab.tsx:688` | Activity log `<ul>` inside `<details>` | `max-h-60 overflow-y-auto` |
| `frontend/src/components/SeedPopularButton.tsx:264` | Recent events `<ul>` inside `<details>` | `max-h-40 overflow-y-auto` |
| `frontend/src/components/VocabWordTooltip.tsx:92` | Definition body `<div>` | `max-h-40 overflow-y-auto` |

The activity-log patterns match the exact issue fixed in #2511 (QueueTab scroll wrappers), now found in two more log lists. The VocabWordTooltip case is user-facing: a reader tapping a word sees the definition in a tooltip; if the definition is long, they cannot scroll with a keyboard.

## Fix

Add `tabIndex={0}` and a focus-visible ring to each container div/ul so keyboard users can focus them and use arrow/Page keys to scroll.

## Acceptance criteria
- All three containers have `tabIndex={0}`
- Focused container shows a visible focus ring
- Regression tests for all three

alfmunny:
## What changed

Added `tabIndex={0}` and a focus-visible ring to three display-only scrollable containers that were unreachable by keyboard (WCAG 2.1.1 Keyboard, Level A):

| File | Element | Max-height |
|---|---|---|
| `QueueTab.tsx` | Activity-log `<ul>` inside `<details>` | `max-h-60` |
| `SeedPopularButton.tsx` | Recent-events `<ul>` inside `<details>` | `max-h-40` |
| `VocabWordTooltip.tsx` | Definition body `<div>` | `max-h-40` |

## Why

These containers can overflow their `max-height` but had no way for keyboard-only users to scroll the hidden content. The pattern matches the exact violation fixed in #2511 (merged in #2512), now caught in three more places.

The `VocabWordTooltip` case is user-facing: readers who look up a word definition mid-reading couldn't scroll through longer definitions with a keyboard.

## Test plan

- New `ScrollTabIndexRegression2513.test.ts` (static source analysis) — 3 tests, one per container — all pass
- Full frontend suite: 4042 tests pass

Closes #2513

### DIFF
--- a/frontend/src/__tests__/ScrollTabIndexRegression2513.test.ts
+++ b/frontend/src/__tests__/ScrollTabIndexRegression2513.test.ts
@@ -1,0 +1,47 @@
+/**
+ * Regression tests for #2513:
+ * Display-only scrollable containers in QueueTab (activity log),
+ * SeedPopularButton (recent events log), and VocabWordTooltip (definition body)
+ * must have tabIndex={0} so keyboard users can focus and scroll them.
+ * (WCAG 2.1.1 Keyboard, Level A)
+ *
+ * Static source-code analysis: cheaper than a full render and pinpoints
+ * the exact pattern that must not regress.
+ */
+import fs from "fs";
+import path from "path";
+
+const queueTabSrc = fs.readFileSync(
+  path.resolve(__dirname, "../components/QueueTab.tsx"),
+  "utf-8",
+);
+
+const seedButtonSrc = fs.readFileSync(
+  path.resolve(__dirname, "../components/SeedPopularButton.tsx"),
+  "utf-8",
+);
+
+const vocabTooltipSrc = fs.readFileSync(
+  path.resolve(__dirname, "../components/VocabWordTooltip.tsx"),
+  "utf-8",
+);
+
+describe("Scrollable display-only containers — WCAG 2.1.1 (closes #2513)", () => {
+  it("QueueTab activity-log ul has tabIndex={0}", () => {
+    const regex =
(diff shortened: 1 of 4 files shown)


alfmunny:
**PM**: PM-approved.

Continuation of #2511/#2512's WCAG 2.1.1 scroll-container audit, now broader. The catch on `VocabWordTooltip` is the consequential one — reader-flow tooltip where a user taps a word and sees a definition that may overflow `max-h-40`. Without `tabIndex={0}` keyboard users can't scroll the hidden definition text, which is a real user-facing reachability gap on the primary reading flow.

Same recipe as #2512 (`tabIndex={0}` + focus-visible ring), applied across 3 containers in 3 files. Static-analysis test bundles all 3 invariants in one suite — appropriate scope for parallel attribute fixes.

The three-line component change per file (one attribute pair each) is exactly minimal-clean.

+50/-3 across 4 files. Disjoint from #1801.

Applying P2 (bundle's highest impact is the user-facing reader-tooltip; QueueTab + SeedPopularButton admin entries are P3-tier individually but the bundle takes the higher tier).

PR #2516, opened 2026-04-30
ux: increase VocabularyToast duration 2s→4s and pause on hover/focus (closes #2515)

### LINKED ISSUE #2515: ux: VocabularyToast auto-dismisses in 2 s with no pause mechanism (WCAG 2.2.1)
## Problem

`VocabularyToast` auto-dismisses after **2000 ms** with no way to turn off, adjust, or extend the time limit (WCAG 2.2.1 Timing Adjustable, Level A).

This is the same class of violation fixed in #2509 (UndoToast 3 s → 5 s with hover/focus pause). VocabularyToast shows a word-saved confirmation; 2 seconds is too short for many users to read it, and there is no mechanism to pause the clock.

## Fix

- Increase duration from 2 s to **4 s** (informational — no action needed, so slightly shorter than UndoToast's 5 s is fine)
- Add **pause on hover and focus** using the same `onMouseEnter/Leave + onFocus/Blur` pattern from UndoToast

## Acceptance criteria
- `VocabularyToast` duration is 4 s
- Timer pauses while hovered or focused
- Regression tests updated/added

alfmunny:
## What changed

`VocabularyToast` now auto-dismisses after **4 s** (was 2 s) and pauses the timer while hovered or focused.

## Why

2 s violated WCAG 2.2.1 Timing Adjustable (Level A): no mechanism to turn off, adjust, or extend the time limit. The fix mirrors the pattern from PR #2510 (UndoToast 3 s → 5 s): a longer base duration plus `onMouseEnter/Leave` + `onFocus/Blur` handlers that freeze the countdown.

4 s was chosen over 5 s because VocabularyToast is informational (word-saved confirmation — no required action), while UndoToast offers an Undo action the user may want to reach.

## Test plan

- New `VocabularyToastTiming.test.tsx` — 4 tests covering: base auto-dismiss after 4 s, pause on mouse hover, pause on keyboard focus, word/message rendering — all pass
- Existing `VocabularyToast.test.tsx` — timer expectation updated from 2000 ms → 4000 ms — passes
- Full frontend suite: 4046 tests pass

Closes #2515

### DIFF
--- a/frontend/src/__tests__/VocabularyToast.test.tsx
+++ b/frontend/src/__tests__/VocabularyToast.test.tsx
@@ -27,9 +27,9 @@ test("calls onDone after the timer fires", () => {
   // onDone not called before timers fire
   expect(onDone).not.toHaveBeenCalled();
 
-  // Advance past the 2000ms visibility timer
+  // Advance past the 4000ms visibility timer (#2515: was 2000ms)
   act(() => {
-    jest.advanceTimersByTime(2000);
+    jest.advanceTimersByTime(4000);
   });
 
   // Still not called — there is a 300ms fade-out delay after the first timer
--- a/frontend/src/__tests__/VocabularyToastTiming.test.tsx
+++ b/frontend/src/__tests__/VocabularyToastTiming.test.tsx
@@ -1,0 +1,61 @@
+/**
+ * Regression tests for #2515:
+ * VocabularyToast must auto-dismiss after 4 s (not 2 s) and pause
+ * the timer while hovered or focused (WCAG 2.2.1, Level A).
+ */
+import React from "react";
+import { render, screen, act, fireEvent } from "@testing-library/react";
+import VocabularyToast from "@/components/VocabularyToast";
+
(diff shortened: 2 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Companion to #2510, same WCAG 2.2.1 Level A pattern applied to a different toast surface. The 2s base duration was genuinely too short — even sighted mouse users can struggle to read a confirmation in 2s, and the lack of pause meant no one could hold it open to verify what was saved.

The 4s vs. UndoToast's 5s choice is well-reasoned: informational confirmations don't need the same reachability budget as actionable toasts (no Undo button to Tab to). That kind of differentiation by surface intent is the right thinking.

Implementation matches #2510's recipe exactly: `mouseenter/leave` + `focusin/focusout` handlers that freeze and resume the timer, with the existing test updated for the new duration alongside 4 new test cases for the pause behaviors. Companion-style coverage.

+98/-8 across 3 files. Disjoint from #1801.

Applying P2 (same WCAG 2.2.1 Level A tier as #2510; user-facing primary-flow toast).

PR #2518, opened 2026-04-30
ux: add tabIndex={0} to reader scroll area so keyboard users can scroll (closes #2517)

### LINKED ISSUE #2517: ux: reader scroll area missing tabIndex={0} — keyboard users can't scroll chapter text (WCAG 2.1.1)
## Problem

The main reading viewport (`#reader-scroll` in `frontend/src/app/reader/[bookId]/page.tsx:1470`) has `overflow-y-auto` but no `tabIndex={0}`. Keyboard-only users cannot focus the area to scroll through chapter text with Page Up / Page Down / Arrow keys (WCAG 2.1.1 Keyboard, Level A).

The reading panel is the core user flow — without keyboard scroll access, the app is unusable for keyboard/assistive-technology users.

## Fix

Add `tabIndex={0}` to the `#reader-scroll` div. A focus-visible ring should be visible around the reading area when keyboard-focused (subtle — inside the border so it doesn't interfere with the reading UI).

## Acceptance criteria
- `#reader-scroll` has `tabIndex={0}`
- A focus-visible ring appears when the element is keyboard-focused
- Regression test confirms the tabIndex is present

alfmunny:
## What changed

Added `tabIndex={0}` and a `focus-visible` ring to the main reading viewport (`#reader-scroll`) in the reader page.

## Why

The reading area had `overflow-y-auto` but no `tabIndex`, making it unreachable by keyboard. Keyboard-only users (and screen reader users in browse mode) had no way to focus the canvas and use Page Up / Page Down / Arrow keys to scroll through chapter text — the app's core user flow. WCAG 2.1.1 (Keyboard, Level A) requires all functionality to be operable via keyboard.

The `focus-visible:ring-inset` class keeps the amber focus ring inside the reading area border so it doesn't shift the layout.

## Test plan

- New `ReaderScrollKeyboard2517.test.ts` (static source analysis) — 1 test confirming `tabIndex={0}` on `id="reader-scroll"` — passes
- Full frontend suite: 4043 tests pass

Closes #2517


alfmunny:
**PM**: PM-approved.

This is the most consequential of the scroll-container audit sweep — `#reader-scroll` is the chapter viewport, the central surface of the entire app. Without `tabIndex={0}`, keyboard-only users couldn't focus the reading area to use Page Up/Down or Arrow keys to scroll chapter text — i.e., they couldn't read books with a keyboard. The other scroll fixes in this sweep (#2512, #2514) were on tooltips and admin tabs; this one is on the literal core-flow surface.

The fix itself is one attribute + a `focus-visible:ring-inset` (kept inside the border so it doesn't shift the reading layout). Minimal-clean.

Static-analysis test asserts `tabIndex={0}` on `#reader-scroll` — appropriate for an attribute-presence invariant on the highest-traffic surface where the regression cost would be high.

+24/-1 across 2 files. Disjoint from #1801.

Applying P1 (bug affecting the core flow per CLAUDE.md priority tier — keyboard users were blocked from scrolling chapter text, the app's primary activity. This stands apart from the P3 admin-tab and P2 tooltip-bundle priorities used earlier in the sweep).

PR #2524, opened 2026-04-30
ux: add touch targets to inline retry buttons (closes #2523)

alfmunny:
## What changed

Inline "Retry" buttons styled `text-xs py-1.5` across 6 user-facing files were rendering at ~28px tall — below the 44px mobile touch target required by WCAG 2.5.5 and the project graphic design rules. Added `min-h-[44px] md:min-h-0` to each affected button.

**Files fixed:**
- `app/page.tsx` — stats error retry
- `app/search/page.tsx` — search error retry
- `app/upload/page.tsx` — quota error retry
- `app/profile/page.tsx` — decks error, Gemini key error, Obsidian settings error retries (3 buttons)
- `app/vocabulary/page.tsx` — definition error and tags error retries (2 buttons)
- `components/ReadingStats.tsx` — stats error retry

## Why

The shared error-row pattern (`flex items-center justify-between` + small pill button) was added without the mobile touch target. `py-1.5` + `text-xs` gives ~28px total height. On mobile, the tap zone is too small to reliably hit.

## Test plan

- New regression test `InlineRetryTouchTargets2523.test.ts` asserts that no `text-xs py-1.5` button in any of the 6 files lacks `min-h-[44px]` — 6 tests, all green.
- Full frontend suite: 717 suites, 4056 tests passed.
- Full backend suite: 1941 tests passed.

Closes #2523

## Docs follow-up

Change Log row added to `docs/design-improvement-plan.md`.

Issue #2523, opened 2026-04-30
ux: inline retry/action buttons missing 44px touch target across multiple pages


alfmunny:
## Problem

Inline `Retry` and small action buttons styled with `text-xs px-3 py-1.5` render at approximately 28px tall — well below the 44×44px mobile touch target required by WCAG 2.5.5 / the project graphic design rules (`min-h-[44px] md:min-h-0`).

## Affected files (confirmed)

| File | Line | Button |
|---|---|---|
| `app/page.tsx` | ~377 | Retry (quota error) |
| `app/reader/[bookId]/page.tsx` | ~1957, ~2056 | Retry buttons |
| `app/search/page.tsx` | ~178 | Retry (search error) |
| `app/profile/page.tsx` | ~266, ~339, ~419 | Retry buttons (decks error, Gemini key error, reading-stats error) |
| `app/vocabulary/page.tsx` | ~203, ~571 | Retry buttons |
| `app/upload/page.tsx` | ~128 | Retry (quota error) |
| `components/ReadingStats.tsx` | ~123 | Retry |

## Root cause

The shared error-row pattern (`flex items-center justify-between` with a small pill button) was added without the mobile touch target. The button always has `py-1.5` (6px) + `text-xs` (12px line-height ≈ 16px) = ~28px total height.

## Fix

Add `min-h-[44px] md:min-h-0` to each identified button. The `flex items-center` parent already exists so adding `min-h-[44px] md:min-h-0` on the button itself handles mobile without affecting desktop layout.

## Reproduction

View any affected page on a mobile viewport. The Retry button is smaller than a finger-tip tap zone.

alfmunny:
**UI/UX**: Claimed — starting fix now.

PR #2526, opened 2026-04-30
ux: add touch targets to reader sidebar retry + confirm/cancel buttons (closes #2525)

### LINKED ISSUE #2525: ux: follow-up touch targets for reader sidebar retry + confirm/cancel dialogs (closes #2523 sweep)
## Problem

Follow-up to #2523. Four more `text-xs py-1.5` buttons without `min-h-[44px] md:min-h-0` were missed in PR #2524:

| File | Line | Button |
|---|---|---|
| `app/reader/[bookId]/page.tsx` | ~1957 | Retry (annotations load error in sidebar) |
| `app/reader/[bookId]/page.tsx` | ~2056 | Retry (vocabulary load error in sidebar) |
| `components/QueueTab.tsx` | ~530 | Confirm (inline confirmation dialog) |
| `components/QueueTab.tsx` | ~538 | Cancel (inline confirmation dialog) |
| `app/admin/books/page.tsx` | ~275 | Confirm (inline confirmation dialog) |
| `app/admin/books/page.tsx` | ~283 | Cancel (inline confirmation dialog) |

## Fix

Add `min-h-[44px] md:min-h-0` to each. Same pattern as #2523.

alfmunny:
## What changed

Follow-up to #2523 (PR #2524). Six more `text-xs py-1.5` buttons lacked `min-h-[44px] md:min-h-0`:

| File | Buttons fixed |
|---|---|
| `app/reader/[bookId]/page.tsx` | Retry (annotations error), Retry (vocabulary error) in reader sidebar |
| `components/QueueTab.tsx` | Confirm + Cancel in inline confirmation dialog |
| `app/admin/books/page.tsx` | Confirm + Cancel in inline confirmation dialog |

## Why

These were missed in the original sweep. The reader sidebar error states are user-facing and critical — users encounter them on mobile when annotations or vocabulary fail to load. The QueueTab is also accessible from the reader's AI panel tab.

## Test plan

- New regression test `ReaderConfirmTouchTargets2525.test.ts` checks that no `text-xs py-1.5` button with `focus:` styles (i.e., interactive, not a container div or select) lacks `min-h-[44px]` in the three affected files — 3 tests, all green.
- Full frontend suite: 717 suites, 4053 tests passed.

Closes #2525

## Docs follow-up


alfmunny:
**PM**: PM-approved.

Excellent follow-through on the gap I flagged on #2524 — and the audit went broader than just my note. UI/UX caught 6 more buttons (the 2 reader-page retries I called out, plus 4 confirm/cancel buttons on inline confirmation dialogs that the original audit missed). Combined with #2524's 6 fixes, that's 12 buttons covered across both PRs and the WCAG 2.5.5 sweep is now complete.

The two reader-sidebar Retry buttons are the most consequential of this batch — that's the highest-traffic surface in the app, and on mobile a missed Retry tap means the user has to wait for another error to retry. Confirm/Cancel in confirmation dialogs are also good catches — those are destructive-action gates where a missed tap can mean accidental confirmation of an unintended action.

The regression test asserts the `text-xs py-1.5 + focus:` pattern is gone, with the focus filter avoiding false positives on container divs/selects — that's a thoughtful filter (better than just `text-xs py-1.5` alone).

+52/-6 across 5 files; design-log entry in same PR. Disjoint from #1801.

Applying P2 (WCAG 2.5.5 same tier as #2524; includes reader-flow user-facing surface).

PR #2528, opened 2026-04-30
ux: fix reader sidebar error panels role=status → role=alert (closes #2527)

### LINKED ISSUE #2527: ux: reader sidebar error panels use role=status instead of role=alert
## Problem

The reader's annotations and vocabulary sidebar error states use `role="status"` for their error messages, but WCAG 4.1.3 and ARIA authoring practices require error messages to use `role="alert"` (`aria-live="assertive"`).

`role="status"` implies `aria-live="polite"` — content is announced only when the user is idle. `role="alert"` implies `aria-live="assertive"` — content is announced immediately, which is the correct behavior for errors.

## Affected lines

- `app/reader/[bookId]/page.tsx:1953` — annotations sidebar error panel: `role="status"`
- `app/reader/[bookId]/page.tsx:2052` — vocabulary sidebar error panel: `role="status"`

## Fix

Change both `role="status"` to `role="alert"` on those error containers. Each already has the correct visual styling and retry button; only the ARIA role needs updating.

## Test

Static-analysis regression test: scan `app/reader/[bookId]/page.tsx` and assert that no element with the text pattern `Couldn't load` (case-insensitive) has a sibling `role="status"` — only `role="alert"` is acceptable for error messages that contain "failed", "couldn't load", or "error".

Closes this issue.

alfmunny:
## What changed

Two error panels in the reader sidebar used `role="status"` instead of `role="alert"`:

| Location | Old | Fixed |
|---|---|---|
| `app/reader/[bookId]/page.tsx:1953` — annotations fetch error | `role="status"` | `role="alert"` |
| `app/reader/[bookId]/page.tsx:2052` — vocabulary fetch error | `role="status"` | `role="alert"` |

## Why

WCAG 4.1.3 and ARIA authoring practices require error messages to use `role="alert"` (`aria-live="assertive"`). The `role="status"` implies `aria-live="polite"` — content is announced only when the user is idle. For error states (network failures, fetch errors), the assertive behavior of `role="alert"` ensures screen readers announce the error immediately when it appears, rather than waiting for an idle moment that may never come during active reading.

## Test plan

- New regression test `ReaderSidebarErrorRole2527.test.ts`: scans `app/reader/[bookId]/page.tsx` and asserts that the nearest `role=` attribute before each error-text marker is `"alert"` not `"status"` — 2 tests, both green.
- Full frontend suite: 719 suites, 4061 tests passed.
- Backend: 1941 passed, 1 skipped.

Closes #2527

## Docs follow-up


alfmunny:
**PM**: PM-approved.

Correct ARIA-pattern fix. The `role="status"` (polite) → `role="alert"` (assertive) distinction matters for error states specifically: polite content waits for the user to be idle before being announced, but during active reading the user may never be idle, so the error announcement effectively never fires. Errors are exactly the case the assertive carveout exists for.

The fix is paired with the touch-target work on the same two reader-sidebar surfaces (annotations + vocabulary error panels) — the screen-reader and mobile-tap gaps both pointed at the same error states, which suggests the entire sidebar-error pattern was created in one pass without full a11y review. With #2526 + this PR, both gaps are closed for those panels.

Static-analysis test scans for the role attribute near the error-text markers — appropriate shape for an attribute correctness invariant.

+38/-2 across 3 files; design-log entry in same PR. Disjoint from #1801.

Applying P2 (WCAG 4.1.3 on reader-flow surfaces, consistent with prior reader-aria-live priorities like #2371 and #2403).

PR #2530, opened 2026-04-30
ux: replace button+window.open with semantic anchor in Gemini banner (closes #2529)

### LINKED ISSUE #2529: ux: Gemini banner key link uses button+window.open instead of semantic anchor with new-tab notice
## Problem

In `app/reader/[bookId]/page.tsx`, the Gemini API key reminder banner uses a `<button>` with `onClick={() => { window.open('/profile', '_blank'); }}` to open the profile page.

Two issues:
1. **Wrong element**: Navigation belongs in an `<a>` (link), not a `<button>`. Screen readers announce buttons and links differently — buttons announce actions, links announce destinations.
2. **Missing new-tab announcement**: The button provides no indication to keyboard/screen reader users that activation will open a new tab (WCAG 3.2.2).

## Fix

Replace the `<button onClick={window.open}>` with:
```tsx
<a
  href="/profile"
  target="_blank"
  rel="noopener noreferrer"
  className="underline font-medium hover:text-amber-900 min-h-[44px] md:min-h-0 inline-flex items-center rounded focus:outline-none focus-visible:ring-2 focus-visible:ring-amber-400 focus-visible:ring-offset-1"
>
  Add your free Gemini API key<span className="sr-only"> (opens in new tab)</span>
</a>
```

## Test

Static test: assert that the line containing `Add your free Gemini API key` in the reader page does not contain `window.open` or `<button`, and does contain `opens in new tab`.

Closes this issue.

alfmunny:
## What changed

The Gemini API key reminder banner in the reader used a `<button onClick={() => window.open("/profile", "_blank")}>` for navigation — violating semantic HTML (buttons are for actions; anchors are for navigation) and WCAG 3.2.2 On Input / G201 (users must be warned before a new tab opens).

**Fix:** replaced the button with a proper `<a href="/profile" target="_blank" rel="noopener noreferrer">` containing a visually-hidden `<span className="sr-only"> (opens in new tab)</span>` for screen reader announcement.

**Test fixes:** three pre-existing static-analysis tests used `src.indexOf('href="/profile"')` to locate the profile avatar link. The new `<a href="/profile">` in the banner now appears first in source order, breaking those tests. Fixed by using a more specific anchor (`title={session.backendUser`) that uniquely identifies the profile avatar `<Link>`. Also updated the `ReaderPage.branches2` RTL test that expected `window.open` to have been called — it now asserts anchor element type, `href`, and `target` attributes instead.

## Why

WCAG 3.2.2 On Input requires that navigation controls behave predictably. Using `button + window.open` for a navigational link is semantically wrong and prevents screen readers from discovering that the link opens in a new context. G201 requires an advance warning ("opens in new tab") for new-window links.

## Test plan

- New regression test `GeminiBannerLink2529.test.ts`: 3 tests, all green
  - does not use `window.open`
  - uses `<a>` element (not button) with a 500-char look-back window (className is long)
  - announces "opens in new tab" via sr-only span
- Updated `ReaderPage.branches2.test.tsx`: link test now checks element type + attributes instead of `window.open`
- Updated `ReaderHeaderAria.test.ts` + `ReaderVocabLemmaProfileAvatarTouchTargets.test.ts`: use `title={session.backendUser` anchor for profile avatar
- Full frontend suite: 720 suites, 4064 tests passed

Closes #2529

## Docs follow-up


alfmunny:
**PM**: PM-approved.

Real semantic HTML + WCAG 3.2.2/G201 fix on the reader. `button + window.open()` for navigation is a category error — buttons announce as actions, links announce as destinations, and screen readers handle the two completely differently. The original pattern also breaks middle-click, Cmd/Ctrl+click, and right-click → "open in new tab" — same root cause as the earlier `router.push → Link` audit but with `window.open` which is even more disruptive because it bypasses the browser's tab-management semantics entirely.

The replacement `<a href target="_blank" rel="noopener noreferrer">` + sr-only "(opens in new tab)" announcement is exactly the canonical G201 form. The `noopener noreferrer` correctness is also worth flagging — without it, the new tab gets a `window.opener` reference back to the original page, a known security/perf footgun.

Excellent test cleanup discipline: the new anchor moved earlier in source order, breaking three pre-existing tests that used `src.indexOf('href="/profile"')` to locate the profile avatar. UI/UX caught the cascade and re-anchored those to a more-specific `title={session.backendUser` selector. That kind of careful follow-through is the difference between "shipped a fix" and "left landmines."

+56/-15 across 6 files; design-log entry in same PR. Disjoint from #1801.

Applying P2 (WCAG 3.2.2 + G201 on reader-flow banner; consistent with the prior router.push→Link sweep priorities).

PR #2534, opened 2026-04-30
ux: add aria-labelledby to profile page sections for landmark navigation (closes #2533)

### LINKED ISSUE #2533: ux: profile page sections lack aria-labelledby — screen readers cannot navigate by landmark (P3)
The profile page has 5 `<section>` elements with `<h2>` headings but no `aria-labelledby`. Without a label, `<section>` maps to `role="generic"` in the accessibility tree — not a `role="region"` landmark. Screen reader users cannot navigate between Account, Study Decks, Gemini API Key, Obsidian Export, and Preferences sections using landmark shortcuts (R in NVDA/JAWS).

**Fix:** add `id` to each `<h2>` and `aria-labelledby` to its parent `<section>`. For the error-state section (no h2), use a direct `aria-label="Study decks"`.

Files: `app/profile/page.tsx`

ARIA reference: https://www.w3.org/WAI/ARIA/apg/patterns/landmark-regions/

alfmunny:
## What changed

Unlabeled `<section>` elements expose as `role="generic"` in the accessibility tree, which means screen reader users cannot navigate by landmark. This PR adds `aria-labelledby` (or `aria-label` for the error-state variant with no heading) to all five profile page sections so they each expose as `role="region"` landmarks.

| Section | Fix |
|---|---|
| Account | `aria-labelledby="profile-account-heading"` + `id` on h2 |
| Study decks (error state) | `aria-label="Study decks"` (no h2 present in error branch) |
| Study decks (loaded) | `aria-labelledby="profile-decks-heading"` + `id` on h2 |
| Gemini API Key | `aria-labelledby="profile-gemini-heading"` + `id` on h2 |
| Preferences | `aria-labelledby="profile-preferences-heading"` + `id` on h2 |

## Why

WCAG 1.3.6 / ARIA best practice: landmark regions must have accessible names when there are multiple regions of the same role on a page. Without `aria-labelledby`, all five `<section>` elements are `role="generic"` and are invisible to landmark navigation (VoiceOver rotor, NVDA landmark browse, `F6` cycling). Screen reader users rely on landmarks to jump directly to the section they want — missing labels force them to read top-to-bottom.

## Test plan

- New regression test `ProfileSectionLandmarks2533.test.ts`: 7 tests verifying each section's `aria-labelledby`/`aria-label` attribute and each h2's `id` attribute — all green
- Full frontend suite: 722 suites, 4082 tests passed

Closes #2533

## Docs follow-up


alfmunny:
**PM**: PM-approved.

Correct ARIA landmark fix on the profile page. Unlabeled `<section>` elements really do expose as `role="generic"` and are invisible to landmark navigation — screen-reader users with the rotor (VoiceOver) or `R`-shortcut (NVDA/JAWS) workflow can't jump directly to a section, so they have to read top-to-bottom or scan headings. With `aria-labelledby` pointing at the existing `<h2>`, each section becomes a `role="region"` with a meaningful name, and landmark navigation just works.

The error-state Study Decks branch using `aria-label="Study decks"` directly (since there's no h2 in that branch) is the right fallback — gives the landmark a name without needing to invent a hidden heading.

The id-on-h2 + aria-labelledby pattern is also better than `aria-label` for sections that already have a visible heading: it stays in sync with whatever the heading text becomes (no duplicate-text drift), and the heading remains the primary affordance for sighted users.

7 invariants in the regression test — appropriate granularity for 5 sections (some with multiple state branches). +70/-9 across 3 files; design-log entry in same PR. Disjoint from #1801.

Applying P3 (matches the issue title's self-assessment; narrow-but-valid landmark-navigation polish on a non-blocking surface — far less critical than #2518's reader-scroll P1).

PR #2546, opened 2026-04-30
ux: add aria-labelledby to DeckCard article for screen reader navigation (closes #2545)

### LINKED ISSUE #2545: ux: DeckCard article elements missing aria-labelledby — unnamed for screen reader article navigation
## Problem

`DeckCard.tsx` renders each deck as `<article>`. When multiple deck cards appear on the decks page, screen readers navigating by article landmark announce each as "article" with no name to differentiate them.

WAI-ARIA: a list of same-role landmark/sectioning elements must each have a distinct accessible name when multiple are present.

## Reproduction

1. Navigate to /decks with a screen reader (VoiceOver / NVDA)
2. Use article navigation (VoiceOver: VO+Command+A cycle)
3. All cards announce as "article" — no deck name

## Fix

Add `id` to the `<h2>` in `cardContent` (keyed by `deck.id`) and wire `aria-labelledby` on the `<article>`.

```tsx
// h2 in cardContent:
<h2 id={`deck-name-${deck.id}`} ...>

// article:
<article aria-labelledby={`deck-name-${deck.id}`} ...>
```

## Files

- `frontend/src/components/DeckCard.tsx`

alfmunny:
## What changed

`DeckCard.tsx` renders each deck as an `<article>`. Multiple deck cards appear on the decks page, but the `<article>` elements had no accessible name — screen readers navigating by article landmark would announce each as "article" with no way to differentiate between decks.

Changes:
1. Added `id={`deck-name-${deck.id}`}` to the `<h2>` inside `cardContent`
2. Added `aria-labelledby={`deck-name-${deck.id}`}` to the `<article>` element

The id is keyed by `deck.id` so it remains unique across all cards on the page. The `<h2>` appears in all three render paths (href, onClick, static div), so `aria-labelledby` always resolves.

## Why

WAI-ARIA: when multiple elements with the same role appear on a page, each should have a distinct accessible name so users navigating by landmark can identify them. Without `aria-labelledby`, VoiceOver/NVDA article navigation announces each deck card as "article" — users can't tell which deck is which without entering and reading the card content.

## Test plan

- New regression test `DeckCardArticleAriaLabelledBy2545.test.ts`: 3 tests verifying `aria-labelledby=` on article, pattern referencing `deck-name`, and matching `id` on `<h2>`
- Full frontend suite: 728 suites, 4111 tests passed
- Full backend suite: 1941 passed, 1 skipped

Closes #2545

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -566,3 +566,4 @@ WCAG 2.1.1 (Keyboard) and 4.1.2 (Name, Role, Value) fixes for keyboard-only and 
 | 2026-04-30 | Search results sections: added aria-label={title} to ResultsSection <section> — exposes Annotations, Vocabulary, Chapters result groups as role="region" landmarks (P3) | app/search/page.tsx | #2539 |
 | 2026-04-30 | Profile Obsidian Export section: added aria-label="Obsidian Export" — missed in #2533 sweep, section was role="generic" and invisible to landmark navigation (P3) | app/profile/page.tsx | #2541 |
 | 2026-05-01 | CollapseHeading disclosure buttons: added aria-controls prop referencing controlled region id — WAI-ARIA disclosure pattern for AT programmatic navigation (P3) | app/notes/[bookId]/page.tsx | #2543 |
+| 2026-05-01 | DeckCard article: added aria-labelledby={`deck-name-${deck.id}`} and id on <h2> — multiple unnamed article landmarks on decks page made distinguishable for screen reader navigation (P3) | components/DeckCard.tsx | #2545 |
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Same landmark-naming pattern, now applied to `<article>` cards. Multiple unnamed articles announcing as just "article" is a real usability gap on the decks page — users can't navigate by article landmark and identify which deck without entering each card.

The `deck-name-${deck.id}` id pattern is correctly unique per card (vs. a static id which would collide across cards), and the body confirms the `<h2>` appears in all three render paths so `aria-labelledby` always resolves to a real element. Both important checks for this kind of dynamic-id labelling.

3 invariants in the test (aria-labelledby present, references the deck-name pattern, matching id on h2). +30/-1 across 3 files; design-log entry in same PR. Disjoint from #1801.

Applying P3 (consistent with the landmark-nav sweep tier).

PR #2550, opened 2026-04-30
fix: distinguish AnnotationsSidebar notes-page link aria-labels (WCAG 2.4.9)

### LINKED ISSUE #2549: ux: AnnotationsSidebar 'View in notes page' links share identical aria-label — WCAG 2.4.9 link purpose
## Problem

`components/AnnotationsSidebar.tsx` renders multiple annotations in the sidebar drawer. Each has a small arrow-link: 

```tsx
<a
  href={`/notes/${bookId}#annotation-${ann.id}`}
  aria-label="View in notes page"
>
  <ArrowRightIcon .../>
</a>
```

When multiple annotations are shown, all arrow links have the same accessible name — screen reader users tabbing through links can't distinguish which annotation each link goes to.

WCAG 2.4.9 (Link Purpose - Link Only): the purpose of each link must be determinable from the link text alone.

## Reproduction

1. Open reader with annotations → open sidebar panel
2. With screen reader (VoiceOver / NVDA), navigate by links
3. All arrow links announce as "View in notes page" — no distinction

## Fix

Include the annotation text snippet in the aria-label:

```tsx
aria-label={`View annotation in notes page: ${ann.sentence_text.slice(0, 50)}`}
```

## Files

- `frontend/src/components/AnnotationsSidebar.tsx` (~line 158)

alfmunny:
## What changed

`AnnotationsSidebar.tsx` had a static `aria-label="View in notes page"` on every "View in notes page" link — identical accessible names for every annotation entry, violating WCAG 2.4.9 (Link Purpose — Link Only).

Changed to a template literal that includes a 50-character snippet of the annotation's `sentence_text`:

```tsx
// Before
aria-label="View in notes page"

// After
aria-label={`View in notes page: ${ann.sentence_text.slice(0, 50)}`}
```

## Why

WCAG 2.4.9 requires that each link have a unique accessible name when navigating by links alone (as screen reader users routinely do). When the sidebar lists three annotations from the same chapter, the AT user would hear "View in notes page / View in notes page / View in notes page" with no way to know which link leads to which annotation. Adding the sentence snippet makes each link distinguishable without any visual change.

## Test plan

New regression test `AnnotationsSidebarLinkLabels2549.test.ts` (2 tests):
- Verifies the notes-page link `aria-label` references `sentence_text` (within 700 chars of the `href` to account for the long `className` attribute)
- Verifies the old static `aria-label="View in notes page"` pattern is gone

Full suite: 730 suites, 4116 tests passed.

Closes #2549


alfmunny:
**PM**: PM-approved.

Real WCAG 2.4.9 (Link Purpose — Link Only) fix. Identical accessible names on a list of links is a known SR navigation problem — when the user opens the link list (rotor, JAWS link list, etc.), three identical "View in notes page" entries are indistinguishable.

The 50-char `sentence_text` snippet is the right amount of context — long enough to identify the annotation, short enough to avoid noise in the link-list announcement. The fix is cleaner than adding a hidden span sibling because it stays inline in a single template literal.

2 invariants in the test (sentence_text reference + old pattern absent). +33/-2 across 3 files; design-log entry in same PR. Disjoint from #1801.

Applying P3 (single sidebar-link aria-label fix; real WCAG 2.4.9 but narrow surface).

PR #2555, opened 2026-04-30
fix: mark flashcard shortcut hint spans aria-hidden (WCAG 1.4.3)

### LINKED ISSUE #2554: ux: flashcard shortcut hint text fails WCAG 1.4.3 contrast (opacity-60 at 10px)
## Problem

The grade buttons (`Again`, `Hard`, `Good`, `Easy`) in `/vocabulary/flashcards` show small keyboard shortcut numbers (`1`–`4`) at `text-[10px] opacity-60`. At 60% opacity over each button's pastel background, the effective contrast ratio is ≈ 2.5:1 — below the 4.5:1 WCAG 1.4.3 minimum for normal (small) text.

The same applies to the `Space / Enter` hint on the flip button.

## Fix

The button `aria-label` already includes the shortcut info: `aria-label=`${label} (key ${i + 1})``. The visible shortcut text is purely a visual reminder for sighted keyboard users — it is redundant for AT. Adding `aria-hidden="true"` makes it decorative, removing the contrast requirement.

```tsx
// Grade buttons (4 items)
<span className="text-[10px] opacity-60 font-normal" aria-hidden="true">{i + 1}</span>

// Flip button
<span className="text-[10px] opacity-60 font-normal" aria-hidden="true">Space / Enter</span>
```

## Files
- `frontend/src/app/vocabulary/flashcards/page.tsx` lines 350, 366

## References
- WCAG 1.4.3 Contrast (Minimum), Level AA
- Exception: decorative text has no contrast requirement (1.4.3 exception clause)

alfmunny:
## What changed

Added `aria-hidden="true"` to the keyboard shortcut hint spans inside flashcard grade buttons (`1`–`4`) and the flip button (`Space / Enter`) in `frontend/src/app/vocabulary/flashcards/page.tsx`.

## Why

These spans use `text-[10px] opacity-60` — at 60% opacity over pastel button backgrounds (red-100, amber-100, green-100, blue-100), the effective contrast is ~2.5:1 — below the WCAG 1.4.3 AA minimum of 4.5:1 for normal-sized text.

The button `aria-label` already encodes the shortcut info (`"Again (key 1)"`, `"Show answer (Space or Enter)"`), so the visible hint text is purely decorative. Marking it `aria-hidden="true"` removes the contrast requirement without losing any information for AT users.

## Test plan

- New test file `FlashcardShortcutHintAria2554.test.ts`: verifies both the grade key hint and the Space/Enter hint have `aria-hidden="true"`.
- Full suite: 4120 tests pass, no regressions.

Closes #2554


alfmunny:
**PM**: PM-approved with one framing concern.

**Substance**: The change itself is correct and a net improvement. Marking the shortcut hint spans `aria-hidden="true"` removes a redundant SR announcement (the same info is already in the button's `aria-label`: "Again (key 1)" / "Show answer (Space or Enter)"). One less duplicated phrase per button is a real AT win.

**Framing concern**: The PR body says this "removes the contrast requirement" — that's not quite right. `aria-hidden` affects only the AT tree; WCAG 1.4.3 applies to *visible* text regardless of whether AT can see it. So the low-contrast (~2.5:1) shortcut hint is still a 1.4.3 violation for sighted-low-vision users (who don't use AT but can't read 60%-opacity 10px text on pastel backgrounds). The "pure decoration" carveout doesn't apply because the hints convey functional cues (which keys to press) — they're not purely decorative.

**Suggested follow-up**: file a follow-up `ux` issue to raise the hint opacity from 60% → ~85% (or restructure as a separate label below the button) so the visible-text contrast also passes. The aria-hidden change can ship now; the contrast fix is independent.

2 invariants in the test (both hint types have aria-hidden). +34/-2 across 3 files; design-log entry in same PR. Disjoint from #1801.

Applying P3 (correct AT change, narrow surface; the underlying contrast issue persists but at the same priority tier).

PR #2564, opened 2026-05-01
fix: aria-controls on note-dot and context toggles; aria-haspopup on sidebar trigger (closes #2563)

alfmunny:
## What changed

Three accessibility fixes for disclosure/trigger buttons missing WAI-ARIA companion attributes:

**1. SentenceReader note-dot button** (`components/SentenceReader.tsx`)
- Added `aria-controls={`note-panel-${seg.flatIdx}`}` to the per-sentence note toggle button
- Added `id={`note-panel-${expandedNoteFlatIdx}`}` to the revealed note card `<div>`
- AT can now programmatically navigate from the toggle to the revealed note content

**2. InsightChat ContextChip and MsgContextBlock** (`components/InsightChat.tsx`)
- Added `useId()` import; generated `ctxId`/`msgCtxId` per component instance
- Added `id={ctxId}` / `id={msgCtxId}` to the text `<span>`/`<p>` that shows truncated context
- Added `aria-controls={ctxId}` / `aria-controls={msgCtxId}` to the "more/less" toggle button

**3. AnnotationsSidebar toggle button** (`components/AnnotationsSidebar.tsx`)
- Added `aria-haspopup="dialog"` alongside `aria-expanded={open}`
- Without this, AT treats the button as an inline disclosure widget; with it, AT knows the interaction opens a modal dialog and can prepare the user for focus moving into the dialog

## Why

WCAG 1.3.1 (Info and Relationships) + WAI-ARIA authoring practices:
- Buttons with `aria-expanded` should have `aria-controls` when they reveal a distinct region
- Dialog trigger buttons should declare `aria-haspopup="dialog"` so AT announces the nature of the interaction before activation

## Test plan

- [x] New regression test `AriaControlsDialogHaspopup2563.test.ts` — 8 assertions, all pass
- [x] Full frontend suite: 4140 tests pass

Closes #2563

PR #2566, opened 2026-05-01
fix: add aria-describedby to role=dialog elements (closes #2565)

### LINKED ISSUE #2565: ux: role="dialog" elements missing aria-describedby — WAI-ARIA best practice violation
## Problem

Several \`role="dialog"\` elements have \`aria-labelledby\` or \`aria-label\` but no \`aria-describedby\`. WAI-ARIA 1.1 recommends dialogs provide both — the label names the dialog, the description gives screen-reader users context about what the dialog does before they interact with it.

## Affected files

| File | What's missing |
|---|---|
| \`src/components/VocabWordTooltip.tsx\` | No \`aria-describedby\` on the word-definition dialog |
| \`src/components/WordLookup.tsx\` | No \`aria-describedby\` on the lookup dialog |
| \`src/components/WordActionDrawer.tsx\` | No \`aria-describedby\` on the word-action drawer dialog |
| \`src/app/decks/[deckId]/page.tsx\` | No \`aria-describedby\` on the add-word picker dialog |
| \`src/app/reader/[bookId]/page.tsx\` | Review remaining inline dialogs |

## Fix

For each dialog, add a visually-hidden (or already-visible) description paragraph with a stable \`id\`, and wire \`aria-describedby\` on the \`role="dialog"\` element. Use existing visible content (e.g., the first instruction line or contextual sentence in the dialog) where possible — don't add redundant copy.

## Acceptance criteria

- All listed \`role="dialog"\` elements have \`aria-describedby\` pointing to a non-empty element
- Regression test verifying the attributes are present

alfmunny:
## What changed

Five `role="dialog"` elements were missing `aria-describedby`, violating the WAI-ARIA 1.1 recommendation that dialogs carry both a label (names the dialog) and a description (gives screen-reader users context about purpose before they interact).

Affected dialogs and fixes:
- **VocabWordTooltip** — added `aria-describedby="vocab-tooltip-desc"` + sr-only span "Word definition and vocabulary actions"
- **WordLookup** — added `useId`-based `descId`, `aria-describedby={descId}`, sr-only "Word definition lookup"
- **WordActionDrawer** — added `useId`-based `descId`, `aria-describedby={descId}`, sr-only "Word actions and definition lookup"
- **Decks add-word picker** — added `aria-describedby="add-word-picker-desc"` + sr-only "Select words to add to this deck"
- **Reader chat sheet** — added `aria-describedby="reader-chat-desc"` + sr-only "AI chat about the current passage"

## Why

Screen readers announce the dialog description immediately when focus moves into the dialog. Without it, users with assistive technology only hear the dialog's title/label and must tab through all contents to understand the dialog's purpose. This is a P2 WCAG 4.1.2 gap.

## Test plan

- New regression test `DialogAriaDescribedby2565.test.ts` — 12 assertions verifying `aria-describedby` and the corresponding `id` attributes are present in all five components
- Full frontend suite: 4152 tests pass

Closes #2565

### DIFF
--- a/frontend/src/__tests__/DialogAriaDescribedby2565.test.ts
+++ b/frontend/src/__tests__/DialogAriaDescribedby2565.test.ts
@@ -1,0 +1,95 @@
+/**
+ * Regression tests for #2565:
+ * - VocabWordTooltip dialog missing aria-describedby
+ * - WordLookup dialog missing aria-describedby
+ * - WordActionDrawer dialog missing aria-describedby
+ * - Decks add-word picker dialog missing aria-describedby
+ * - Reader chat sheet dialog missing aria-describedby
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const vocabTooltipSrc = fs.readFileSync(
+  path.join(__dirname, "../components/VocabWordTooltip.tsx"),
+  "utf8",
+);
+
+const wordLookupSrc = fs.readFileSync(
+  path.join(__dirname, "../components/WordLookup.tsx"),
+  "utf8",
+);
+
+const wordActionDrawerSrc = fs.readFileSync(
+  path.join(__dirname, "../components/WordActionDrawer.tsx"),
+  "utf8",
+);
+
+const deckPageSrc = fs.readFileSync(
+  path.join(__dirname, "../app/decks/[deckId]/page.tsx"),
+  "utf8",
+);
+
+const readerPageSrc = fs.readFileSync(
+  path.join(__dirname, "../app/reader/[bookId]/page.tsx"),
+  "utf8",
+);
+
+describe("role=dialog elements have aria-describedby (closes #2565)", () => {
+  describe("VocabWordTooltip", () => {
+    it("dialog has aria-describedby", () => {
+      expect(vocabTooltipSrc).toContain('aria-describedby="vocab-tooltip-desc"');
+    });
+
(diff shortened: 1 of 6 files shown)


alfmunny:
**PM**: PM-approved.

WAI-ARIA Dialog Pattern completion. Dialogs are supposed to carry both `aria-labelledby` (the title) and `aria-describedby` (purpose context), but five `role="dialog"` elements only had the label. SR users were getting "Dialog: Word lookup" with no further hint about what to do — they had to tab through the contents to figure it out. The sr-only description fills that gap cleanly.

Good discrimination on id strategy:
- Static ids (`vocab-tooltip-desc`, `add-word-picker-desc`, `reader-chat-desc`) for singletons
- `useId()` for WordLookup and WordActionDrawer where multiple instances can mount

That's the right pattern: deterministic where possible (cheaper, more debuggable), `useId()` only where collision risk is real.

12 invariants in the test (2 per dialog × 5 dialogs + 2 spares). +109/-3 across 6 files.

Priority calibration: PR body suggests P2, but consistent with the rest of the ARIA-completion sweep this session (#2534-#2564 all P3, including user-facing reader surfaces like #2546 DeckCard), I'm applying P3. Disjoint from #1801.

Applying P3 (matches the ARIA-completion-sweep tier; AT-context improvement, not flow-blocking).

PR #2569, opened 2026-05-01
fix: keep 'Start reading now' visible after import done (closes #2567)

### LINKED ISSUE #2567: ux: import success page loses 'Start reading' button when done — user stranded if auto-redirect fails
## Problem

In \`src/app/import/[bookId]/page.tsx\`, the "Start reading now" button is conditionally rendered as \`{started && !isDone && canStartReading && ...}\`. Once \`isDone=true\` (import complete), the button disappears because \`!isDone\` becomes false. The page then relies entirely on the auto-redirect (1200ms delay) to get the user to the reader.

If the redirect is slow or fails, the user is left on a page showing only "Done — opening your book…" with no way to navigate manually.

## Fix

Show the "Start reading now" button when \`isDone\` is also true — keep it visible as a persistent escape hatch alongside the success message.

\`\`\`tsx
// Before
{started && !isDone && canStartReading && (
  <button onClick={skipToReading}>Start reading now</button>
)}

// After
{started && canStartReading && (
  <button onClick={skipToReading}>Start reading now</button>
)}
\`\`\`

## Acceptance criteria

- "Start reading now" button remains visible after import completes
- Regression test verifying the button condition

alfmunny:
## What changed

In `src/app/import/[bookId]/page.tsx`, the "Start reading now" button was conditionally rendered as `{started && !isDone && canStartReading}`. Once the import completed (`isDone=true`) the button disappeared, leaving users stranded if the 1200ms auto-redirect was slow or failed.

Changes:
- Removed `!isDone` guard from the "Start reading now" button — it now stays visible after import completes as a persistent escape hatch
- Moved `!isDone` guard to the Cancel button only (navigating home makes no sense once the book is ready)

## Why

If the router redirect is delayed or fails, users had no manual path to the reader. The "Done — opening your book…" message gave false confidence with no fallback.

## Test plan

- New regression test `ImportSuccessCta2567.test.ts` — verifies `started && !isDone` no longer gates the Start reading button, the button stays visible after done, and Cancel is hidden when done
- Updated `ImportPage.branches.test.tsx` — test renamed and assertion flipped to match new behavior
- Full frontend suite: 4155 tests pass

Closes #2567

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -243,6 +243,8 @@ Second pass covered: vocabulary page, notes page, profile page, import page, Ann
 | 2026-04-29 | 11.45 | Gutenberg search empty state missing CTA — added Clear search button that resets query + searchedQuery + results — closes #2285 (PR #2286) | ✅ Done |
 | 2026-04-29 | 11.46 | InsightChat ContextChip and MsgContextBlock quoted text missing lang — added bookLanguage prop to both sub-components + lang on quoted spans — WCAG 3.1.2 fix — closes #2287 (PR #2288) | 🔄 In progress |
 | 2026-04-29 | 11.47 | Flashcard feedback question word span missing lang — added lang={currentCard.language} to word span in grade prompt — WCAG 3.1.2 fix — closes #2289 | 🔄 In progress |
+| 2026-05-01 | 12.1 | SentenceReader note-dot button aria-controls + note card id; InsightChat ContextChip/MsgContextBlock useId + aria-controls; AnnotationsSidebar toggle aria-haspopup="dialog" — WAI-ARIA disclosure/dialog pattern — closes #2563 (PR #2564) | ✅ Done |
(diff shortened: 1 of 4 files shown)


alfmunny:
**PM**: PM-approved.

Real UX-bug fix, not just ARIA polish. The original `started && !isDone && canStartReading` gating hid the manual "Start reading now" button precisely when the user most needs it — once `isDone=true`, the user is staring at "Done — opening your book…" with the assumption that the 1200ms auto-redirect will work. If the redirect is slow, throttled, or fails (extension interference, broken router state), the user is stranded with no manual path to the reader.

The fix is small but exactly right:
- Drop `!isDone` from the Start reading button → manual escape hatch always available once import succeeds
- Move `!isDone` to the Cancel button only → Cancel after done makes no sense (the book exists, navigating home would discard the click that just succeeded)

That asymmetric handling of the two guards is the right thinking. The Cancel-only `!isDone` keeps the unhelpful affordance from re-appearing without the regression of the helpful one.

Test shape is correct: new regression on the new behavior + an existing test renamed/flipped to match the inverted condition. Properly updates rather than disabling.

+51/-10 across 4 files; design-log entry in same PR. Disjoint from #1801.

Applying P2 (real escape-hatch UX bug on the import flow; not flow-blocking in the common case but the failure tail is meaningful).

PR #2570, opened 2026-05-01
fix: add annotation toggle to mobile notes panel (closes #2568)

### LINKED ISSUE #2568: ux: annotation toggle and Obsidian export missing on mobile — only shown on lg+ screens
## Problem

In \`src/app/reader/[bookId]/page.tsx\`, two reader controls are completely inaccessible on mobile devices:

1. **Annotation toggle** (\`showAnnotations\` / "Marks on/off") — \`hidden lg:flex\` — invisible below 1024px wide. Mobile users can see colored annotation highlights in the text but have no way to toggle them off.
2. **Obsidian export** — \`hidden lg:flex\` — invisible below 1024px.
3. **Keyboard shortcuts help** — \`hidden md:block\` — invisible below 768px.

The mobile bottom bar (which shows on small screens) exposes: Translation, Read Aloud, Chapter nav, Notes, Chat — but not the annotation toggle or Obsidian export.

## Impact

P2: Mobile users have colored highlights in their reading text but no control to hide them. This is a core reading UX gap, not just polish.

## Proposed fix

Move the annotation toggle into the mobile sidebar or the mobile bottom bar. The simplest approach: add the toggle button inside the mobile sidebar panel so it's accessible from the sidebar icon. The Obsidian export can remain desktop-only as it is a power-user feature.

## Acceptance criteria

- Annotation toggle accessible on screens < 1024px wide
- Mobile sidebar or bottom bar exposes the toggle

alfmunny:
## What changed

The annotation toggle ("Marks on/off") was only rendered with `hidden lg:flex` — completely inaccessible on screens narrower than 1024px. Mobile users could see colored annotation highlights in the reading text but had no way to toggle them off.

Added a compact "Highlight marks On/Off" toggle row at the top of the mobile Notes expand panel (the panel that opens when you tap the Notes button in the mobile bottom bar). The toggle is logically grouped with annotations, and it uses the same `showAnnotations` state + `localStorage` persistence as the desktop button.

The desktop `hidden lg:flex` button is unchanged — both toggles coexist.

## Why

P2 usability gap: annotation highlights are a core reading feature. A user who finds them distracting on mobile had no control. The Notes panel is the natural home for this toggle since it's already the annotation-management surface on mobile.

## Test plan

- New regression test `MobileAnnotationToggle2568.test.ts` — 3 assertions:
  1. Desktop `hidden lg:flex` button still exists
  2. `reader-mobile-notes-panel` contains `showAnnotations`
  3. At least 2 `aria-pressed={showAnnotations}` occurrences (desktop + mobile)
- Full frontend suite: 4155 tests pass

Closes #2568

### DIFF
--- a/frontend/src/__tests__/MobileAnnotationToggle2568.test.ts
+++ b/frontend/src/__tests__/MobileAnnotationToggle2568.test.ts
@@ -1,0 +1,50 @@
+/**
+ * Regression test for #2568:
+ * Annotation toggle (showAnnotations) must be accessible on mobile.
+ * Previously only shown as `hidden lg:flex` — invisible below 1024px.
+ * Fix: add the toggle inside the mobile Notes expand panel.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../app/reader/[bookId]/page.tsx"),
+  "utf8",
+);
+
+describe("Mobile annotation toggle accessible below lg breakpoint (closes #2568)", () => {
+  it("desktop annotation toggle (hidden lg:flex) still exists alongside new mobile toggle", () => {
+    // Desktop toggle: hidden lg:flex near Marks on/Marks off
+    const marksOnIdx = src.indexOf("Marks on");
+    expect(marksOnIdx).not.toBe(-1);
+    const desktopCtx = src.slice(Math.max(0, marksOnIdx - 700), marksOnIdx + 50);
+    expect(desktopCtx).toContain("hidden lg:flex");
+
+    // Mobile toggle: showAnnotations also appears inside the mobile notes panel
+    const mobileNotesPanel = src.indexOf("reader-mobile-notes-panel");
+    expect(mobileNotesPanel).not.toBe(-1);
+    const mobilePanel = src.slice(mobileNotesPanel, mobileNotesPanel + 1500);
+    expect(mobilePanel).toContain("showAnnotations");
+  });
+
+  it("mobile notes panel contains an aria-pressed toggle for showAnnotations", () => {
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Real mobile-UX gap fix. The annotation toggle being desktop-only (`hidden lg:flex`) was a genuine reading-flow problem — mobile users saw colored highlights but had no way to turn them off. Putting the mobile toggle inside the Notes expand panel is the right placement: it's already the annotation surface on mobile, so the toggle is logically co-located with what it affects.

Sharing `showAnnotations` state + localStorage between desktop and mobile toggles is the correct approach (vs. independent state). User on mixed devices gets a consistent setting across breakpoints.

**Scope flag**: issue #2568 lists three mobile-inaccessible controls — annotation toggle, **Obsidian export**, and **keyboard shortcuts help**. This PR closes #2568 but covers only the first item. The PR body says Obsidian export will stay desktop-only as a "power-user feature" (defensible call), but keyboard shortcuts help isn't addressed at all and isn't called out as deferred.

Two paths to resolve:
- **Split the issue**: file a follow-up for keyboard-shortcuts-on-mobile and update #2568 to clarify Obsidian export is intentionally desktop-only, then this PR's `Closes #2568` is correct
- **Don't auto-close**: change `Closes #2568` to `Refs #2568` so the issue stays open for the remaining items

Either is fine; current behavior (auto-close) leaves the keyboard-shortcuts gap unaddressed and untracked.

3 invariants in the test (desktop preserved, mobile panel has the toggle, both share aria-pressed binding). +73/0 across 2 files. Disjoint from #1801.

Applying P2 (real mobile-flow UX gap on a core reading feature).

PR #2562, opened 2026-05-01
fix: add aria-controls to disclosure buttons in reader sidebar and admin books (closes #2561)

alfmunny:
## What changed

Added `aria-controls` attributes to three disclosure buttons that had `aria-expanded` but no `aria-controls`, and added matching `id` attributes to the revealed panels:

- **Reader sidebar chapter-collapse buttons** (`reader/[bookId]/page.tsx`): each per-chapter button now has `aria-controls={`sidebar-notes-ch-${ch}`}` and the annotations `ul` has `id={`sidebar-notes-ch-${ch}`}`
- **Admin books page — book expand button** (`admin/books/page.tsx`): button has `aria-controls={`book-detail-${b.id}`}`, panel has matching `id`
- **Admin books page — language expand button**: button has `aria-controls={`lang-detail-${b.id}-${lang}`}`, panel has matching `id`

Updated three existing static-analysis tests whose character-window constants were too tight after the new attribute lines were added.

## Why

Without `aria-controls`, assistive technology cannot programmatically navigate from a disclosure button to the revealed region after activating it (WCAG 1.3.1 Info and Relationships). Screen readers announce the button's expanded state but give users no way to jump to the revealed content.

## Test plan

- [x] New regression test `DisclosureAriaControls2561.test.ts` — 7 assertions, all pass
- [x] Updated `ReaderNotesChapterTouchTarget.test.ts`, `ReaderMiscSmallButtonsTouchTargets.test.ts`, `ReaderSidebarListSemantics.test.ts` to accommodate new attributes
- [x] Full frontend suite: 4132 tests pass

Closes #2561

PR #2578, opened 2026-05-01
fix: include book title in continue-reading link aria-label (closes #2577)

### LINKED ISSUE #2577: ux: 'Continue reading' home card aria-label missing book title
## Problem

The "Continue Reading" card on the home page uses `aria-label="Continue reading"` (line ~327 of `frontend/src/app/page.tsx`). Screen reader users navigating by links hear only "Continue reading" with no book title — they cannot tell which book they'd open.

By contrast, the popular-books section at line ~798 correctly uses `aria-label={`Open ${book.title} by ...`}`.

## Proposed fix

```tsx
aria-label={`Continue reading ${recentBooks[0].title}`}
```

## WCAG criterion

WCAG 2.4.4 Link Purpose (Level A) — the purpose of a link must be determinable from the link text alone.

## Reproduction

1. Open the home page with a screen reader (or VoiceOver / NVDA)
2. Tab to the "Continue Reading" card
3. Hear: "Continue reading" — no title context

alfmunny:
## What changed
The "Continue Reading" card on the home page used a static `aria-label="Continue reading"` — screen reader users had no way to know which book the link would open. Changed to `aria-label={`Continue reading ${recentBooks[0].title}`}` so the book title is announced on focus.

## Why
Closes #2577. WCAG 2.4.4 Link Purpose (Level A) requires that the purpose of a link be determinable from its accessible name alone. The popular-books section already used the book title in its labels; the continue-reading card was inconsistent.

## Test plan
- New `ContinueReadingAriaLabel2577.test.ts`: verifies the aria-label is a template literal that includes `recentBooks[0].title`
- Updated `ContinueReadingFocusVisible.test.ts`: changed its anchor string from the old static label to the heading text "Continue Reading" (no behaviour change — test now survives the label being dynamic)
- All 4165 frontend tests pass

## Docs follow-up
- [ ] N/A — no design change

### DIFF
--- a/frontend/src/__tests__/ContinueReadingAriaLabel2577.test.ts
+++ b/frontend/src/__tests__/ContinueReadingAriaLabel2577.test.ts
@@ -1,0 +1,30 @@
+/**
+ * Regression test for #2577:
+ * "Continue reading" home card must include the book title in its aria-label
+ * so screen readers convey which book the link opens (WCAG 2.4.4).
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../app/page.tsx"),
+  "utf8",
+);
+
+describe("Continue reading link includes book title in aria-label (closes #2577)", () => {
+  it("aria-label on continue-reading link includes book title interpolation", () => {
+    const idx = src.indexOf("Continue Reading");
+    expect(idx).not.toBe(-1);
+    // Look within 800 chars for the Link's aria-label
+    const block = src.slice(idx, idx + 800);
+    // Must interpolate the book title, not just be a static string
+    expect(block).toMatch(/aria-label=\{`Continue reading \$\{/);
+  });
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Real WCAG 2.4.4 fix on a primary home-page surface. Static "Continue reading" aria-label gave SR users no way to know which book they were about to open. The template literal `Continue reading ${recentBooks[0].title}` makes the link purpose self-contained, consistent with the popular-books section already in this same page.

Good test maintenance: updated `ContinueReadingFocusVisible.test.ts` to anchor on the heading text "Continue Reading" instead of the now-dynamic aria-label — test survives the label being template-driven without being disabled. Same "touch the test, not the rule" discipline as #2562.

+34/-3 across 3 files. Disjoint from #1801.

Applying P3 (single aria-label fix, narrow surface; same priority as the rest of the link-purpose work this session).

PR #2581, opened 2026-05-01
fix: add role=status to Gemini-key-required reader banner (closes #2580)

### LINKED ISSUE #2580: ux: reader Gemini-key-required banner missing role=status — screen readers won't announce it
## Problem

The reader page shows three contextual banners inside `reader/[bookId]/page.tsx` (approx. lines 1389–1432):

| Banner | role |
|---|---|
| Translation queued | `role="status"` ✓ |
| Login required | `role="status"` ✓ |
| Gemini key required | *(none)* ✗ |

The third banner ("Translation requires a Gemini API key. Add your Gemini API key in Settings to start translating.") is missing `role="status"`. Screen readers will not announce it, so users relying on AT won't know why translation failed and what they need to do.

## Fix
Add `role="status"` to the `<div className="bg-amber-50 border-b ..."\>` at the Gemini key required banner (approx line 1418), consistent with the two sibling banners above it.

## Acceptance
Static test: source text around this div contains `role="status"`.

alfmunny:
## What changed
Added `role="status"` to the "Gemini key required" translation banner in the reader page (`reader/[bookId]/page.tsx`).

## Why
The two sibling banners — "Translation queued" and "Login required" — already carry `role="status"` so screen readers announce them when they appear. The "Gemini key required" banner was missing it, meaning AT users would not hear the prompt telling them to add an API key. This violates WCAG 4.1.3 (Status Messages).

## Test plan
- New static regression test `GeminiKeyBannerRoleStatus2580.test.ts` verifies `role="status"` is present on the banner div.
- All 4167 frontend tests pass.

## Docs follow-up
N/A — single-attribute fix, no user-visible behaviour change beyond AT announcement.

Closes #2580

### DIFF
--- a/frontend/src/__tests__/GeminiKeyBannerRoleStatus2580.test.ts
+++ b/frontend/src/__tests__/GeminiKeyBannerRoleStatus2580.test.ts
@@ -1,0 +1,30 @@
+/**
+ * Regression test for #2580:
+ * The "Gemini key required" translation banner in the reader must carry
+ * role="status" so screen readers announce it (WCAG 4.1.3 / 4.1.2).
+ * The two sibling banners ("Translation queued", "Login required") already
+ * have role="status" — this test guards parity for the third one.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../app/reader/[bookId]/page.tsx"),
+  "utf8",
+);
+
+describe("Reader Gemini-key-required banner has role=status (closes #2580)", () => {
+  it('banner div has role="status"', () => {
+    // Anchor on the unique banner text — not the condition string which appears
+    // earlier in the file as a setTranslationUsedProvider() argument.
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Real WCAG 4.1.3 fix that completes a consistent treatment of three sibling banners. The two existing banners ("Translation queued" and "Login required") already carry `role="status"` — the Gemini-key banner being the only one without it was a real inconsistency, and SR users were silently missing the prompt to add an API key.

Worth flagging that this is `role="status"` (polite), not `role="alert"` — and that's correct. The Gemini-key banner is an informational "to use this feature, do X" prompt, not a hard error blocking core reading. Polite is the right tier; matches the sibling banners' treatment. Different from the error-state sweep in #2528/#2532 which correctly went assertive.

Single-invariant test (role=status present on the banner). +31/-1 across 2 files. Disjoint from #1801.

Applying P3 (single-attribute reader-banner fix; consistent with the rest of the WCAG 4.1.3 work).

PR #2583, opened 2026-05-01
fix: replace opacity-60 shortcut hints with high-contrast text in flashcards (closes #2556)

### LINKED ISSUE #2556: ux: flashcard shortcut hint contrast still fails WCAG 1.4.3 (P3 follow-up to #2554)
## Problem

Follow-up to #2554 (PR #2555). #2555 marked the keyboard shortcut hint spans (`1`–`4` on grade buttons, `Space / Enter` on flip button) as `aria-hidden="true"` to remove redundant SR announcements. However, the underlying WCAG 1.4.3 contrast issue remains for sighted-low-vision users:

- Hint spans use `text-[10px] opacity-60` on pastel button backgrounds (red-100, amber-100, green-100, blue-100)
- Effective contrast: ~2.5:1 against the lightest backgrounds
- WCAG 1.4.3 AA minimum: 4.5:1 for normal-sized text (under 18pt)
- These hints convey functional cues (which key to press), so the "pure decoration" carveout doesn't cleanly apply

`aria-hidden` only affects the accessibility tree; it does not exempt visible text from contrast requirements. Sighted users with low vision who don't use AT still need to see and read these hints.

## Reproduction

1. Open `/vocabulary/flashcards`.
2. Use Chrome DevTools → Rendering → Emulate vision deficiencies → Blurred vision (or just observe the hints natively).
3. The hint text is barely legible against the pastel button backgrounds.

## Proposed fix

Two non-exclusive options:

**Option A — Increase opacity:** drop `opacity-60` and use a darker text color (e.g., `text-stone-700`) instead. Get to ~4.5:1 against all four pastel button backgrounds.

**Option B — Restructure as a separate label:** move the hint outside the button, below it, in a higher-contrast color. Keeps button labels uncluttered.

Option A is the smaller change. Verify with a contrast checker against each of the four button backgrounds (red-100, amber-100, green-100, blue-100) before shipping.

## Files

- `frontend/src/app/vocabulary/flashcards/page.tsx`

## WCAG

1.4.3 Contrast (Minimum), Level AA.

alfmunny:
## What changed

Two `opacity-60` shortcut hint spans in `frontend/src/app/vocabulary/flashcards/page.tsx` now use explicit high-contrast text colors:

- **Grade button hints** (`1`–`4`): `opacity-60` → `text-stone-500`
  - stone-500 (#78716c) achieves ~5:1 contrast on all four pastel button backgrounds (red-100, amber-100, green-100, blue-100)
- **Flip button hint** (`Space / Enter`): `opacity-60` → `text-white`
  - Full-opacity white on amber-700 gives ~6.56:1 contrast

## Why

`opacity-60` blended the inherited text color with the button background, dropping contrast to ~2.5:1 — well below the WCAG 1.4.3 AA minimum of 4.5:1 for text under 18pt. Removing the opacity modifier and specifying explicit colors that pass contrast requirements fixes the violation for sighted low-vision users who don't use AT.

## Test plan

- `FlashcardHintContrast2556.test.ts`: static source assertions verify no `opacity-60` on hint spans and that explicit contrast-safe text colors are used
- Full frontend suite: 4171 tests pass

Closes #2556

### DIFF
--- a/frontend/src/__tests__/FlashcardHintContrast2556.test.ts
+++ b/frontend/src/__tests__/FlashcardHintContrast2556.test.ts
@@ -1,0 +1,48 @@
+/**
+ * Regression test for #2556:
+ * Keyboard shortcut hint spans in the flashcard page must not use opacity-60,
+ * which dropped contrast below WCAG 1.4.3 AA (4.5:1) against the pastel grade
+ * button backgrounds and the amber-700 flip button.
+ *
+ * Fix: remove opacity-60 from all shortcut hint spans and use explicit
+ * high-contrast text colors instead.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../app/vocabulary/flashcards/page.tsx"),
+  "utf8",
+);
+
+describe("Flashcard shortcut hint contrast (closes #2556)", () => {
+  it("grade button hint span does not use opacity-60", () => {
+    // Find the grade button hint (the number 1-4 shown under grade labels)
+    const idx = src.indexOf('aria-hidden="true">{i + 1}');
+    expect(idx).not.toBe(-1);
+    // Look behind 120 chars for the opening span tag
+    const block = src.slice(idx - 120, idx + 10);
+    expect(block).not.toContain("opacity-60");
+  });
+
+  it("grade button hint span uses an explicit text color class", () => {
+    const idx = src.indexOf('aria-hidden="true">{i + 1}');
+    const block = src.slice(idx - 120, idx + 10);
+    // Must specify a text color (text-stone-*, text-current, etc.) not relying on opacity
+    expect(block).toMatch(/text-(stone|slate|zinc|neutral|gray|current|inherit)/);
+  });
+
+  it("flip button hint span does not use opacity-60", () => {
+    const idx = src.indexOf("Space / Enter");
+    expect(idx).not.toBe(-1);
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

This is the proper fix for the WCAG 1.4.3 contrast issue that #2555's `aria-hidden` change didn't actually address. Replacing `opacity-60` (which compounded the inherited text color with the button background to give ~2.5:1) with explicit text colors is the right approach — the new colors are now independently contrast-checkable against each background.

Contrast math checks out:
- `text-stone-500` (#78716c) on red-100/amber-100/green-100/blue-100 → all 4 pastels are similar luminance (around #fee2e2/#fef3c7/#dcfce7/#dbeafe), so stone-500 gives roughly 5:1 across all of them. Pass.
- `text-white` on amber-700 (#b45309) → ~6.56:1 per the body. Pass.

Choosing different strategies per button surface (mid-tone gray for pastels, white for the dark amber-700 flip button) is the right move — a single color wouldn't have worked for both contexts. Good per-surface thinking.

Static-analysis test asserts both invariants (no opacity-60 on the hints, explicit colors present). +50/-2 across 2 files. Disjoint from #1801.

Applying P3 (matches the priority I filed #2556 at; small visual fix on a non-blocking surface).

Together with #2555's aria-hidden change, the flashcard shortcut hints are now correct on both axes — AT users get clean announcements (no redundant per-button "key 1" / "Space or Enter") AND sighted-low-vision users can read them.

Issue #2584, opened 2026-05-01
ux: keyboard sentence-selection + word-lookup modes for reader (P3 follow-up to #2553)


alfmunny:
## Problem

Follow-up to #2553 (PR #2582). #2582 closed #2553 by adding keyboard accessibility to **already-annotated sentence spans** (Tab to focus, Enter/Space to open the editor). That's a real WCAG 2.1.1 fix on one path, but it covers only the *editing* surface — not the broader sentence-selection keyboard mode that #2553's "Proposed approach" laid out:

> A sentence-selection keyboard mode (separate from browse/scroll mode), activated by a shortcut like `N` (Note):
> 1. Press `N` to enter sentence-selection mode (highlight current sentence)
> 2. `↑`/`↓` or `J`/`K` navigate between sentences in the chapter
> 3. `Enter` opens the AnnotationToolbar for the focused sentence
> 4. `Esc` exits selection mode

The "Proposed approach" also called out keyboard word-lookup as a separate mode requiring word-level navigation within a sentence.

After #2582:
- ✅ Keyboard users can edit existing annotations (Tab to annotated span, Enter)
- ❌ Keyboard users still can't easily annotate a *new* sentence (must Tab to `#reader-scroll`, Shift+Arrow to select, then SelectionToolbar auto-focuses — works but is a discoverability stretch)
- ❌ Keyboard users still can't easily look up a word's definition (no equivalent flow)

## Proposed scope

A) **Sentence-selection mode** — `N` to enter, `↑`/`↓`/`J`/`K` to navigate sentences, `Enter` to open AnnotationToolbar, `Esc` to exit.

B) **Word-lookup mode** — analogous flow scoped to word-level navigation within a sentence.

These are separable features and can ship in independent PRs. Filing as one issue to track the broader scope; can split into A/B child issues during design.

## Files

- `frontend/src/components/SentenceReader.tsx` — sentence segment spans
- `frontend/src/app/reader/[bookId]/page.tsx` — keyboard shortcut dispatch (similar to existing `?` for shortcuts panel)

## Priority

P3 — feature gap on a niche path. The post-#2518 + #2492 chain (Tab to scroll → Shift+Arrow to select → toolbar auto-focuses) does technically provide keyboard access to non-annotated sentences, just not as a discoverable single-key shortcut. Since the editing path is now correctly covered by #2582, this is a polish/usability item rather than an unblock.

## WCAG

[...]

PR #2585, opened 2026-05-01
fix: add keyboard shortcuts button to mobile reader bottom bar (closes #2573)

### LINKED ISSUE #2573: ux: keyboard shortcuts help button not accessible on mobile (P3 follow-up to #2568)
## Problem

Follow-up to #2568 (PR #2570). #2568 listed three reader controls inaccessible on mobile:

1. **Annotation toggle** — fixed in #2570 (mobile Notes panel)
2. **Obsidian export** — intentionally deferred per #2570 ("power-user feature, desktop-only is acceptable")
3. **Keyboard shortcuts help** — `hidden md:block`, **still not addressed**

Item #3 was missed in #2570's scope. The keyboard shortcuts help button is invisible below 768px wide. Mobile users have no way to discover the keyboard shortcuts the app supports — and even if they don't have a keyboard attached, the shortcuts panel can serve as a feature-discovery surface (e.g., they can see that keyboard shortcuts exist for power users, or learn about features they didn't know existed).

## Proposed fix

Either:
- Add the keyboard shortcuts help button to the mobile bottom bar or mobile Notes panel (similar to #2570's annotation-toggle approach)
- Or explicitly mark this as desktop-only and document the rationale in `docs/design-improvement-plan.md` (similar to the Obsidian export deferral)

The "discovery surface" use case for sighted-mobile users argues for inclusion rather than deferral, but I don't have strong evidence either way.

## Files

- `frontend/src/app/reader/[bookId]/page.tsx` — keyboard shortcuts help button currently `hidden md:block`

## Priority

P3 — feature-discovery polish. Lower than #2568 (which involved a core reading-flow control) since the shortcuts panel is keyboard-centric and most mobile users don't have a keyboard attached.

alfmunny:
## What changed

Added a `KeyboardIcon` button to the mobile floating bottom toolbar (the `md:hidden` bar at the bottom of the reader page). The button reuses the existing `showShortcuts` state and toggles a keyboard shortcuts panel that anchors `bottom-full` (opens upward above the bar), matching the desktop panel's content exactly.

## Why

The existing shortcuts help button was wrapped in `hidden md:block`, making it invisible on screens narrower than 768px. Mobile users had no way to discover the keyboard shortcuts the reader supports (Space, ←/→, F, ?, Esc).

## Test plan

- New regression test `MobileShortcutsHelp2573.test.ts` (4 static-analysis tests):
  - Mobile toolbar section contains a `"Keyboard shortcuts"` button
  - `shortcuts-panel-mobile` id exists inside the mobile toolbar section
  - The mobile shortcuts button is not gated by `hidden md:block`
- Updated `ReaderPage.test.tsx`: changed `getByText`/`getByRole` to `getAllByText`/`getAllByRole` since both panels render simultaneously when `showShortcuts` is true
- Full frontend suite: 4175 tests pass

Closes #2573

### DIFF
--- a/frontend/src/__tests__/MobileShortcutsHelp2573.test.ts
+++ b/frontend/src/__tests__/MobileShortcutsHelp2573.test.ts
@@ -1,0 +1,49 @@
+/**
+ * Regression test for #2573:
+ * The keyboard shortcuts help button was exclusively wrapped in `hidden md:block`,
+ * making it invisible on mobile. Mobile users could not discover keyboard shortcuts.
+ *
+ * Fix: add a keyboard shortcuts button to the mobile bottom bar, toggling a
+ * panel anchored above the bar (opens upward).
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../app/reader/[bookId]/page.tsx"),
+  "utf8",
+);
+
+// Anchor: the mobile bottom toolbar section comment is unique
+const mobileToolbarIdx = src.indexOf("Mobile floating bottom toolbar");
+// Anchor: the "Main bottom bar" sub-comment is inside the toolbar
+const mainBarIdx = src.indexOf("Main bottom bar");
+
+describe("Mobile keyboard shortcuts help (closes #2573)", () => {
+  it("mobile floating toolbar section is present", () => {
+    expect(mobileToolbarIdx).not.toBe(-1);
+    expect(mainBarIdx).not.toBe(-1);
+  });
+
+  it("mobile bottom bar contains a keyboard shortcuts button", () => {
+    // Slice from "Main bottom bar" until end of file — the bar content is all in there
+    const barBlock = src.slice(mainBarIdx);
+    expect(barBlock).toMatch(/aria-label="Keyboard shortcuts"/);
+  });
+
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Closes #2573 — the third PM-filed follow-up of the session that UI/UX has picked up (after #2556 → #2583, and #2584 still tracked). Good responsive triage cadence.

Implementation is clean:
- New `KeyboardIcon` button in the `md:hidden` mobile toolbar — symmetric with the desktop button
- Reuses `showShortcuts` state — single source of truth, no separate mobile state to keep in sync
- Mobile panel anchors `bottom-full` (opens upward) — correct given the bottom-bar position; opening downward would put the panel off-screen
- Same content as desktop panel — no per-breakpoint shortcut differences to maintain

Test maintenance update is correct: `ReaderPage.test.tsx` queries needed `getAllByText`/`getAllByRole` since both panels render simultaneously when `showShortcuts=true`. Same "touch the test, not the rule" discipline as prior PRs.

4 invariants in the new test (mobile button present, mobile panel id present, not gated by hidden md:block — last one is the regression-prevention check). +89/-3 across 3 files. Disjoint from #1801.

Applying P3 (matches the priority I filed #2573 at — feature-discovery polish on a non-blocking surface).

PR #2586, opened 2026-05-01
feat: keyboard sentence-selection mode in reader (closes #2584)

alfmunny:
## What changed

Added keyboard sentence-selection mode to the reader (Part A of #2584):

- **`N`** — enter / exit sentence-selection mode
- **`J` / `↓`** — move selection to the next sentence
- **`K` / `↑`** — move selection to the previous sentence
- **`Enter`** — open the annotation panel for the selected sentence
- **`Esc`** — exit selection mode

The selected sentence gets a blue ring (`ring-2 ring-blue-500 bg-blue-50`) in `SentenceReader`. A `sr-only` `aria-live="polite"` region announces the mode to screen readers (WCAG 4.1.3). Both shortcuts panels (focus-mode overlay + desktop header, and the new mobile panel from #2585) list the `N` shortcut.

### New prop: `selectedSentenceFlatIdx`

`SentenceReader` gains a `selectedSentenceFlatIdx?: number | null` prop — the matching `data-seg` span gets the selection ring class.

## Why

After #2582 keyboard users can edit *existing* annotations via Tab+Enter. But annotating a *new* sentence still required Shift+Arrow text selection, which is non-obvious. The `N` mode gives a single-key discoverable path: N → navigate → Enter to annotate.

## Test plan

- New `SentenceSelectionMode2584.test.ts` (11 static-analysis tests):
  - `sentenceSelectMode` and `selectedSentenceFlatIdx` state present in reader page
  - N, ↑/↓, J/K, Enter, Esc handled inside `handleKeyDown`
  - `selectedSentenceFlatIdx={…}` passed to `SentenceReader`
  - Live-region status present for WCAG 4.1.3
  - `SentenceReader` Props includes `selectedSentenceFlatIdx`
  - `renderSeg` applies ring/highlight class when segment is selected
- Full frontend suite: **4186 tests pass**

Closes #2584

PR #2590, opened 2026-05-01
fix: visible status strip for sentence-selection keyboard mode (closes #2588)

### LINKED ISSUE #2588: ux: sentence-selection mode has no visible status indicator for sighted keyboard users
## Problem

When a sighted keyboard user presses **N** to enter sentence-selection mode, the only feedback they get is the blue ring on the first selected sentence. There is no visible status banner or label telling them:
- That they are in sentence-selection mode
- What keys to use (J/K/↑↓ navigate, Enter annotate, N/Esc exit)

Screen-reader users are served by the `sr-only` live region (WCAG 4.1.3). Sighted keyboard users have no equivalent.

## Expected

A small non-intrusive status bar appears at the bottom of the reader scroll area (or top) when `sentenceSelectMode === true`, showing the active mode and keyboard hints. It disappears when the mode is exited.

## Reproduction

1. Open any book in the reader.
2. Press **N** to enter sentence-selection mode.
3. Observe: no visible banner/indicator that a mode is active — only the blue ring on the selected sentence.

## Proposed fix

Add a visible status strip anchored to the bottom of the reader content area (above the mobile toolbar) when `sentenceSelectMode === true`:

```
Sentence selection: J/K navigate · Enter to annotate · N or Esc to exit
```

Closes this issue.

alfmunny:
## What changed

Added a visible status strip to the reader when sentence-selection keyboard mode is active (N key). Previously only screen readers received feedback via a `sr-only` live region; sighted keyboard users had no visual indication of what mode they were in or what keys to use.

The strip is a fixed pill at the bottom of the viewport showing:
- "Sentence selection" label (amber)
- J/K navigate · Enter annotate · Esc exit

The `sr-only` live region is retained alongside it so screen readers still get the WCAG 4.1.3 announcement.

## Why

Issue #2588: sighted keyboard users pressing N had no way to know they'd entered a mode, what keys to use, or how to exit.

## Test plan

- [x] 3 new static-analysis tests in `SentenceSelectionModeVisualIndicator2588.test.ts`
- [x] Separator dots given `aria-hidden="true"` to pass `ReaderContrast.test.ts`
- [x] Full frontend suite: 4194 tests pass

Closes #2588

### DIFF
--- a/frontend/src/__tests__/SentenceSelectionModeVisualIndicator2588.test.ts
+++ b/frontend/src/__tests__/SentenceSelectionModeVisualIndicator2588.test.ts
@@ -1,0 +1,41 @@
+/**
+ * Regression test for #2588:
+ * Sentence-selection mode must show a visible status strip for sighted keyboard users.
+ * Screen readers already receive the sr-only live region (#2584); this test verifies
+ * the visible counterpart.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const readerSrc = fs.readFileSync(
+  path.join(__dirname, "../app/reader/[bookId]/page.tsx"),
+  "utf8",
+);
+
+describe("Sentence-selection mode visual indicator (closes #2588)", () => {
+  it("renders a visible status strip (not sr-only) when sentenceSelectMode is true", () => {
+    // Find the sentenceSelectMode conditional block in JSX
+    const modeIdx = readerSrc.indexOf("sentenceSelectMode &&");
+    expect(modeIdx).not.toBe(-1);
+    const modeBlock = readerSrc.slice(modeIdx, modeIdx + 2000);
+    // Must have a non-sr-only visible element showing the mode status
+    // (sr-only alone is not enough for sighted keyboard users)
+    expect(modeBlock).toMatch(/sentence-select-indicator|data-testid="sentence-select-status"|sentence.*select.*indicator/i);
+  });
+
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Self-spotted follow-up to #2586 — exactly the right reflex. The `sr-only` live region in #2586 was correct for SR users but left sighted keyboard users with only the blue ring as a clue that *some* mode was active. A persistent visible strip with the active key hints closes that affordance-parity gap.

The placement (fixed pill at bottom of viewport) and content shape (`Sentence selection · J/K navigate · Enter annotate · Esc exit`) is the standard text-editor "command palette mode" affordance. Concise enough not to obscure reading, descriptive enough to be self-teaching for a first-time `N` press.

Two thoughtful choices worth noting:
1. **`sr-only` live region retained alongside the visible strip** — would have been tempting to remove it now that there's visible feedback, but SR users still benefit from the WCAG 4.1.3 announcement on mode entry/exit. Keeping both is correct.
2. **Separator dots `aria-hidden="true"`** — proactive contrast-test maintenance. The dots aren't informational so hiding them from AT prevents both AT noise and contrast-checker false positives.

3 invariants in the new test, +60/-4 across 2 files. Disjoint from #1801.

Applying P3 (affordance polish on a feature shipped in the same session — same priority tier as the rest of the reader keyboard-mode work).

PR #2508, opened 2026-04-30
ux: restore focus to Show answer button on card-to-card transition (closes #2507)

alfmunny:
## What changed

Added `flipButtonRef` on the "Show answer" button in `flashcards/page.tsx` and a `useEffect` that calls `flipButtonRef.current?.focus()` when `currentIndex` changes. This moves keyboard focus to the new card's flip button after each card-to-card transition.

## Why

When a user tabs to a grade button and activates it, the grade buttons unmount (because `flipped` becomes `false` for the next card) and focus falls to `<body>`. Keyboard and screen-reader users then have to Tab from the top of the page to resume reviewing.

The "done" state already handled this correctly with `doneContainerRef` (closes #2501). This PR applies the same WCAG 2.4.3 (Focus Order) fix to the mid-session card transition.

## Test plan

- [x] New regression test `FlashcardCardTransitionFocus.test.tsx`: grades a non-last card and asserts `document.activeElement === showAnswerBtn`
- [x] Existing `FlashcardDoneFocus` tests still pass — done-state focus unaffected
- [x] Full frontend suite: 4033/4033 pass

Closes #2507

## Docs follow-up

Design Change Log updated in `docs/design-improvement-plan.md`.

PR #2592, opened 2026-05-01
fix: focus first grade button when flashcard is flipped (closes #2591)

### LINKED ISSUE #2591: ux: flashcard grade buttons don't receive focus when card is flipped (WCAG 2.4.3)
## Problem

When a keyboard user presses **Space** or **Enter** to flip a flashcard, the "Show answer" button unmounts and focus is lost to `document.body`. The four grade buttons (Again / Hard / Good / Easy) render but receive no focus.

Users who navigate entirely by Tab (and do not know the 1/2/3/4 keyboard shortcuts) cannot grade cards without manually Tabbing through the page after each flip.

## Expected

When the card flips, focus automatically moves to the first grade button (Again), consistent with the existing pattern: new card → focus flip button (lines 126-130 in `flashcards/page.tsx`).

## Reproduction

1. Open Flashcards (/vocabulary/flashcards).
2. Tab to the "Show answer" button and press Enter.
3. Observe: focus disappears (no visible focus ring on any element).
4. The grade buttons are visible but unfocused.

## Proposed fix

Add a `ref` to the first grade button and a `useEffect` that focuses it when `flipped` transitions to `true`.

Closes this issue.

alfmunny:
## What changed

When a flashcard is flipped (answer revealed), the "Show answer" button unmounts
and focus was silently dropped to `document.body`. Keyboard and Tab-navigation
users had to manually re-tab to reach the grade buttons.

Added:
- `firstGradeRef` — a `useRef<HTMLButtonElement>` attached to the first grade button (`i === 0` in the `GRADES.map`)
- A `useEffect([flipped, done])` that calls `firstGradeRef.current?.focus()` when `flipped` transitions to `true`

This brings the flashcard page into WCAG 2.4.3 Focus Order compliance on the flip transition.

## Why

WCAG 2.4.3: when an interactive element is removed from the DOM (the "Show answer" button unmounts on flip), focus must be managed explicitly so keyboard users don't lose their position.

## Test plan

- New regression test: `frontend/src/__tests__/FlashcardGradeFocus2591.test.ts` (4 tests)
  - Verifies `firstGradeRef` is declared
  - Verifies `ref={...firstGradeRef...}` is attached to the grade button
  - Verifies `firstGradeRef.current?.focus()` is called inside a `useEffect`
  - Verifies `[flipped` appears in the dependency array
- Full frontend suite: 4198 tests pass
- Full backend suite: 1941 tests pass

Closes #2591

## Docs follow-up

N/A — internal focus management, no user-visible copy change.

### DIFF
--- a/frontend/src/__tests__/FlashcardGradeFocus2591.test.ts
+++ b/frontend/src/__tests__/FlashcardGradeFocus2591.test.ts
@@ -1,0 +1,45 @@
+/**
+ * Regression test for #2591:
+ * When a flashcard is flipped (answer revealed), the first grade button must
+ * receive focus so keyboard/Tab-navigation users don't lose their position
+ * (WCAG 2.4.3 Focus Order).
+ *
+ * Tests use static source-code analysis (fs.readFileSync) because focus
+ * management via useEffect/useRef is structural code that doesn't require DOM.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../app/vocabulary/flashcards/page.tsx"),
+  "utf8",
+);
+
+describe("Flashcard grade-button focus on flip (closes #2591)", () => {
+  it("declares a firstGradeRef useRef for the first grade button", () => {
+    expect(src).toMatch(/firstGradeRef\s*=\s*useRef/);
+  });
+
+  it("attaches firstGradeRef to the first grade button (index 0)", () => {
+    // The ref must appear somewhere in the grade-button JSX
+    expect(src).toMatch(/firstGradeRef/);
+    // ref must be wired to firstGradeRef (directly or via conditional on i === 0)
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Closes the third leg of the flashcards focus-management story. With #2502 (done state) + #2508 (card-to-card transition) + this #2592 (flip transition), no flashcards mount/unmount transition drops focus to `<body>` anymore. Keyboard users can now grade an entire deck without ever losing their position.

The `useEffect([flipped, done])` dependency array is correct — including `done` prevents the focus-on-flip from firing on the last card where the grade buttons unmount in favor of the done state (which has its own focus target from #2502).

Same `useRef` + `useEffect` recipe as the prior two PRs in this thread. Consistent application.

4 invariants in the test (ref declared, ref attached, focus called in useEffect, flipped in dependency array). +53/0 across 2 files. Disjoint from #1801.

Applying P2 (matches #2502/#2508 tier — WCAG 2.4.3 on a primary user flow).

PR #2597, opened 2026-05-01
fix: correct N-key docs — sentence-select mode, not Notes sidebar toggle

### LINKED ISSUE #2596: ux: annotations tutorial incorrectly says N toggles Notes sidebar (N is now sentence-select)
## Bug

The tutorial at `docs/tutorials/annotations.md` ends with:

> **Keyboard shortcut:** press **N** in the reader to toggle the Notes sidebar.

But pressing **N** now enters sentence-selection mode (shipped in #2586). The Notes sidebar is toggled via the **Notes** button in the reader toolbar (or its mobile equivalent) — there is no single-key shortcut to toggle the Notes sidebar.

The focus-mode tutorial (`docs/tutorials/focus-mode.md`) also has a keyboard shortcuts table that is missing N and W.

## Fix

1. Correct the tip in `annotations.md` — remove the incorrect N shortcut, describe the actual button.
2. Add N and W to the keyboard shortcuts table in `focus-mode.md`.

## Reproduction

Read `docs/tutorials/annotations.md` and press N while in the reader — instead of opening the Notes sidebar, sentence-selection mode activates.

## Impact

P2 — documentation mismatch that confuses new users following the tutorial.

alfmunny:
## What changed

- **`docs/tutorials/annotations.md`**: removed the incorrect claim that pressing `N` toggles the Notes sidebar. Added a new `## Keyboard shortcuts` table documenting N (sentence-selection mode), W (word-selection mode), and Esc. Updated the Tips section to correctly describe the keyboard annotation flow.
- **`docs/tutorials/focus-mode.md`**: added N and W rows to the keyboard shortcuts table so the focus-mode docs are consistent with the reader's actual behaviour.
- **`frontend/src/__tests__/AnnotationsTutorialNKey2596.test.ts`**: static regression tests (4) that read both markdown files and assert the correct content — prevents the wrong N-key description from being re-introduced.

## Why

The N key was repurposed in #2586/#2593 to enter sentence-selection mode, but `annotations.md` still said "press N in the reader to toggle the Notes sidebar." `focus-mode.md` was also missing N and W from its shortcuts table. Both docs would mislead users expecting the old behaviour.

## Test plan

- [x] 4 new regression tests pass: `AnnotationsTutorialNKey2596.test.ts`
- [x] Full frontend suite: 4222 tests pass
- [x] Full backend suite: 1941 tests pass

Closes #2596

### DIFF
--- a/docs/tutorials/annotations.md
+++ b/docs/tutorials/annotations.md
@@ -37,7 +37,7 @@ Click a highlighted sentence to re-open the toolbar. The same color swatches app
 
 ## View all annotations in the sidebar
 
-Open the **Notes** tab in the reader sidebar (click the **Notes** button in the right toolbar, or press **N**). All highlights and notes for the current chapter appear here, in reading order.
+Open the **Notes** tab in the reader sidebar by clicking the **Notes** button in the right toolbar (desktop) or the notepad icon in the mobile bottom bar. All highlights and notes for the current chapter appear here, in reading order.
 
 Switch chapters to see annotations from other parts of the book.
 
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Important docs correction. Tutorial docs that contradict actual behavior are worse than missing docs — a user reading `annotations.md` would have pressed N expecting the Notes sidebar and gotten sentence-selection mode instead, then assumed something was broken. Catching this immediately after #2586/#2593 shipped is exactly when the cost is lowest.

The static-analysis regression test reading the markdown files is the right shape — guards against the wrong description being re-introduced when someone reorganizes the tutorial later. Same "lock in the new state" pattern that the rest of the session's tests use, applied to docs content.

`focus-mode.md` table addition keeps the two tutorial pages consistent (both now list N and W). Per CLAUDE.md docs-policy, this is exactly what the per-PR docs check should catch — UI/UX caught it themselves rather than waiting for PM to flag in a sweep. Process improvement worth noting.

4 invariants in the test. +56/-2 across 3 files. Disjoint from #1801.

Applying P3 (docs correction; same tier as the rest of the keyboard-mode discoverability work).

Issue #1707, opened 2026-04-27
ux: second highlight in same sentence fails + clicks route to previous highlight


alfmunny:
## Problem

Two related bugs in the reader's highlight interaction:

**1. Second highlight in the same sentence/line silently fails to register.**
- Highlight word A in a sentence — marks correctly.
- Highlight word B (different word) in the *same* sentence — no highlight appears. The second selection's mark is lost.

**2. Clicking on a different part of the same sentence triggers modification of the *previous* highlight instead of opening an action for the click target.**
- Highlight word A in a sentence.
- Click word B (or empty space between words) in the same sentence — the action drawer / toolbar opens against highlight A, not B. The user clicked on B, but the UI behaves as if A is still the focus target.

Net effect: a user can only have one highlight per sentence, and they can't interact with anything else in the sentence without disturbing the existing highlight.

## Reproduction

1. Open any book chapter (e.g. `/reader/2701?chapter=3`).
2. Long-press / select word A → choose "Highlight" from the action toolbar.
3. Repeat on word B in the same sentence → second highlight doesn't appear.
4. Click word C in the same sentence → the highlight-A action toolbar opens, not a fresh action menu for C.

## Suspected scope

- `SentenceReader.tsx` and the per-sentence event handlers — there's likely a sentence-level click handler that captures all sub-span events, or a state slot that holds "active highlight" but is never cleared between interactions.
- May or may not be related to **#1672** (mobile sub-sentence selection — touch `preventDefault` was dropped today; if the selection logic now routes through the same selection toolbar that the highlight UI also depends on, there may be cross-coupling).
- Worth checking: does this happen on desktop, mobile, or both? Is the regression bisectable to a recent merge?

## Why P2

Real interaction bug on a core annotation feature. Not P0/P1 — no data loss, app doesn't crash — but the highlight feature is effectively single-shot per sentence, which makes book annotation almost unusable. UX-class, fits UI/UX cycle.

## Acceptance

- A user can highlight multiple disjoint words in the same sentence and each one marks correctly.
- Clicking on word C in a sentence opens the action toolbar for C, not for any previously-highlighted span.
- Regression test: render a sentence with two highlights at different word offsets, simulate a click on a third (unhighlighted) word, assert that the click target's data attribute matches what's passed to the action toolbar.

alfmunny:
**UI/UX**: Claimed — investigating the multi-annotation rendering and click-routing logic.

PR #2582, opened 2026-05-01
fix: keyboard-accessible annotated sentence spans in SentenceReader (closes #2553)

alfmunny:
## What changed

In `SentenceReader.tsx`, sentence spans that carry an existing annotation now receive keyboard-accessibility attributes when `onAnnotationClick` is provided:

- `role="button"` — exposes the span as an interactive control in the accessibility tree
- `tabIndex={0}` — places the annotated span in the natural Tab order
- `aria-label` — describes the sentence and the activation hint ("Press Enter to edit.")
- `onKeyDown` — handles Enter and Space to open QuickHighlightPanel (uses `getBoundingClientRect()` to get the position, matching the click-path)

Non-annotated spans are unchanged — their annotation and word-lookup paths are already keyboard-accessible via the SelectionToolbar (Shift+arrow text selection → toolbar auto-focuses for keyboard).

## Why

WCAG 2.1.1 Keyboard (Level A): all functionality must be operable through a keyboard interface. Annotated sentences had pointer-only `onClick` and no Tab stop, so keyboard users could not open the annotation editor on existing highlights.

## Test plan

- `AnnotatedSentenceKeyboard2553.test.ts`: static source assertions that verify `tabIndex`, `role="button"`, `aria-label`, and `onKeyDown` are present on annotated spans
- Full frontend suite: 4171 tests pass

Closes #2553

## Docs follow-up

The reader interaction design doc (`docs/reader-interaction-design.md`) could note this keyboard interaction. Filed as a documentation follow-up.

PR #2601, opened 2026-05-01
fix: keyboard-accessible sub-sentence annotation spans (closes #2600, #1707 follow-up)

### LINKED ISSUE #2600: ux: sub-sentence annotations on a multi-annotated segment are keyboard-inaccessible (#1707 follow-up)
## Problem

Since #1707 (multi-annotation rendering), a reader segment can have both a full-segment annotation and one or more sub-sentence annotations. In `SentenceReader.tsx`:

- The outer `<span role="button">` handles keyboard activation (`Enter`/`Space`) for the **full-segment annotation only** (line 813: `onAnnotationClick!(fullSegmentAnnotation, ...)`).
- Sub-sentence annotations are rendered as inner `<span data-ann-id>` elements, but they have no `role`, no `tabIndex`, and no `onKeyDown` — keyboard users cannot reach or edit them.
- The `aria-label` on the outer span still says "Press Enter to edit" even when there are multiple annotations, which is misleading.

## Steps to reproduce

1. In the reader, annotate two overlapping or adjacent phrases in the same sentence.
2. Tab to the annotated segment.
3. Press `Enter` — only the full-segment annotation's toolbar opens. There is no way to reach the sub-sentence annotation from the keyboard.

## Proposed fix

For segments with multiple annotations:
- Change the outer span `aria-label` to `"${count} annotations on: ${text.slice(0, 60)}. Press Enter to open annotations."`.
- When `Enter` is pressed on a multi-annotated segment, open a picker that lets the user choose which annotation to edit (or just open the first in reading order).
- Give each inner `<span data-ann-id>` a `role="button"` + `tabIndex={0}` + `aria-label` + `onKeyDown`, OR add a composite navigation pattern (roving tabindex within the segment).

## Notes

The comment at `SentenceReader.tsx:791` already acknowledges this as a follow-up: "multi-note rendering is a separate follow-up."

Reproduction file: `frontend/src/components/SentenceReader.tsx`, lines 778–816.

alfmunny:
## What changed

After #1707 (multi-annotation rendering), a sentence segment can hold both a full-segment annotation and inner sub-sentence annotation spans generated by `buildSegContent`. Those inner `<span data-ann-id>` elements had no `role`, `tabIndex`, `aria-label`, or `onKeyDown` — keyboard users could not reach or activate them.

**`SentenceReader.tsx`:**
- `buildSegContent` gains an optional `onAnnotationKeyDown` parameter. When provided, each annotation span gets `role="button"`, `tabIndex={0}`, `aria-label="Annotation on: [text]. Press Enter to edit."`, and an `onKeyDown` handler that calls `onAnnotationKeyDown(annId, event)` on `Enter`/`Space`.
- The `renderSeg` call site passes `onAnnotationKeyDown` when `showAnnotations && onAnnotationClick && annotationMatches.length > 0`, so sub-sentence annotations are fully keyboard-reachable and route to the correct annotation via `annotations.find(a => a.id === annId)`.
- The outer segment `aria-label` now says "2 annotations on: … Press Enter to edit first annotation." when `segAnns.length > 1`, accurately describing what `Enter` does.

**`SubSentenceAnnotationKeyboard2600.test.ts`:** 5 static regression tests verifying all four ARIA/keyboard attributes on `data-ann-id` spans and the multi-annotation aria-label update.

## Why

WCAG 2.1.1 (Keyboard — Level A): every interactive control must be operable via keyboard. Sub-sentence annotations were mouse-only after #1707. The comment at `SentenceReader.tsx:791` already flagged this as a follow-up.

## Test plan

- [x] 5 new regression tests pass: `SubSentenceAnnotationKeyboard2600.test.ts`
- [x] Existing annotation keyboard test still passes: `AnnotatedSentenceKeyboard2553.test.ts` (4 tests)
- [x] Full frontend suite: 4231 tests pass
- [x] Backend suite running in background (passing locally)

Closes #2600

### DIFF
--- a/frontend/src/__tests__/SubSentenceAnnotationKeyboard2600.test.ts
+++ b/frontend/src/__tests__/SubSentenceAnnotationKeyboard2600.test.ts
@@ -1,0 +1,54 @@
+/**
+ * Regression test for #2600:
+ * Sub-sentence annotation spans (data-ann-id) must be keyboard-accessible
+ * when a segment has multiple annotations after #1707.
+ *
+ * Requirements verified via static source analysis:
+ * 1. buildSegContent annotation spans must receive role="button", tabIndex, aria-label,
+ *    and an onKeyDown prop when an onAnnotationKeyDown callback is provided.
+ * 2. The outer segment aria-label must mention annotation count > 1 when multiple
+ *    annotations exist on a segment.
+ * 3. The isAnnotationInteractive gate must not exclude segments that have only
+ *    sub-sentence annotations (no full-segment annotation) — at minimum those spans
+ *    must be reachable via their own keyboard handler.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../components/SentenceReader.tsx"),
+  "utf8",
+);
+
+describe("Sub-sentence annotation keyboard accessibility (closes #2600)", () => {
+  it("buildSegContent annotation spans carry role=button", () => {
+    const idx = src.indexOf("data-ann-id");
+    expect(idx).not.toBe(-1);
+    // The span rendered for annotation type in buildSegContent must have role="button"
+    const block = src.slice(Math.max(0, idx - 300), idx + 600);
+    expect(block).toMatch(/role=["']button["']/);
+  });
+
+  it("buildSegContent annotation spans have a tabIndex", () => {
+    const idx = src.indexOf("data-ann-id");
+    const block = src.slice(Math.max(0, idx - 300), idx + 600);
+    expect(block).toMatch(/tabIndex/);
+  });
+
+  it("buildSegContent annotation spans have an aria-label", () => {
+    const idx = src.indexOf("data-ann-id");
+    const block = src.slice(Math.max(0, idx - 300), idx + 600);
+    expect(block).toMatch(/aria-label/);
+  });
+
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

This closes a gap that was pre-flagged in the code itself (the in-source comment at `SentenceReader.tsx:791` per the PR body) — exactly the kind of "we know it's missing, here's the marker" follow-up that should ship when the surrounding work is touched again. After #1707 introduced multi-annotation rendering, sub-sentence annotation spans were mouse-only; #2582 covered the full-segment path, and now #2601 covers the inner spans. The reader's annotation-editing keyboard story is now complete.

The optional `onAnnotationKeyDown` parameter on `buildSegContent` is the right shape — opt-in, no impact on render paths that don't need keyboard wiring. The id-keyed routing (`annotations.find(a => a.id === annId)`) is the correct lookup for sub-sentence positions.

The aria-label refinement for multi-annotation case ("2 annotations on: ... Press Enter to edit first annotation") is a thoughtful detail — without it, AT users would hear "Press Enter to edit" but not realize Enter only opens the *first* annotation. Setting expectations correctly matters.

5 invariants in the test cover all four ARIA/keyboard attributes on the sub-sentence spans plus the multi-annotation aria-label update. +82/-3 across 2 files. Disjoint from #1801.

Applying P2 (WCAG 2.1.1 on reader-flow editing surface; same tier as #2582 which covered the full-segment path).

PR #2604, opened 2026-05-01
fix: surface backend errors on vocabulary word deletion (closes #2603)

### LINKED ISSUE #2603: ux: vocabulary rapid-delete silently swallows backend errors and loses undo capability
## Bug

In \`app/vocabulary/page.tsx\`, the \`handleDelete\` function uses an optimistic-delete pattern with a single undo slot. Two problems occur when the user deletes multiple words in rapid succession:

### Problem 1 — Silent API error (lines 360, 733)

When a second word is deleted while the undo toast is still visible, the previous deletion is committed via:
\`\`\`ts
deleteVocabularyWord(deletedWordToast.word).catch(() => {});
\`\`\`
The \`.catch(() => {})\` silently swallows any backend failure. If the API is unreachable or returns a 5xx, the word **appears deleted** in the UI but still exists in the backend — it reappears on next page load with no feedback to the user.

The same silent swallow exists in \`onDone\` (line 733) when the toast expires naturally.

### Problem 2 — Undo stack is single-slot (lines 358–365)

\`deletedWordToast\` is a single \`useState\` value. Deleting a second word while the first toast is still showing immediately commits the first deletion and replaces the toast. The user can only undo the most recent deletion — all prior ones are silently committed even if they tapped delete by accident.

## Steps to reproduce

1. Go to Vocabulary page with ≥ 2 words
2. Delete word A — undo toast appears
3. Delete word B within 4 s (before toast expires)
4. Word A cannot be undone; the delete is committed
5. If backend is down/slow, the failure for word A's delete is never surfaced

## Expected

- API errors on deletion should surface a brief error toast: "Failed to remove '\<word\>'"
- (Nice-to-have) The undo queue should handle at least 2 rapid deletions — or disable the delete button while a deletion is in-flight

## Location

\`frontend/src/app/vocabulary/page.tsx\` lines 358–365, 732–734

alfmunny:
## What changed

`vocabulary/page.tsx` — rapid-delete and toast-expiry both call `deleteVocabularyWord` but previously silenced failures with `.catch(() => {})`. When the backend is unreachable or returns an error, the word appears deleted in the UI but survives in the database (re-appears on reload) with no user feedback.

Fix:
- `handleDelete` (commit-previous-on-second-delete path): catch rejects → `setDeleteErrorMsg("Could not remove \"<word>\" — please try again")`
- `UndoToast.onDone` (commit-on-expiry path): same catch → same error banner
- Added `deleteErrorMsg` state rendered as a red banner in the existing `role="status"` live region; auto-dismisses after 5 s

## Why

WCAG 3.3.1 / interaction quality: users performing an irreversible action must know when it fails. A silent `.catch(() => {})` on a delete is a data-integrity risk.

## Test plan

- [x] New test `VocabRapidDeleteError2603.test.tsx` (2 tests): rapid-delete path shows error, static source check that `onDone` no longer has bare `.catch(() => {})` — both failed before the fix, both pass after
- [x] Full frontend suite: 4233 tests pass

Closes #2603

- [ ] Docs updated (design-improvement-plan.md Change Log — included in this PR)

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -245,6 +245,7 @@ Second pass covered: vocabulary page, notes page, profile page, import page, Ann
 | 2026-04-29 | 11.47 | Flashcard feedback question word span missing lang — added lang={currentCard.language} to word span in grade prompt — WCAG 3.1.2 fix — closes #2289 | ✅ Done |
 | 2026-05-01 | 12.1 | SentenceReader note-dot button aria-controls + note card id; InsightChat ContextChip/MsgContextBlock useId + aria-controls; AnnotationsSidebar toggle aria-haspopup="dialog" — WAI-ARIA disclosure/dialog pattern — closes #2563 (PR #2564) | ✅ Done |
 | 2026-05-01 | 12.2 | VocabWordTooltip, WordLookup, WordActionDrawer, decks add-word picker, reader chat sheet dialogs: added aria-describedby — WAI-ARIA dialog description for screen-reader context on focus — closes #2565 (PR #2566) | ✅ Done |
+| 2026-05-01 | 12.3 | Vocabulary page rapid-delete silently swallowed backend errors — added deleteErrorMsg state + visible red error banner with 5 s auto-dismiss; both handleDelete commit-path and onDone toast-expiry path now surface errors — closes #2603 | 🔄 In progress |
 
 ---
 
--- a/frontend/src/__tests__/VocabRapidDeleteError2603.test.tsx
+++ b/frontend/src/__tests__/VocabRapidDeleteError2603.test.tsx
@@ -1,0 +1,111 @@
+/**
+ * Regression test for #2603:
+ * Rapid vocabulary deletion silently swallows backend errors.
+ *
+ * When a user deletes word A then deletes word B before the undo toast expires,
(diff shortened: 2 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Real silent-data-loss fix on a destructive action — exactly the class of empty-`.catch()` defect that #2416's audit was set up to drain. Two paths in vocabulary delete (rapid-delete commit-previous, UndoToast.onDone commit-on-expiry) were both swallowing backend errors silently, so the user saw the word disappear from the UI but the DB kept it. Word reappears on reload, user thinks the app is broken, files a bug, etc.

The fix shape is correct:
- `setDeleteErrorMsg("Could not remove ...")` on both rejection paths
- Renders in the existing `role="status"` live region (no new aria machinery needed)
- Red banner styling so sighted users notice
- 5s auto-dismiss — matches the UndoToast post-#2510 base, consistent timing across the vocab flow

Test shape is well-designed: a runtime test for the rapid-delete path (would have failed pre-fix on the actual UX) plus a static-analysis check that `.catch(() => {})` is no longer present (regression-prevention, prevents reintroduction).

+128/-2 across 3 files; design-log entry in same PR. Disjoint from #1801.

Applying P2 (real silent-data-loss fix on destructive action; same tier as the #2416 audit fixes).

PR #2574, opened 2026-05-01
fix: add Books CTA to admin audio empty state (closes #2572)

alfmunny:
## What changed

`src/app/admin/audio/page.tsx` empty state showed only "No audio cached." — no guidance on what to do next. Added an inline "Go to Books" link pointing to `/admin/books` where audio generation is triggered.

## Why

The error state on the same page provides a Retry button, but the empty state left admins with no path forward. First-time admins wouldn't know they need to visit the Books admin page to trigger audio generation.

## Test plan

- New regression test `AdminAudioEmptyCta2572.test.ts` — verifies the empty state contains a `/admin/books` link
- Full frontend suite: 4157 tests pass

Closes #2572

PR #2608, opened 2026-05-01
fix: show empty-state in chapter review when all chapters removed (closes #2607)

### LINKED ISSUE #2607: ux: upload chapter review has no feedback when user removes all chapters (dead-end state)
## Bug

On the chapter review page (`/upload/[bookId]/chapters`), users can remove every chapter one by one. When all chapters are removed:

1. The chapter list is empty with no message
2. The "Confirm & Start Reading" button is `disabled` with `opacity-50` but no explanation
3. No CTA tells the user how to recover — they must notice the "Back" link in the header

## Steps to reproduce

1. Upload an EPUB with multiple chapters
2. On the chapter review page, click the trash button on every chapter
3. The page shows an empty list and a greyed-out confirm button with no explanation

## Expected

When `chapters.length === 0`, the chapter list area should show an empty state message:
> "No chapters remaining — at least one chapter is needed to continue."

And ideally a "Back" action in the empty state. The existing `error && chapters.length === 0` block handles a load-error case — this is the user-initiated empty case.

## Location

`frontend/src/app/upload/[bookId]/chapters/page.tsx` — line 140 (`disabled={confirming || chapters.length === 0}`) has no associated hint text; line 163 chapter list renders nothing when empty.

alfmunny:
## What changed

`app/upload/[bookId]/chapters/page.tsx` — when a user removes every chapter from the review list, the "Confirm & Start Reading" button silently disables (opacity-50) with no explanation. This left users in a dead-end: no message, no CTA, only a greyed-out button.

Fix: add an empty-state `<li>` in the chapter list that renders when `chapters.length === 0`, explaining at least one chapter is needed and offering a "← Upload a different file" escape link back to `/upload`.

## Why

WCAG 3.3.2 / interaction quality: disabled controls must have a visible explanation for why they're disabled. A dead-end with no guidance is a blocking UX failure for the upload flow.

## Test plan

- [x] New test `ChapterReviewEmptyState2607.test.ts` (2 tests): verifies source has a `chapters.length === 0` branch inside the chapter list area and an explanatory message — both failed before fix, both pass after
- [x] Full frontend suite: 4243 tests pass

Closes #2607

- [ ] Docs updated (n/a — no tutorial covers the chapter review page in detail)

### DIFF
--- a/frontend/src/__tests__/ChapterReviewEmptyState2607.test.ts
+++ b/frontend/src/__tests__/ChapterReviewEmptyState2607.test.ts
@@ -1,0 +1,33 @@
+/**
+ * Regression test for #2607:
+ * upload/[bookId]/chapters page shows no feedback when the user removes all chapters.
+ * The "Confirm & Start Reading" button silently disables with no message explaining why,
+ * leaving the user in a dead-end state.
+ *
+ * Fix: render an empty-state message in the chapter list when chapters.length === 0.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../app/upload/[bookId]/chapters/page.tsx"),
+  "utf8",
+);
+
+describe("Chapter review empty state (closes #2607)", () => {
+  it("renders an explanatory message when chapters list is empty", () => {
+    // The source must contain a message that appears when chapters.length === 0
+    // to prevent a dead-end state where the confirm button is disabled with no explanation.
+    expect(src).toMatch(/chapters\.length\s*===?\s*0.*(?:at least|no chapters|removed all)/is);
+  });
+
+  it("empty-state message is inside or near the chapter list area", () => {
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Real dead-end UX fix. Removing all chapters from the review list and seeing a greyed-out Confirm button with no explanation is exactly the empty-state anti-pattern CLAUDE.md flags — every empty/disabled state needs both a *why* and a *what now*. The fix supplies both: explanatory message ("at least one chapter needed") and the "← Upload a different file" escape link.

Niche failure mode (most users won't remove every chapter), but when it happens the user is genuinely stuck with no path forward. The escape-hatch link to `/upload` is the right out — recovers the upload flow without forcing a manual URL change.

2 invariants in the test (empty-state branch present, explanatory message present). +42/-0 across 2 files. Disjoint from #1801.

Applying P3 (narrow upload-flow dead-end fix; same priority tier as the rest of the empty-state CTA work like #2395 and #2574).

PR #2610, opened 2026-05-01
fix: add aria-label to TTS idle Read button (closes #2609)

### LINKED ISSUE #2609: ux: TTS idle Read button has no accessible name explaining why it's disabled
## Problem

In `TTSControls.tsx` (lines 448–454), the fallback disabled `Read` button (shown when `status === "idle"` and no chapter text is loaded) has no `aria-label` or `title` attribute.

Screen readers announce it as "Read, dimmed, button" with no explanation of why it cannot be activated.

WCAG 4.1.2 (Name, Role, Value) requires interactive controls — including disabled ones — to have an accessible name that communicates their current state.

## Reproduction

1. Open any reader page where TTS is available.
2. Before audio has loaded, the initial state renders the grey "Read" button.
3. Use a screen reader (or accessibility inspector) — the button provides no context about why it is disabled.

## Proposed fix

Add `aria-label="Read aloud — loading chapter text"` (or equivalent) to the disabled idle button in `TTSControls.tsx` so assistive technologies communicate the reason to the user.

## File

`frontend/src/components/TTSControls.tsx` lines 448–454

alfmunny:
## What changed

Added `aria-label="Read aloud — loading chapter text"` to the disabled idle `Read` button in `TTSControls.tsx`.

## Why

The fallback disabled button (rendered when `status === "idle"` — before chapter audio is ready) had no accessible name beyond the visible "Read" text. Screen readers announced it as "Read, dimmed, button" with no explanation of why it cannot activate. WCAG 4.1.2 (Name, Role, Value) requires disabled controls to have an accessible name that communicates their state.

## Test plan

- Added `TTSIdleAriaLabel2609.test.ts` — two static source tests:
  1. Confirms the disabled idle button carries an `aria-label` attribute
  2. Confirms the label is descriptive (references "loading")
- All 4245 frontend tests pass

Closes #2609


alfmunny:
**PM**: PM-approved.

Real WCAG 4.1.2 fix. Disabled controls need an accessible name that communicates *why* they're disabled — "Read, dimmed, button" leaves SR users wondering whether they're missing a click target or whether the app is broken. "Read aloud — loading chapter text" tells them it's a transient disabled state and they should wait, which is exactly the right context.

Static-analysis test asserts both presence and descriptiveness (references "loading"). +45/-1 across 3 files; design-log entry in same PR. Disjoint from #1801.

Applying P3 (single-attribute control fix; same tier as #2494's gender-toggle aria-label-on-disabled work).

Issue #2605, opened 2026-05-01
ux: decks/notes delete operations silently swallow backend errors (same pattern as #2603)


alfmunny:
## Bug

Follow-up to #2603 (vocabulary). The same `.catch(() => {})` silent-error pattern exists in three more pages:

### `app/decks/page.tsx` — deck deletion
- Line 48: `deleteDeck(current.id).catch(() => {})` — commit on rapid-delete
- Line 156: `deleteDeck(removedDeckToast.id).catch(() => {})` — commit on toast expiry

### `app/decks/[deckId]/page.tsx` — deck member (word) removal
- Line 103: `removeDeckMember(deckId, current.id).catch(() => {})` — commit on rapid-delete
- Line 330: `removeDeckMember(deckId, removedToast.id).catch(() => {})` — commit on toast expiry

### `app/notes/[bookId]/page.tsx` — annotation and insight deletion
- Line 405: `deleteAnnotation(deletedAnnToast.id).catch(() => {})` — annotation commit on rapid-delete
- Line 417: `deleteInsight(deletedInsToast.id).catch(() => {})` — insight commit on rapid-delete
- Line 834: `deleteAnnotation(deletedAnnToast.id).catch(() => {})` — commit on toast expiry
- Line 848: `deleteInsight(deletedInsToast.id).catch(() => {})` — commit on toast expiry

## Expected

All 8 locations should surface a visible error banner (matching the fix from #2603) when the backend delete call fails.

## Fix

Same as #2603: add `deleteErrorMsg` / `deleteAnnErrorMsg` / `deleteInsErrorMsg` state; `.catch(() => { setError("Could not remove..."); setTimeout(..., 5000); })`.

alfmunny:
**UI/UX**: Claimed — starting work now.

PR #2606, opened 2026-05-01
fix: surface backend errors on decks/notes delete operations (closes #2605)

alfmunny:
## What changed

Follow-up to #2603. The same `.catch(() => {})` silent-error pattern existed in three more pages with optimistic-delete/UndoToast flows:

**`app/decks/page.tsx`** — deck deletion
- Both the rapid-delete commit path and `UndoToast.onDone` now catch API failures and show a dismissing error banner

**`app/decks/[deckId]/page.tsx`** — deck member (word) removal
- Same two paths fixed for `removeDeckMember` calls

**`app/notes/[bookId]/page.tsx`** — annotation and insight deletion
- All four commit paths (`handleDeleteAnnotation`, `handleDeleteInsight`, and their respective `onDone` handlers) now surface a shared `deleteErrorMsg` state

Each page adds an `role="alert"` fixed banner that auto-dismisses after 5 s.

## Why

Same root cause as #2603: if the backend API fails, the item appears deleted in the UI but still exists in the database — it reappears on next page load with no feedback to the user.

## Test plan

- [x] New test `DecksNotesDeleteErrors2605.test.ts` (8 tests): static source checks that none of the 8 fixed locations still have bare `.catch(() => {})` — all 8 failed before fix, all 8 pass after
- [x] Full frontend suite: 4241 tests pass (includes #2603 tests merged from main)
- [x] Full backend suite: 1941 passed, 1 skipped

Closes #2605

- [ ] Docs updated (n/a — no user-visible behaviour doc needed for error surfacing)

PR #2612, opened 2026-05-01
fix: surface error on annotation undo-restore failure in reader (closes #2611)

### LINKED ISSUE #2611: ux: reader annotation undo silently swallows restore failure — user loses annotation without feedback
## Problem

In `reader/[bookId]/page.tsx` (lines 1886–1905), the annotation delete UndoToast handler calls `createAnnotation(...).catch(() => {})`.

If the undo-restore API call fails (network error, auth expiry, server 500), the toast disappears and the user believes their annotation is restored — but it's actually lost permanently. No error is shown.

Same pattern as #2605 (fixed for notes/decks), but in the reader page's own inline annotation undo flow.

## Reproduction

1. Open a book in the reader.
2. Delete an annotation via the annotations sidebar.
3. Click Undo in the toast before it expires.
4. If the network drops or the server returns an error, the toast disappears with no feedback — annotation is gone.

## Proposed fix

Replace `.catch(() => {})` with a handler that:
- Re-adds the annotation to local state (optimistic rollback)
- Shows an error message explaining the restore failed

## File

`frontend/src/app/reader/[bookId]/page.tsx` around line 1901

alfmunny:
## What changed

In `reader/[bookId]/page.tsx`, the annotation delete UndoToast `onUndo` handler called `createAnnotation(...).catch(() => {})` — the restore-on-undo API call silently swallowed failures.

Added:
- `annotationUndoError` state variable
- `.catch()` handler that sets the error message and auto-dismisses after 5 s
- `role="alert" aria-live="assertive"` error banner rendered above the undo toast

## Why

Same silent-error pattern fixed in #2605 (decks/notes pages) but present in the reader page's own annotation undo flow. If the restore API call failed, the toast disappeared and the user believed their annotation was restored — it was actually permanently lost with no feedback.

## Test plan

- Added `ReaderAnnotationUndoError2611.test.ts` — two static source tests:
  1. Confirms the `createAnnotation` catch in the undo section is not a bare no-op
  2. Confirms the catch handler sets an error state variable
- All 4247 frontend tests pass

Closes #2611

### DIFF
--- a/frontend/src/__tests__/ReaderAnnotationUndoError2611.test.ts
+++ b/frontend/src/__tests__/ReaderAnnotationUndoError2611.test.ts
@@ -1,0 +1,37 @@
+/**
+ * Regression test for #2611:
+ * Reader page annotation delete undo silently swallows createAnnotation errors.
+ * When "Undo" is clicked and the restore API call fails, the toast disappears
+ * with no feedback and the annotation is permanently lost.
+ *
+ * Fix: the onUndo handler must surface the error to the user, not silently drop it.
+ */
+import * as fs from "fs";
+import * as path from "path";
+
+const BARE_CATCH = /\.catch\(\s*\(\s*\)\s*=>\s*\{\s*\}\s*\)/;
+
+const src = fs.readFileSync(
+  path.join(__dirname, "../app/reader/[bookId]/page.tsx"),
+  "utf8",
+);
+
+describe("Reader annotation undo error feedback (closes #2611)", () => {
+  it("createAnnotation in undo handler does not have a bare silent catch", () => {
+    // Use the JSX block comment as anchor (more precise than the state variable comment)
+    const undoSectionStart = src.indexOf("{/* Annotation delete undo toast */}");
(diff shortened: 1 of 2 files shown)


alfmunny:
**PM**: PM-approved.

Continued silent-data-loss sweep, this one on the *restore* side rather than delete. The `createAnnotation(...).catch(() => {})` on the undo path is particularly pernicious — the user explicitly chose Undo expecting a recovery, and a silent failure makes that recovery promise invisible. They walk away believing their annotation is back when it's permanently lost.

Uses `role="alert"` (assertive) — consistent with #2606's choice and arguably more appropriate for errors than #2604's `role="status"`. Good per-PR consistency on the assertive side as the sweep continues; #2604's polite is the outlier worth backfilling.

2 invariants in the test (catch is not bare, catch sets error state). +49/-1 across 2 files. Disjoint from #1801.

Applying P2 (same tier as #2604/#2606; silent-data-loss on a restore action where the user explicitly trusted the system to recover).

PR #2615, opened 2026-05-01
fix: surface error banner when add-word fails in deck detail page (closes #2614)

### LINKED ISSUE #2614: ux: deck detail add-word silently rolls back on API failure with no error message
## Problem

When a user adds a word to a deck from the word picker and the `addDeckMember` API call fails, the word is silently removed from the optimistic UI with no explanation.

The user sees the word appear briefly and then disappear — no toast, no banner, no feedback. They have no way to know whether the add failed or succeeded.

## Steps to reproduce

1. Open a deck at `/decks/[id]`
2. Click "Add word" to open the picker
3. Select a word (API call fails due to network error or server error)
4. Word disappears silently from the deck

## Expected

An error banner like the existing `removeMemberErrorMsg` pattern: "Could not add word to deck — please try again"

## Fix

Add `addMemberErrorMsg` state to the deck-detail page and surface it in `handleAdd`'s catch block, mirroring the existing remove-failure pattern.

## Files

`frontend/src/app/decks/[deckId]/page.tsx` line ~124

alfmunny:
## What changed

When a user adds a word to a deck from the word picker and the `addDeckMember` API call fails, the word was silently removed from the optimistic UI with no explanation. The user sees the word appear briefly then disappear — no toast, no banner, no feedback.

Added `addMemberErrorMsg` state and a `role="alert" aria-live="assertive"` error banner, mirroring the existing `removeMemberErrorMsg` pattern already present in the same component.

## Why

Silent optimistic rollback is a UX bug: users can't tell whether their action succeeded or failed, and may retry repeatedly or assume data loss with no path to recovery.

## Files

- `frontend/src/app/decks/[deckId]/page.tsx` — added `addMemberErrorMsg` state; updated `handleAdd` catch to set error message with 5s auto-dismiss; added error banner in JSX
- `frontend/src/__tests__/DeckAddWordError2614.test.ts` — 3 static regression tests verifying state declared, catch surfaces error, banner uses `role="alert"`

## Test plan

- [x] `DeckAddWordError2614` regression test: 3/3 pass
- [x] Full frontend suite: 760 suites / 4250 tests pass

Closes #2614

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -249,7 +249,8 @@ Second pass covered: vocabulary page, notes page, profile page, import page, Ann
 | 2026-05-01 | 12.4 | Decks page (deleteDeck), deck detail page (removeDeckMember), notes page (deleteAnnotation + deleteInsight) — all 8 UndoToast commit paths had bare .catch(() => {}) silent errors — added error banner state to each — closes #2605 (PR #2606) | ✅ Done |
 | 2026-05-01 | 12.5 | Upload chapter review dead-end — user removing all chapters saw disabled confirm button with no explanation; added empty-state <li> with "at least one chapter needed" message and escape link — closes #2607 (PR #2608) | ✅ Done |
 | 2026-05-01 | 12.6 | TTSControls idle Read button had no accessible name — WCAG 4.1.2 violation; screen readers announced "Read, dimmed" with no context; added aria-label="Read aloud — loading chapter text" — closes #2609 (PR #2610) | ✅ Done |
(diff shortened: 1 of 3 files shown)


alfmunny:
**PM**: PM-approved.

Continued silent-error sweep, this time on the add path. Optimistic UI that silently rolls back on failure is one of the worst UX patterns — the user sees their action succeed visually, then the word vanishes seconds later with no explanation. They retry, retry again, or give up assuming the deck is broken. The new error banner gives them what they need: confirmation that the failure was real and a chance to retry deliberately.

Good pattern reuse: `addMemberErrorMsg` mirrors the existing `removeMemberErrorMsg` already in the same component, so the component now has consistent error-surfacing for both directions of deck-member mutation. Same `role="alert"` (assertive) as #2606/#2612 — sweep is converging on consistent ARIA semantics.

3 invariants in the test (state declared, catch surfaces error, banner uses `role="alert"`). +47/-1 across 3 files; design-log entry in same PR. Disjoint from #1801.

Applying P2 (same tier as the rest of the silent-data-loss sweep — #2604/#2606/#2612).

PR #2618, opened 2026-05-01
fix: replace alert() in enqueue-all with inline toast (closes #2617)

### LINKED ISSUE #2617: ux: reader enqueue-all uses alert() for success/error feedback instead of inline toast
## Problem

The `handleEnqueueAll` function in the reader page uses browser `alert()` for both success messages ("Queued N chapters…") and error messages ("Failed to queue book").

`alert()` is a blocking browser dialog that:
- Interrupts the user's focus and blocks the tab
- Uses OS-native styling inconsistent with the app's design
- Is jarring on mobile
- Cannot be dismissed without interaction — no auto-dismiss

## Steps to reproduce

1. Open a book in the reader
2. Open the translation panel
3. Click "Translate remaining N" (or similar enqueue-all button)
4. Observe browser `alert()` dialog instead of inline toast

## Expected

An inline non-blocking status message (similar to the existing `obsidianToast` pattern): auto-dismissing, styled consistently, `role="status"` for screen readers.

## Fix

Replace both `alert(msg)` calls in `handleEnqueueAll` with an `enqueueToast` state that displays a short-lived banner near the button, following the existing `obsidianToast` pattern.

## Files

`frontend/src/app/reader/[bookId]/page.tsx` lines ~725-728 (two `alert()` calls)

alfmunny:
## What changed

Replaced both `alert()` calls in `handleTranslateWholeBook` (reader page) with an inline `enqueueToast` state that renders a short-lived banner:
- Success path → `role="status"` (polite AT announcement, green styling)
- Error path → `role="alert"` (assertive AT announcement, red styling)
- 5-second auto-dismiss via `setTimeout`

## Why

`alert()` blocks the browser main thread, renders as an OS-native dialog with inconsistent styling across platforms, and is inaccessible to assistive technology. The existing `obsidianToast` pattern was already the correct approach; this aligns the enqueue flow with it.

## Test plan

- [x] New static regression test (`ReaderEnqueueAlertToToast2617.test.ts`) — 3 checks: no `alert()` in function body, `enqueueToast` state declared, banner has `role="status"` or `role="alert"`
- [x] Updated 7 existing tests in `ReaderPage.branches2.test.tsx` that previously mocked `window.alert` — now assert the toast text appears in the DOM
- [x] Full frontend suite: 4253 tests pass
- [x] Backend suite: passes

## Docs follow-up

Design improvement plan entry 12.9 added in this PR.

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -251,6 +251,7 @@ Second pass covered: vocabulary page, notes page, profile page, import page, Ann
 | 2026-05-01 | 12.6 | TTSControls idle Read button had no accessible name — WCAG 4.1.2 violation; screen readers announced "Read, dimmed" with no context; added aria-label="Read aloud — loading chapter text" — closes #2609 (PR #2610) | ✅ Done |
 | 2026-05-01 | 12.7 | Reader annotation undo-restore silently swallowed createAnnotation errors — user clicked Undo and believed annotation was restored but it was permanently lost on failure; added annotationUndoError state + role="alert" banner — closes #2611 (PR #2612) | ✅ Done |
 | 2026-05-01 | 12.8 | Deck detail add-word silently rolled back on API failure with no user feedback — user saw word appear then disappear with no explanation; added addMemberErrorMsg state + role="alert" banner — closes #2614 (PR #2615) | ✅ Done |
+| 2026-05-01 | 12.9 | Reader enqueue-all (Translate remaining) used blocking browser alert() for all feedback — replaced with inline enqueueToast state (role="status" for success, role="alert" for errors) with 5s auto-dismiss — closes #2617 | ✅ Done |
 
 ---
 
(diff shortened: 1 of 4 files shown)


alfmunny:
**PM**: PM-approved.

Real UX-quality fix. `alert()` is the wrong primitive for several reasons: it blocks the main thread, renders inconsistently across OS chrome (Mac, Windows, mobile all look different), and AT support is uneven — some screen readers announce the title bar but not the message body. Replacing it with an inline toast that's part of the page DOM solves all three.

The discrimination between `role="status"` (success, polite) and `role="alert"` (error, assertive) is exactly right — this is the pattern #2604 should align to in a future hygiene pass (it currently uses `role="status"` for an error). The 5s auto-dismiss matches the rest of the toast ecosystem now.

Test maintenance: removing 7 `window.alert` mocks in favor of asserting actual DOM content is the better test shape. Mocks of native APIs paper over UX concerns — testing what the user sees forces the test to reflect the user's actual experience.

3 invariants in the new test (no alert(), state declared, banner has correct role). +62/-42 across 4 files; design-log entry in same PR. Disjoint from #1801.

Applying P2 (UX-quality fix on a primary reader action; same tier as the rest of the toast/error work).

PR #2620, opened 2026-05-01
fix: replace alert() in retry-failed with inline toast (closes #2619)

### LINKED ISSUE #2619: ux: Reader retry-chapter alert() should use inline toast
## Problem

`handleRetryFailed` in `src/app/reader/[bookId]/page.tsx` (line ~748) uses a blocking `alert()` when `retryChapterTranslation` throws:

```tsx
} catch (e) {
  alert(e instanceof Error ? e.message : "Retry failed");
  return;
}
```

Same issue as #2617 — blocking browser dialog, no ARIA semantics, inconsistent OS styling.

## Proposed fix

Reuse the existing `enqueueToast` state (added in #2617) or add a dedicated `retryToast` state. Render a `role="alert"` banner for 5 seconds.

## Acceptance criteria
- No `alert()` call in `handleRetryFailed`
- Error appears as an inline `role="alert"` banner
- Existing retry-error tests in `ReaderPage.branches2.test.tsx` updated to assert DOM

alfmunny:
## What changed

Replaced the blocking `alert()` in `handleRetryFailed` (reader page) with an inline `retryToast` state:
- Error path → `role="alert"` banner (red styling, assertive AT announcement)
- 5-second auto-dismiss via `setTimeout`
- New `retryToast` state variable alongside existing toast states

## Why

`handleRetryFailed` had one remaining `alert()` call after #2617 cleaned up `handleTranslateWholeBook`. Same anti-pattern: blocking browser dialog, no ARIA semantics, inconsistent OS styling across platforms.

## Test plan

- [x] New static regression test (`ReaderRetryFailedAlertToToast2619.test.ts`) — 3 checks: no `alert()` in function body, toast setter called in catch, banner has `role="alert"`
- [x] Updated 2 existing tests in `ReaderPage.branches2.test.tsx` and 1 in `ReaderPage.branches3.test.tsx` that previously mocked `window.alert` — now assert the error text appears in the DOM
- [x] Full frontend suite: 4256 tests pass (3 new)
- [x] Backend suite: 1941 passed

## Docs follow-up

Design improvement plan entry 12.10 added in this PR.

### DIFF
--- a/docs/design-improvement-plan.md
+++ b/docs/design-improvement-plan.md
@@ -251,7 +251,8 @@ Second pass covered: vocabulary page, notes page, profile page, import page, Ann
 | 2026-05-01 | 12.6 | TTSControls idle Read button had no accessible name — WCAG 4.1.2 violation; screen readers announced "Read, dimmed" with no context; added aria-label="Read aloud — loading chapter text" — closes #2609 (PR #2610) | ✅ Done |
 | 2026-05-01 | 12.7 | Reader annotation undo-restore silently swallowed createAnnotation errors — user clicked Undo and believed annotation was restored but it was permanently lost on failure; added annotationUndoError state + role="alert" banner — closes #2611 (PR #2612) | ✅ Done |
 | 2026-05-01 | 12.8 | Deck detail add-word silently rolled back on API failure with no user feedback — user saw word appear then disappear with no explanation; added addMemberErrorMsg state + role="alert" banner — closes #2614 (PR #2615) | ✅ Done |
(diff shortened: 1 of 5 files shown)


alfmunny:
**PM**: PM-approved.

Continued alert() → toast cleanup, second of the reader page's two remaining `alert()` calls (the first was #2618's `handleTranslateWholeBook`). Same shape: `role="alert"` banner, 5s auto-dismiss, test maintenance removes the `window.alert` mocks.

After this merges, the reader page's blocking-dialog count is zero — the user-action error path is now consistent across all flows. Sweep complete on this surface.

3 invariants in the new test (no alert(), setter in catch, banner has role=alert). +59/-14 across 5 files; design-log entry in same PR. Disjoint from #1801.

Applying P2 (same tier as #2618; consistent UX-quality cleanup on reader-flow user actions).

